PR #12, opened 2017-08-18
Use default AWS credentials

### LINKED ISSUE #5: wal-g should be less smart about AWS credentials
`wal-g` requires the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` in order to [prepare an `aws.Config` that uses them](https://github.com/wal-g/wal-g/blob/a533befe61ddb8035bee77dfe9a23e17dcbc6e51/upload.go#L93-L96). This allows `wal-g` to return an error message early if it's not configured, but this approach breaks three other methods of supplying credentials which I actually use:

*  `AWS_PROFILE` pointing to something defined in `~/.aws/{config,credentials}`,
* EC2 instances [with dynamic credentials from an IAM role](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials), and
* Containers with dynamic credentials, both via [IAM roles for tasks](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) inside EC2 and [`iam-docker`](https://github.com/swipely/iam-docker) outside EC2

The AWS SDK for Go enables all of these by default, in addition to allowing configuration via `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Removing the environment check and leaving `Credentials` unspecified will preserve the current behavior _and_ re-enable all the other ways that typical AWS tools search for credentials.

### LINKED ISSUE #10: AWS_SECURITY_TOKEN is not really optional
Readme states `AWS_SECURITY_TOKEN` is optional, I keep getting:
```
postgres_1  | 2017/08/18 20:44:54 FATAL: Did not set the following environment variables:
postgres_1  | AWS_SECURITY_TOKEN
```

willglynn:
This fixes #5, and also fixes #10 by accident.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,16 +36,12 @@ To connect to Amazon S3, WAL-G requires that these variables be set:
 
 * `WALE_S3_PREFIX` (eg. `s3://bucket/path/to/folder`)
 * `AWS_REGION`(eg. `us-west-2`)
-* `AWS_ACCESS_KEY_ID`
-* `AWS_SECRET_ACCESS_KEY`
+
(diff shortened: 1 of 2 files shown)


willglynn:
Closing due to 68f31ea9aeab73235224f8dbd4396c63cba97818.

Issue #6, opened 2017-08-18
AWS_REGION should be optional


willglynn:
Right now, the user has to specify `WALE_S3_PREFIX` and `AWS_REGION`. However, `WALE_S3_PREFIX` indicates an S3 bucket, S3 buckets have globally unique names, and the [`s3:GetBucketLocation` API](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.GetBucketLocation) can take a S3 bucket name and return its region. `AWS_REGION` should therefore be optional, with `wal-g` able to determine the bucket's region automatically.

fdr:
Unfortunately, it's not optional in every region, and new regions will make it less and less optional, on account of SigV4. GetBucketLocation is not always granted via IAM and I'm not too keen to expend extra API round-trips for this.

willglynn:
I didn't say `wal-g` should disregard `AWS_REGION`, I suggested it should fall back to calling `s3:GetBucketLocation` if `AWS_REGION` were unspecified. Paired with #5, this would allow `wal-g` to run knowing only `WALE_S3_PREFIX`, with implicit defaults functioning for all the other parameters.

Would you accept a PR to this effect?

PR #13, opened 2017-08-18
Fix region handling and make it optional

willglynn:
I was going to open a PR per my suggestion in #6 anyway, and this seems as good a time as any. 

I believe 68f31ea9aeab73235224f8dbd4396c63cba97818 broke `AWS_REGION`, entirely removing its effect on the AWS configuration. This commit a) restores `AWS_REGION` and b) makes it optional, falling back to `s3:GetBucketLocation` if it is unspecified.

@fdr didn't like this idea because it's an extra API call that might fail, but I insist that there's no downside. If the user specifies `AWS_REGION`, it'll get used, period. If the user doesn't specify `AWS_REGION`, WAL-G can call `s3:GetBucketLocation`. If that fails too, WAL-G terminates -- which is what it does now. If it succeeds, WAL-G can proceed without requiring the user to specify `AWS_REGION`. This change adds another _success_ mode, not another failure mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,17 +35,16 @@ Configuration
 To connect to Amazon S3, WAL-G requires that these variables be set:
 
 * `WALE_S3_PREFIX` (eg. `s3://bucket/path/to/folder`)
-* `AWS_REGION`(eg. `us-west-2`)
-* `AWS_ACCESS_KEY_ID`
-* `AWS_SECRET_ACCESS_KEY`
+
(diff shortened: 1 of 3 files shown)


fdr:
The downside is a silently reduced performance, but perhaps it's okay. What I don't get is why AWS_REGION is not handled by the SDK. There appears to be code to do so. One of the things I was hoping to do was get rid of credential management code as seen in WAL-E (because it pre-dated first-party SDKs)

Issue #15, opened 2017-08-19
Support alternative S3 implementations


bithavoc:
Additionally, in order to support other non-official S3 implementations such as https://minio.io/ we would need to be able to setup a few custom settings:

```
Endpoint:         aws.String(os.Getenv("AWS_ENDPOINT")),
S3ForcePathStyle: aws.Bool(os.Getenv("AWS_S3_FORCE_PATH_STYLE") == "true"),
```

Example:
```
WALE_S3_PREFIX: s3://backups/wal-g
AWS_ENDPOINT: http://minio:9000
```

`S3ForcePathStyle` is required so the aws go sdk doesn't try to call `http://backups.s3.amazonaws.com/wal-g` and instead reach the minio server using sub path resources as in `http://minio:9000/backups/wal-g`.

Source https://github.com/minio/cookbook/blob/master/docs/aws-sdk-for-go-with-minio.md

fdr:
Closing in favor of PR.

PR #16, opened 2017-08-19
Add support for AWS_S3_FORCE_PATH_STYLE and AWS_ENDPOINT env variables

bithavoc:
adds support for S3-compatible uploads #15

### DIFF
--- a/upload.go
+++ b/upload.go
@@ -82,6 +82,13 @@ func Configure() (*TarUploader, *Prefix, error) {
 		return nil, nil, errors.Wrapf(err, "Configure: failed to get AWS credentials; please specify AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY")
 	}
 
+	if endpoint := os.Getenv("AWS_ENDPOINT"); endpoint != "" {
+		config.Endpoint = aws.String(endpoint)
+	}
+
+	s3ForcePathStyle, _ := strconv.ParseBool(os.Getenv("AWS_S3_FORCE_PATH_STYLE"))
+	config.S3ForcePathStyle = aws.Bool(s3ForcePathStyle)
+
 	region := os.Getenv("AWS_REGION")
 	if region == "" {
 		region, err = findS3BucketRegion(bucket, config)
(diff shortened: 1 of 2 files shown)


fdr:
Are there any tests you think we ought to add?

bithavoc:
@fdr can you give me an idea how could we test this? the Configure method doesn't return the inferred aws config where I could be checking the correct values have been assigned based on the environment variables, unless we have a private method build the aws config and we check it against environment variables.

fdr:
Might be nice to be able to connect environment variables to a properly set up struct somehow.

upload.go
```diff
@@ -82,6 +82,13 @@ func Configure() (*TarUploader, *Prefix, error) {
 		return nil, nil, errors.Wrapf(err, "Configure: failed to get AWS credentials; please specify AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY")
 	}
 
+	if endpoint := os.Getenv("AWS_ENDPOINT"); endpoint != "" {
+		config.Endpoint = aws.String(endpoint)
+	}
+
+	s3ForcePathStyle, _ := strconv.ParseBool(os.Getenv("AWS_S3_FORCE_PATH_STYLE"))
```
fdr: I see you ignore the error here...doesn't it need to be reported to the user somehow?

bithavoc: I don't think so, the default for `config.S3ForcePathStyle` is false, `strconv.ParseBool` parses correctly: `true`, `false`, `1`, `0`. other than `true`  or `1` is false which is a good default.

fdr: So if the user types "tru" we should interpret it as "false"? Seems not so great.

fdr: or the not entirely uncommon "on" vs. "off".

x4m:
This worked for me and is very use full for testing. Thanks.

tbartelmess:
This would be great

fdr:
Let's make the alteration to the error handling so typos or other renditions of "true" do not result in a "false" and commit it.

x4m:
I'll fix bool parsing in following commit.

PR #24, opened 2017-09-10
GPG encryption

x4m:
Hi! 
Here's encryption via Go implementation of OpenGPG.
It reads env var WALE_GPG_KEY_ID and asks gpg tool to export key. Than this key is feeded to OpenGPG.

Some things:
1. I use code from pull 16 for testing. It's very convenient to use local minio.io in testing. Let me drop these lines before the merge since this PR is about another feature
2. Same thing on Postgres10 compatibility. I can push that small thing in another PR.
3. There are few TODOs:
3a. Checking that S3 bucket exists (now we keep trying to upload). I'm not planning to fix this in this PR
3b. Cache key in user's HOME dir. Right now GPG is invoked on every run to obtain keys. I think it would be good to cache public keys to a file. I think it worth fixing in this PR.
3c. Support for GPG passphrases. Currently, only passwordless keys will go, but I do not see this as an important limitation. I'd like to fix this in another PR.
4. If you think there should be more tests, plz give me some ideas what else to test. Now encryption is embedded in all previous tests and one more test for encryption roundtrip introduced.

Any thoughts, ideas, and code design considerations are welcome.


x4m:
Here is commit addressing point 3B
https://github.com/x4m/wal-g/commit/5cb2c4fba9dcb2c39d447d65f75f881df90fd933
If everything is OK I can add it to this PR

x4m:
I've been reading through comments on HackerNews and got an idea: seems like different folks want different crypto. GPG has strong sides and has week sides. M.b. we could extract simple interface from Crypter, make crypto pluggable and provide some default implementations, like GPG?

What do you think about it?

fdr:
What second backend do people want? And why should we be so convinced we can design an abstraction without actually implementing at least a prototype of an alternative backend that enough people want, and are going to want for a long while? (WAL-E perpetually started to out-last its dependencies) The whole model of a "keyring", for example, may differ.

At least one use case, though, would be injection of an alternative no-op crypter for testing.

x4m:
I've implemented ExternalGPGCrypter for tests.
I need WAL-G to be direct replacement of WAL-E, so, before pushing, I plan to add tests with some small WALs actually compressed by WAL-E and more rigorous compatibility testing.
But. This test do not strictly require abstraction layer and I can drop it.
This abstraction does not incur performance penalty since interface only covert reader to reader and writer to writer. I've implemented this just in case we will be willing to implement libsodioum per this comment in HN https://news.ycombinator.com/item?id=15049691 .
If we have grounds to doubt necessity of other crypto backends I can rollback abstraction.

x4m:
I have rebased this feature distilling everything from temp fixes. 

Mock Crypter is used for test, but it is the only usage of interface abstraction. 

There are some WAL-E compatibility tests, by default they are skipped: I didn't find way to safely delete key from external GPG. These tests, if run, will install "walg-server-test" keys, to run compatibility tests user need to elevate trust of imported key. I can automate this part, but do not think it is good idea: adding trusted keys to every user running tests. Test of decryption of WAL-E  WAL run safely on Travis. All in all theses tests showed full compatibility with WAL-E encryption.

Now I think feature is commit-ready.

lehni:
Shouldn't `WALE_GPG_KEY_ID` be called `WALG_GPG_KEY_ID`?

x4m:
This setting was taken from WAL-E and was designed to make WAL-E->WAL-G transition as easy as possible.
But, probably, we can double this setting as WALG too...

PR #25, opened 2017-09-15
Postgres 10+ support

x4m:
Postgres 10 uses other naming convention for WAL (wal instead of xlog). Here I recheck version the same way WAL-E is checking.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -31,7 +31,20 @@ func Connect() (*pgx.Conn, error) {
 // fails.
 func StartBackup(conn *pgx.Conn, backup string) (string, error) {
 	var name string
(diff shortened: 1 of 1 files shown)


fdr:
There is probably a design that can work for offline backups, and a test would be nice though hard to arrange, but this is fine since otherwise PG10 support doesn't exist.

x4m:
I think I'll commit this with TODO. When I will get back to this I'll implement offline backups, incapsulate Pg version logic and create test specs for this logic.

How do you think, should I add 9.5 support? Required work is neglectible (after offline impl), but: 
1. We will have to support it for a long period of time
2. It will interfere with backups from replica
I can gist small patch with support for this functionality for those who need it, but I'm not sure it's viable solution for those who do not bother to upgrade Pg.

PR #29, opened 2017-10-01
Incremental backup

x4m:
Hello!
Here is incremental backup implementation. It has commands "backup-push-incr" and "backup-fetch-incr" with syntax identical to "backup-push" and "backup-fetch".
Backup follows this simple algorithm:
1. Fetch description of LATEST backup, if it is not incremental than fallback to full backup with LSN marker in json sentinel
2. Do backup as usual, but for page-organized Postgres files it scans file for pages with LSNs newer than previous incremental backup. 
Backup restoration process:
1. If current backup is based on other backup, first restore base
2. If has base, move base to increment_base dir
3. Restore itself, using increment_base dir
4. Remove increment_base dir

In future I'm planning to build map of changed pages with WAL parsing or hooks from this commitfest entry https://commitfest.postgresql.org/14/1273/

This PR also contains code from PR 16(only code, localized) and 25(full)


fdr:
I'm curious if you could elaborate on the ramifications of how to organize deletion code and interface. I think we need to figure that out before committing.

x4m:
> How to organize backup-delete?

Interface implemented in this PR was a draft, technological preview. Now I have committed most of prerequisites and going to work on this. Let's make more practical interface:

**backup-push** will look for 
- WALG_DELTA_MAX_STEPS: how many delta backups can be between full backups
- WALG_DELTA_ORIGIN: can be LATEST (chaining increments), LATEST_FULL (for bases where volatile part is compact and chaining has no meaning - deltas overwrite each other), or simply name of backup to look for (everything is almost like backup-fetch)

**backup-fetch** behavior is not affected externally.

**backup-list** will output everything we can extract without fetching sentinel: WALFILENAME and date (WALE columns name	last_modified	wal_segment_backup_start)

**backup-list --detail** _i'm not planning to implement this at this PR_ will fetch information on size, delta origin if any, stop WALfilename.

**delete** I will implement subcommand:
- retain [FULL|FIND_FULL] %number%: if FULL is specified keep 5 full backups and everything in the middle
- before [FIND_FULL] %name%
if FIND_FULL is specified WAL-G will calculate minimum backup needed to keep all deltas alive. If FIND_FULL is not specified and call can produce orphaned deltas - call will fail with the list.

pagefile.go
```diff
@@ -0,0 +1,374 @@
...
+// 8 bytes uint file size
+// 4 bytes uint changed pages count N
+// (N * 4) bytes for Block Numbers of changed pages
+// (N * BlockSize) bytes for changed page data
+//
+
+package walg
+
+/*
+// This block is CGo Postgres page header parsing. This is comment in comment. Cool.
```
fdr: Once again, I find myself doubtful of the utility of Cgo here given the simplicity of the code: https://github.com/golang/go/issues/16051, https://www.cockroachlabs.com/blog/the-cost-and-complexity-of-cgo/. Especially for a loop as inner as one-per-page, and given that it composes poorly with several other Go features.

x4m: OK I'll rewrite CGo code to manual parsing by binary.LittleEndian.Uint32
Anyway we have this bit-twiddling even in C
result.lsn = (((uint64_t)data->pd_lsn_h) << 32) + ((uint64_t)data->pd_lsn_l);
so 9 Go calls to parse page header will not be significantly worse. 

These two lines https://github.com/secwall/barman/blob/master/bin/barman-incr#L168 is equivalent to CGo code and seem prettier.

x4m:
> Can you qualitatively explain to me how you store the keys for the incremental backup? It'd be ideal if they lexically sorted adjacent to, but not interleaved with, the full backup it is a delta of. 

Currently delta backups has identical names to full backups. And looking at the name of sentinel you cannot understand is is delta or is it full. Should we change it?
Here is example
`server/basebackups_005/base_000000010000000000000077_backup_stop_sentinel.json`

Also, sentinel.json file stores information on LSN, ModTime of every file and name of the base backup for deltas.

x4m:
I've implemented delete.
Things to be done before commit:
1. `delete` arguments logic not tested, but can be tests
2. Docs
3. CGo page parsing
I hope to do this things next week, but i'm going to be a speaker on conference, not sure time slots will be large enough.

Things that will be left outside the scope of this PR:
1. backup-list --detail
2. WALG_DELTA_ORIGIN %backup_name%

fdr:
Hmm. I think embedding that information in the sentinel is reasonable. For one, it doesn't make new structure in the backup key names: were I given the opportunity to do that, I'd want to do several things at once. We do not foreclose the opportunity to change things later, and that is fine.

It does make it harder to delete dependent backups properly, but perhaps that's okay: simply not doing it and letting the useless deltas age out normally might be alright for most, as would be a simple greedy algorithm (forward-delete until a non-delta is found).

x4m:
All set and done. On Monday I plan to do stress-testing, run through some backup cycles and stare at the code for a while. And if no gut feelings will arise - commit it, then proceed to WAL prefetching.

benjamin-thomas:
Coming from https://github.com/wal-g/wal-g/blob/master/PostgreSQL.md

Could you please document `WALG_DELTA_MAX_STEPS` further. I fail to understand why it defaults to 0, is the code unstable?

It seems logical to me that anyone would want to use this feature, are there any particular things to take into consideration when using this?

x4m:
@benjamin-thomas usual deltas are stable.
Though you rarely need them.
Block-level deltas are there not to save space but to make restore faster.

I want to document the feature or make a blog post about it but still haven't done it...

benjamin-thomas:
Thanks for the feedback @x4m 

I'll gladly read your blog post once you find the time.

PR #32, opened 2017-10-13
Backups on replication slaves

x4m:
Hi!

Here's a sketch to make backups on replicas.
Since 9.6 pg_start_backup() is allowed to be called on replication slave, but pg_xlogfilename() is not.
This is done because this function uses timeline ID which have not meaning on slave. We simply cannot format WAL name for backup, because we cannot be sure in master's timeline.

Here I use LSN for backup name, may be it is better to use time or something like this.


fdr:
The problem is LSNs can collide in parallel timelines, and this may mess up lexical sorting of backups. It may be necessary to have the user specify the timeline number.

x4m:
I can read timeline from control file and then format it according to pg_walfilename() pattern.
Advantages:
1. It's more reliable than asking user
2. We do not bother user too much
Disadvantages:
1. It is not completely reliable
2. We didn't handle responsibility to user
3. There will be complicated code (though easily testable)

I'll investigate into dangers of picking stale timeline.

x4m:
I have implemented reading of timeline id from pg_control file. This new code is covered by tests, though I'm still considering sending a message to pgsql-hackers for comments on this approach.
I do not see any sufficient dangers.

x4m:
Here's the thread on @hackers https://www.postgresql.org/message-id/flat/BF2AD4A8-E7F5-486F-92C8-A6959040DEB6%40yandex-team.ru#BF2AD4A8-E7F5-486F-92C8-A6959040DEB6@yandex-team.ru

fdr:
I'd advise avoiding inline C. Is `pg_control` really so complex?

fdr:
It may be reasonable to wait for weigh-in by `-hackers` before proceeding, though.

x4m:
pg_control has plain but rich binary structure.
I've investigated portable ways to layout structure... and found none. C-compiled structure is the best and most self-describing thing.

I can read pg_control_data[12345:12345+4] (random offset here), this will be short but obscure. Instead I use code actually used to write pg_control file. But I do not insist: this offset will never be changed and we will have like 2 hundred less lines of code.

Michael Paquier joined thread on pgsql-hackers.

x4m:
Per discussion in pgsql-hackers, I see that one of two possible precautions must be implemented:
1. Invalidate backup if it was started on replica and finished on promoted master. This is necessary minimum.
2. Invalidate backup if timeline in pg_control at the beginning and at the end do not match. This guarantees that no timeline changing restartpoint arrived during copying of files. This restartpoint with new timeline could arrive from cascading replicas.

By invalidation I mean not uploading regular sentinel (m.b. upload invalidated.json?). If we had timeline change - than, probably, some bad things are happening, maybe some master is unreachable or dead, it's a good idea to keep a recent backup, but ensure it's proper use. Then I'll take into account these backups when implementing backup-delete.

I'm inclined to implement approach 2.

fdr:
I don't think we get much by inlining Postgres code, unless we need access to a bunch of fields, which we don't. Are there some versioning vields you can check against to make sure that we're parsing `pg_control` okay?

I think approach 2 is fine, one can omit the backup sentinel if there is a timeline switch, and then we're good.

x4m:
Using other fields seems like a good idea: I have a lot of unchecked assumptions on system constants. I'll use:
pg_control version number
Latest checkpoint's TimeLineID
Database block size
WAL block size
Bytes per WAL segment

Considering checking floatFormat for 1234567.0. It serves to another purpose but seems like we can treat this just like some signature bytes since IEEE 754.

x4m:
Well... we do not need all those hacks.

select timeline_id from pg_control_checkpoint()

It will read pg_control from disk, check CRC and do all other stuff.

x4m:
From my POV the last question here is: where to update docs?

fdr:
What do you mean, where? We could keep adding to README. There are some pros and cons to this. I would like to have the documentation under version control and distributed with the source rather than, say, a wiki. Yet, it should browse well in Github.

I don't know if there are better tools in terms of index generation/anchors/etc than github markdown that don't add too much toolchain.

x4m:
Sorry, I was not precise enough. I meant where in README.md?
I've placed some sentences after description of backup-push.

The code is easy part, but the docs are not. I think you have noticed that English if foreign for me, and, probably, it is somewhat pain to read my writings :) So, if you see some unclear or just clumsy words added to docs, please, do not hesitate to ping me. Docs are essential even with 100% compatibility with WAL-E.

PR #33, opened 2017-10-13
Fix backup name extraction in fetch LATEST

x4m:
Currently name is extracted from backup path so that backup can only reside in one subdir inside bucket.
This fix allows deeper nesting.


fdr:
Could a test be written?

x4m:
Sure, I'll compose test for correct parsing of different keys.

PR #34, opened 2017-10-13
Provide descriptive error on no LATEST found in backup-fetch

x4m:
Provide descriptive error on no LATEST found in backup-fetch instead of panic in return sortTimes[0].Name


fdr:
I like it.

PR #36, opened 2017-11-06
Add AWS S3 Server Side Encryption support.

Aevin1387:
Hello! 

This adds an environment variable to enable S3 Server Side Encryption support. The S3 manager handles the rest after the configuration option is set.

### DIFF
--- a/upload.go
+++ b/upload.go
@@ -144,6 +144,10 @@ func Configure() (*TarUploader, *Prefix, error) {
 		upload.SSEKMSKeyId = sseKmsKeyId
 	}
 
+	if serverSideEncryption == "aws:kms" && sseKmsKeyId == "" {
+		return nil, nil, errors.New("Configure: WALG_S3_SSE_KMS_ID must be set if using aws:kms encryption")
+	}
+
 	upload.Upl = CreateUploader(pre.Svc, 20*1024*1024, con) //default 10 concurrency streams at 20MB
 
 	return upload, pre, err
--- a/upload.go
+++ b/upload.go
@@ -144,8 +144,9 @@ func Configure() (*TarUploader, *Prefix, error) {
 		upload.SSEKMSKeyId = sseKmsKeyId
 	}
 
-	if serverSideEncryption == "aws:kms" && sseKmsKeyId == "" {
-		return nil, nil, errors.New("Configure: WALG_S3_SSE_KMS_ID must be set if using aws:kms encryption")
+	// Only aws:kms implies sseKmsKeyId
+	if (serverSideEncryption == "aws:kms") == (sseKmsKeyId == "") {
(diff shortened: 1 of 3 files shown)


fdr:
A nice feature to have. Is there any way a test could be written confirming that configuration is passed through to the struct correctly? IIRC we've struggled with the meaningfulness of such a test in the past, but do you have any ideas?

fdr:
Also, how does one specify particular KMS keys? I haven't used that feature but if we're going to add configuration complexity I want to understand its extent of completeness.

Aevin1387:
I'm sorry for the delay on this. I've finally gotten around to testing using KMS keys. To use a KMS key, a user would set `WALG_S3_SSE` to `aws:kms` and set `WALG_S3_SSE_KMS_ID` to the key ID.

Aevin1387:
@fdr When you have a chance, can you please take a look? I've tested this with both `AES256` and `aws:kms`, and added error handling if server side encryption is set to `aws:kms` and the KMS Key Id is not set.

upload.go
```diff
@@ -134,6 +134,20 @@ func Configure() (*TarUploader, *Prefix, error) {
...
+	if ok {
+		upload.ServerSideEncryption = serverSideEncryption
+	}
+
+	sseKmsKeyId, ok := os.LookupEnv("WALG_S3_SSE_KMS_ID")
+	if ok {
+		upload.SSEKMSKeyId = sseKmsKeyId
+	}
+
+	if serverSideEncryption == "aws:kms" && sseKmsKeyId == "" {
```
x4m: I think we should add some validation code here.
If user have sseKmsKeyId but no serverSideEncryption WAL-G should print an error, if serverSideEncryption is unknown, WAL-G should print warning.

upload.go
```diff
@@ -171,12 +185,22 @@ func (tu *TarUploader) upload(input *s3manager.UploadInput, path string) (err er
...
 		Bucket:       aws.String(tu.bucket),
 		Key:          aws.String(path),
 		Body:         reader,
 		StorageClass: aws.String(tu.StorageClass),
 	}
+
+	if tu.ServerSideEncryption != "" {
+		uploadInput.ServerSideEncryption = aws.String(tu.ServerSideEncryption)
+
+		if tu.ServerSideEncryption == "aws:kms" && tu.SSEKMSKeyId != "" {
```
x4m: SSEKMSKeyId should be setted only for ServerSideEncryption, right?
I'm a bit worried about checking string literal "aws:kms" here, validation logic is duplicated now... but probably it is OK.

x4m:
Cory, please review my changes. If everything is OK. let's merge it.

Aevin1387:
Just tested w/ your changes, looks good to me. :shipit:

x4m:
Thanks!

PR #37, opened 2017-11-06
GPG Encryption Fixes

Aevin1387:
This fixes two issues I found while trying to use the GPG encryption.

1) The GPG key and secret key exports were malformed, causing GPG to always return the first key in the list. 
2) On wal-fetch, the `OpenPGPCrypter` was not being initialized correctly


x4m:
Thanks!

PR #38, opened 2017-11-06
Add support for postgres 9.0 - 9.5

Aevin1387:
This adds support for 9.0-9.5 by modifying the `pg_start_backup` and `pg_stop_backup` commands based on version. 

I grabbed the stop backup query from `wal-e`, but from the testing I've done so far the file_name and offset aren't necessary. Happy to just remove that part of the query and just change it to SELECT *


fdr:
Tests, please. If at least to show the two strings and test that they are switched properly.

x4m:
Let me elaborate a bit on tests and string comparison. @fdr is asking for tests not to be sure that _your_ code works properly. Most probably, it does.
Tests are necessary for future maintenance.

Seems you have done important and tedious work to check Postgres specs on which call where and when should be used. I want to ask you to extract a type whose responsibility is to format queries (and at some point in future - to say that query should or should not be issued). This type can be initialized by Pg version number (for tests) or by connection (if you consider it worth moving querying in). For this type, there must be tests which check outputed queries.

connect.go
```diff
@@ -61,32 +61,22 @@ func (b *Bundle) StartBackup(conn *pgx.Conn, backup string) (backupName string,
...
-	if version >= 100000 {
-		walname = "wal"
-	}
-
-	query := "SELECT case when pg_is_in_recovery() then '' else (pg_" + walname + "file_name_offset(lsn)).file_name end, lsn::text, pg_is_in_recovery() FROM pg_start_backup($1, true, false) lsn"
-	if err = conn.QueryRow(query, backup).Scan(&name, &lsnStr, &b.Replica); err != nil {
-		return "", 0, version, errors.Wrap(err, "StartBackup: pg_start_backup() failed")
-	}
+	queryBuilder := PgQueryBuilder{}
+	queryRunner := PgQueryRunner{queryBuilder: queryBuilder, connection: conn}
```
dd-caleb: maybe create a `NewPgQueryRunner` function so that you don't have to make an empty `queryBuilder`:

~~~go
func NewPgQueryRunner(conn *pgx.Conn) *PgQueryRunner {
    return &PgQueryRunner{queryBuilder:&PgQueryBuilder{},conn:conn} 
}
~~~

And actually maybe it should do GetVersion in the constructor and return an error:

~~~go
func NewPgQueryRunner(conn *pgx.Conn) (*PgQueryRunner, error) {
    r := &PgQueryRunner{queryBuilder:&PgQueryBuilder{},conn:conn} 
    var err error
    r.Version, err = r.getVersion() 
    if err != nil {
        return nil, err
    }
    return r, nil
}
~~~

And make `GetVersion` to be `getVersion` so its private.

dd-caleb: (and you'd reference `queryRunner.Version` instead of `queryBuilder.Version`?

Aevin1387:
@fdr @x4m Sorry for the delay on this. I've heavily refactored the start/stop backup calls and added a query builder with tests. The new query runner utilizes the tested built queries, instead of building the queries directly. This also adds a transaction w/ statement timeout for the backup, as at least on the systems I was testing with, the stop backup could take longer than the statement timeout that we had set to a default.

x4m:
This. Is. Awesome.
I'm going to test and code review on Monday, hope to merge this ASAP.

pg_stop_backup() usually doing WAL switch and waits for archiving WALs. It must be a very overloaded server to wait for it too long (or some archive_command malfunction). But it is, indeed, possible. I need to think about it a bit...

Thanks for doing this!

x4m:
@Aevin1387  you rock, thanks!
@dd-caleb thanks for reviewing!

PR #43, opened 2017-11-28
Apply the correct offset into the list of backups when performing a '…

jnicholls:
…retain X' delete. Apply some gofmt cleanup on commands.go.


x4m:
Yes, that's bug. Thanks for fixing this.

PR #52, opened 2018-01-11
crypto: return an error if `gpg -a --export <key-id>` silenty fails

adamdecaf:
I ran into a situation where gpg -a --export <key-id> was failing and
gpg was printing an error to stderr. This resulted in an empty body
being stored in the cache and empty slice returned.

```
$ gpg -a --export D22173AC > /dev/null
gpg: WARNING: nothing exported

$ echo $?
0
```


x4m:
Thanks! Pushed. One extra good thing about this PR is small refactorings, they are welcomed. I'm sure WAL-G codebase can be better.

PR #53, opened 2018-01-11
upload: abort if WALE_S3_PREFIX doesn't look like a url

adamdecaf:


x4m:
Hi! Thanks for the pull request!
Are you sure this is the only protocol acceptable by Go SDK AWS? If so, surely your code is correct. What is the best way to check that?

adamdecaf:
> Are you sure this is the only protocol acceptable by Go SDK AWS?

I'm not sure about that.. It looks like minio cares more about the endpoint url being changed. How about we error if the Scheme || Host are empty strings? Both are required anyway. 

https://docs.minio.io/docs/aws-cli-with-minio
https://docs.minio.io/docs/golang-client-quickstart-guide
https://open.microsoft.com/2017/11/09/s3cmd-amazon-s3-compatible-apps-azure-storage/

x4m:
Sounds reasonable.
I've seen installations working with https+s3, but this prefix was not working for WAL-G.

adamdecaf:
Updated.

x4m:
Thanks, pushed.

PR #59, opened 2018-01-23
Add option to set path to gpg command

kamabe:
For some reason, Postgres wasn't finding `gpg` on my `PATH` during a restore, even though I could see it was there for the postgres user. This just lets the user set an explicit `gpg` path as an environment variable. If it's not set, it just falls back to the original `gpg` with no explicit path set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,9 @@ To configure GPG key for encryption and decryption. By default, no encryption is
 
  To configure base for next delta backup (only if `WALG_DELTA_MAX_STEPS` is not exceeded). `WALG_DELTA_ORIGIN` can be LATEST (chaining increments), LATEST_FULL (for bases where volatile part is compact and chaining has no meaning - deltas overwrite each other). Defaults to LATEST.
 
+* `GPG_BIN_PATH`
+
+To configure the path to the `gpg` command. If not set, it defaults to `gpg` and is assumed to be on the PATH.
 
 Usage
 -----
--- a/crypto.go
+++ b/crypto.go
@@ -139,7 +139,16 @@ func GetKeyRingId() string {
 	return os.Getenv("WALE_GPG_KEY_ID")
 }
 
-const gpgBin = "gpg"
+func GetGpgBinPath() string {
+	gpgBinPath := os.Getenv("GPG_BIN_PATH")
+	if gpgBinPath == "" {
+		return "gpg"
+	}
+
+	return gpgBinPath
+}
+
+var gpgBin = GetGpgBinPath()
 
 type CachedKey struct {
 	KeyId string `json:"keyId"`


x4m:
Hi, Kayla! Thanks for the pull request! It is always great to see community effort on making WAL-G better.

Indeed, the problem you raised is important and everyone is solving it with different workarounds. It seems to me that PostgreSQL does not path PATH, neither to archive_command nor to restore_command. In my environment, I add PATH manually to envdir settings. But this is ugly, I do not like it, but have not enough resources to make it better.

Your solution is better. But I have few concerns:
1. Right now WAL-G uses only one external tool - gpg. Will it hold forever? If not, we could rename GPG_BIN_PATH -> BIN_PATH. Actually, I do not expect adding more dependencies. I expect removing external gpg as a dependency: once in a while we will teach WAL-G to read keychain without external GPG.
2. Thing that bothers me most: I do not know how WAL-E handles this PATH issue. WAL-E uses a lot of external tools.

May I ask you to investigate 2nd point and conclude which solution is best for developmement and usability of WAL-G?

Thank you!

adamdecaf:
If we're looking to remove the dependency on `gpg` as a binary that'd be cool. Are you thinking of reading the armor data from [x/crypto/openpgp](https://godoc.org/golang.org/x/crypto/openpgp)?

kamabe:
So here's what [`wal-e` does for the `gpg` binary](https://github.com/wal-e/wal-e/blob/cdfaa92698ba19bbbff23ab2421fb377f9affb60/wal_e/pipeline.py#L12). Basically the same thing as what is done in wal-g. No override options there either.

What seems to happen in `wal-e` to account for the PATH is that is [copies the environment](https://github.com/wal-e/wal-e/blob/b86998b0db1bb4c9999f856ecea98af6da974118/wal_e/worker/pg/psql_worker.py#L45) before calling out to another process. We could do the same thing here if we wanted.

x4m:
@adamdecaf Definitely, removing dependency is a good thing here. We are already reading armored data [exactly by opengpg](https://github.com/wal-g/wal-g/blob/master/crypto.go#L64)
We use external gpg only as a storage to preserve WAL-E behavior. We call GPG to export armored keys, we could just read this keys as gpg does.

@kamabe I think I didn't understand you. Did you added [here](https://github.com/wal-g/wal-g/blob/master/crypto.go#L189) and [here](https://github.com/wal-g/wal-g/blob/master/crypto.go#L167) `cmd.Env = os.Environ()` ? Or did you pass PATH like `PATH=/usr/bin wal-g wal-fetch ....`?

stub42:
wal-e and wal-g get the same path as the parent postgres process. This is often not the same as the postgres user's environment, but instead defined in the systemd service file (or similar for other startup mechanisms). eg. under Ubuntu, there is no environment set in the /lib/systemd/systemd/postgres*.service files, and no service override files in /etc/systemd/system/postgresql@.service.d, so the default system environment from /etc/environment is used. This might explain some of the oddities seen when testing, and provides alternatives to overriding PATH in the envdir.

stub42:
@x4m I think the issues are orthogonal.

The patch on this PR is providing a way to specify the path to the GPG executable, for example if you need to use /usr/bin/gpg2 instead of /usr/bin/gpg . I do not think it is a good way of removing the  dependency on GPG, as using this patch the only thing that dependency could be replaced with is an external executable with the same command line arguments as GPG. I think removing the dependency on GPG should be a separate proposal.

I don't think there is any need for WAL-G to do better PATH handling, as there are already at least two places the PATH can be overridden (envdir, specific to WAL-G, and the postgres parent process' environment, specific to all processes spawned by PostgreSQL). The patch may still be useful though, for those cases where the GPG executable we want to run is not the first executable called 'gpg' found in the path (such as Ubuntu 16.04, where /usr/bin/gpg is GPG v1 and /usr/bin/gpg2 is GPG v2)

kamabe:
@x4m Ah I missed adding it to the secret key piece. I only added it to the public export. So it may very well work just fine. I didn't even think about the postgres init script potentially setting the path though. I also didn't think about my PR being applicable in situations where people use different versions of gpg. I don't know how often people have that issue. 

It seems like there are a lot of solutions to this problem, and I'm fine closing this PR if another solution is deemed better. I'm totally fine setting the PATH in envdir for right now, as that was my solution before this PR.

adamdecaf:
👍 From me. 

I've seen go strip env vars (including `PATH` and `HOME`) with `os/exec.Command`.

```go
$ cat /tmp/env.go 
package main

import (
	"fmt"
	"os"
	"os/exec"
)

func main() {
	env := "PATH"
	fmt.Printf("os.Getenv(%s) = %q\n", env, os.Getenv(env))

	// fmt.Sprintf("env %s", env)
	out, _ := exec.Command("echo $PATH").CombinedOutput()

	fmt.Printf("exec.Command PATH = %q\n", string(out))
}
```

```
$ go run /tmp/env.go 
os.Getenv(PATH) = "/usr/local/opt/asdf/bin:/usr/local/opt/asdf/shims:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/adam/code/bin:/Users/adam/bin"
exec.Command PATH = ""
```

x4m:
@adamdecaf , from your snippet I conclude that it's necessary to ```cmd.Env = os.Environ()```. But why then ```PATH=/usr/bin wal-g``` works?

adamdecaf:
I'm not sure.

x4m:
This seems to be adressed. If you have some more question plz reopen.

PR #66, opened 2018-02-11
Fix file close in increment

ferhatelmas:
- first check error on opening file and defer closing it only if there is no error


x4m:
Thanks a lot!

PR #69, opened 2018-02-15
General code simplifications

ferhatelmas:
* unnecessary bool constant and error check

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -1,6 +1,7 @@
 package walg
 
 import (
+	"encoding/binary"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


commands.go
```diff
@@ -481,10 +481,8 @@ func GetDeltaConfig() (max_deltas int, from_full bool) {
 	if hasOrigin {
 		switch origin {
 		case "LATEST":
-			break
```
x4m: Are you sure it's OK to remove this?

x4m:
Thanks! Most of refactorings look great!
I just have a question regarding switch\break, is it really ok to do so?

commands.go
ferhatelmas: Yes, it's fine. https://play.golang.org/p/e4thhtWzYcw

x4m: Cool

x4m:
Thanks, pushed

PR #71, opened 2018-02-19
Defer after checking error

ferhatelmas:
If err isn't nil, deferring will panic so we need to check error and defer only upon success.


x4m:
Many thanks!
I'll draw myself a sticker reminding not to defer that way :)

PR #75, opened 2018-03-05
Fix base-backups on Postgres 9.0-9.5

Aevin1387:
Fixes base-backups on Postgres 9.0-9.5 by skipping the steps to write the BackupLabel and tablespace map. The backup_label is written automatically, and tablespace map does not exist in < 9.6. This got dropped in the refactor of https://github.com/wal-g/wal-g/pull/38 and I forgot to re-test restoring the backup (I only tested generation)

### DIFF
--- a/upload.go
+++ b/upload.go
@@ -352,6 +352,10 @@ func (bundle *Bundle) HandleLabelFiles(conn *pgx.Conn) (uint64, error) {
 		return 0, errors.Wrap(err, "HandleLabelFiles: failed to parse finish LSN")
 	}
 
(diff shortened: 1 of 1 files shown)


x4m:
@Aevin1387 thanks for keeps things working! :)

BTW, how do you think could we set up e2e like here https://travis-ci.org/petergeoghegan/amcheck/builds/349447523?utm_source=github_status&utm_medium=notification ?

If we support so many versions, we, actually, have to check stability against all of them...
To do that we, probably, need an FS storage support, but @tamalsaha is kind of working on this already.

PR #76, opened 2018-03-05
Add extract function.

Aevin1387:
Adds a command to extract a file created by `wal-g`. I've noticed that while `wal-fetch` works correctly, if I try to decrypt + decompress one of the `.lz4` files manually, the decompression fails with an error: 

```
Error 77 : Decoding Failed ! Corrupted input detected !
```

### DIFF
--- a/cmd/wal-g/main.go
+++ b/cmd/wal-g/main.go
@@ -18,6 +18,7 @@ var helpMsg = "  backup-fetch\tfetch a backup from S3\n" +
 	"  backup-list\tprints available backups\n" +
 	"  wal-fetch\tfetch a WAL file from S3\n" +
 	"  wal-push\tupload a WAL file to S3\n" +
+	"  extract\textract a local WAL file\n" +
 	"  delete\tclear old backups and WALs\n"
 
 func init() {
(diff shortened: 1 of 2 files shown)


x4m:
I do not see what is the problem that you are solving...
Can I ask you to elborate a bit? Does WAL-G fail to fetch your WAL?

Aevin1387:
We run a separate process to verify the uploaded WAL files. This process downloads the WAL files, decrypts & extracts them, then uses `pg_xlogdump` to verify them. With `wal-e`, we were able to use `gpg` and `lzop` directly to extract the compressed and encrypted wals, but when we try to use `gpg` and `lz4` against the files uploaded by `wal-g`, `lz4` fails with the error above.

x4m:
OK, I see, verification is a very important thing.
And the Idea about pg_xlogdump is pretty neat.
I do not understand why you want to use external download tool and not just wal-fetch. What kind of download functionality is missing for you? WAL listing? 

IMO the best option would be to verify WAL right there in memory during download. But I see it is quite complicated code in pgxlogdump\pg_waldump.

Final verification would be that we can restore basebackup and all subsequent WALs can be applied, and the data can be fetched after it...

davidfetter:
Bump

Aevin1387:
Closing this because using `wal-fetch` works well enough for our use-case

Issue #82, opened 2018-03-26
Issues with parallel backup-push


Aevin1387:
I've been testing out the parallel backup-push that is currently in master, and have run into a few breaking scenarios. 

First, and foremost, if the backup doesn't crash it never finishes. It gets through the last files, and seems to freeze while waiting for the last few to upload. Example:
```
...
/pg_xlog
/recovery.done
/server.crt
/server.key
Finished writing part 78.
```

Secondly, I've run into a couple of crashes, details of which can be found in [this gist](https://gist.github.com/Aevin1387/d6d30cb23c9a934aa38eed87a0800d8d). I've hit the "Wait Group" exception the vast majority of the time, and the "concurrent map iteration and map write" a handful of times. 

I've not yet been able to track down the cause of any of these issues, but I haven't spent a lot of time looking.

x4m:
@Aevin1387 thanks for your report. Seems like I have to remove parallel backup from master and make pull request out of it to do more testing.

x4m:
This seems to be fixed as for now

PR #83, opened 2018-03-27
Implement parallel backup-push

x4m:
Currently, backup-push uses only one goroutine to read from disk. This PR solves this problem and makes backup-push 1.5x faster according to my load tests.
But I have two concerns about it:
1. Each pipeline consumes ~400MB of RAM. So, with default WALG_MAX_UPLOAD_CONCURRENCY RAM usages is increased by the factor of 10, this seems unacceptable.
2. @Aevin1387 discovered some bugs #82. I suspect that these bugs manifest on PG versions before 9.6 but further investigation is necessary
3. This PR makes disk throttling a necessary feature because it somewhat guarantees exhausting either disk io throughput or network throughput.

I plan to make this parallelism controllable by some env switch and make it off by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,10 @@ To configure how many goroutines to use during backup-fetch  and wal-push, use `
 
 To configure how many concurrency streams to use during backup uploading, use `WALG_UPLOAD_CONCURRENCY`. By default, WAL-G uses 10 streams.
 
+* `WALG_UPLOAD_DISK_CONCURRENCY`
+
+To configure how many concurrency streams are reading disk during ```backup-push```. By default, WAL-G uses 1 stream.
+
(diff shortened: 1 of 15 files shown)


x4m:
Hi, David!

There was a concurrency bug that @Aevin1387 reported, I've just fixed it.
Also, I will do disk parallelism configurable.
Then I'll make a pre-release and perform QA cycle in Yandex test environment.
Then I plan a release.

Thanks for your interest in the topic :)

x4m:
OK, I've implemented the queue for still-uploading-tars which waits when overflown. This helped to control RAM usage on both single-stream read and multi-stream read of data files. I do not observe races now and version seems stable. I will merge this and proceed to prepare the prerelease.

PR #91, opened 2018-04-20
Remove leftover references to WAL-E

davidfetter:
This is a separate project now. It should have its own names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Configuration
 
 To connect to Amazon S3, WAL-G requires that these variables be set:
 
-* `WALE_S3_PREFIX` (eg. `s3://bucket/path/to/folder`)
+* `WALG_S3_PREFIX` (eg. `s3://bucket/path/to/folder`)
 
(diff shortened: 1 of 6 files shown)


x4m:
Thanks, David!

I like the idea. But I have two concerns:
1. This will break existing installations
2. WAL-G is designed to be an easy replacement for WAL-E
3. For example, TestDecryptWALElzo -> TestDecryptWALGlzo is conceptually incorrect, WAL-G does not use Lzo for it's files, Lzo support is here for backward compatibility. You can switch to WAL-G from running WAL-E cluster by configuring your replicas first. WAL-G can wal-fetch files of WAL-E.

x4m:
I'm boarding on a plane right now, will respond on other your items after arrival. Many thanks for your work!

davidfetter:
How about a backward-compatibility mode?

x4m:
I think that it is a good idea to have all "WALE_*" settings backed by corresponding "WALG_*" settings, and then leave small note that WALE settings are supported too, removing them from the main docs section. Sorry for answering so long.

x4m:
@davidfetter may I ask you to look into https://github.com/x4m/wal-g/commit/4eaf7f65da42af8f88372009a4e3d81292c96487 ?
How do you think, is it what we want here? Can we make it better?

x4m:
Thanks! This work was partially merged in #111

PR #95, opened 2018-05-08
Compare file modification times using Equal method

CyberDem0n:
Go == operator compares not just the time instant but also the Location and the monotonic clock reading: https://golang.org/pkg/time/

Using == was causing reading of all files despite no change of MTime.


x4m:
Wow, that makes sense. Thanks!

Issue #97, opened 2018-06-06
Alternative to lz4 compression algorithms


x4m:
Currently, WAL-G significantly outperforms WAL-E in CPU and disk-constrained environments (especially with delta backups, but that's another story). But when the network throughput is a bottleneck - WAL-E is faster. WAL-G backups are simply bigger since it uses lz4 - faster but less efficient compression algorithm (than lzo).
Meanwhile, In many cases, CPU usage of WAL-G is neglectable small. Many CPU cores are simply sleeping, and it worth utilizing them to reduce usage of network bandwidth and storage size.
Which compression formats could be a most viable alternative?
It is highly preferred to use Go implementation with broad usage.

dev1ant:
Zstd?

x4m:
At first glance, I do not see Go implementations, only C-Go bindings. But I think it worth trying.

patsevanton:
wal-g support zstd?

AlexGluck:
How to disable compression? Anyone knows?

Issue #92, opened 2018-04-20
Remove dependency on external GPG


davidfetter:
I'm [not the first person](https://github.com/wal-g/wal-g/pull/59) to have trouble on this score. It seems a lot cleaner just to cut the ties to `gpg` implementations and bundle the functionality with the binary.

x4m:
OK, I think I should do that eventually.
Let's discuss some design decision.

Now WAL-G reflects the functionality of WAL-E: reads env var with key id and asks gpg for keys by id.

I can do:
1. Load keys from env var directly
2. Read gpg key store and extract keys myself

The first approach seems much cleaner. Second, preserves backward compatibility and get reed of gpg totally. First is easy to code with next to none place for bugs.

What do you think?

davidfetter:
I think getting rid of gpg totally is a large part of the point of this exercise.

While keeping it easy to migrate from `wal-e` may be a nice-to-have, it's not something on which it's really worth spending a lot of engineering cycles at this stage of the project. There are already settings `wal-g` has that `wal-e` does not, and as the project matures, there will be more.

To see why having an external dependency on `gpg` is a bad idea, try running `wal-g` on an Ubuntu system using only Ubuntu tooling, i.e. using `pg_ctlcluster`, etc. Oh, and that `gpg` on such systems is not the latest--that's `gpg2`--is sooner rather than later going to be a security issue if it isn't already.

All this is a long-winded way of saying that we should concentrate on the following in the following order of priority:
 
1. Remove the `gpg` dependency while preserving the current usage
2. Figure out better ways to work with stores, as ease of use would dictate.

dragonsmith:
@x4m Hey! Thanks for improving wal-g!

I think the idea to use a default gpg key store is a bit better overall than to store keys in env vars or files - you can manage them with a ready cli tool and use these keys for any other encryption processes. The problem is that the default gnupg package can be still too old (version 1....) and v2 key store is incompatible with v1. It raises the new choice: what key store version should wal-g use by default... Dunno what to suggest with it.

x4m:
@dragonsmith thanks!
As far as I know, @Savichev-Igor was doing some work on organizing crypto around WAL-G. Though I do not know what's the current state of his research.

davidfetter:
With utmost respect, the reason to ship `wal-g` with its own way to handle keys in the first place is that OSs frequently come with wrong versions of it. The `go` ecosystem is tuned up to avert problems of this type if we just work with it.

[...]

Issue #35, opened 2017-10-30
Some LZOP archives don't work


fdr:
I'm still taking this apart but it can be related to #22.

I have this backup that has hundreds of archives, but one of those archives causes a systematic crash, whereas `lzop` seems fine with it. There's not much to do but for me to go through it with a fine-tooth comb, but, FYI.

cc @x4m

x4m:
Is this just one file? Can you share it with me? I'd like take a look.

fdr:
On Mon, Oct 30, 2017 at 10:19 PM Andrey Borodin <notifications@github.com>
wrote:

> Is this just one file? Can you share it with me? I'd like take a look.
>

Nope. Confidential data.

fdr:
Okay. It's a bug, and not the first one, in the rasky pure-go lzo library. The C binding version works perfectly. At this point I have to decide whether I want to involve a C dependency (probably making it optional via build tag) or fix the rasky go-lzo library.

fdr:
I think I'm going to go the route of binding the C lzo library via build tag. LZO support is basically transitional anyway, and sinking time into fixing the pure-go implementation may not be worth it.

PierreDucroquet:
I'm facing a similar issue. I tried replacing wal-e by wal-g for a restore job and got issues with some wal files that could not be uncompressed by wal-g while lzop accept them happily.
I can not share the file for privacy/legal reasons, but I would be happy to help testing new code if needed.

x4m:
@PierreDucroquet , do you use GPG encryption? Can you share error message?
If you have a message "DecompressLzo: did not fill skip", probably, it is another issue.

Just today I've encountered an error, which manifests oftener with GPG encryption: here's the call https://github.com/wal-g/wal-g/blob/master/decompress.go#L46 which must use ioutil.ReadFull(). Hope to fix this week.

PierreDucroquet:
Yep, we have GPG encryption and it had another issue I worked around.

Basically, my patched function looks like : 
```go
func DecompressLzo(d io.Writer, s io.Reader) error {
        skip := 33
        sk := make([]byte, skip)

        n, err := s.Read(sk)
        full_skip := n
        if n < skip {
                for {
                        _sk := make([]byte, skip - full_skip)

                        n, err = s.Read(_sk)
                        full_skip += n
                        if full_skip == skip {
                                break
                        }
                }
        }
        if full_skip != len(sk) {
[...]

PR #99, opened 2018-06-26
Lzma and zstd

Tinsane:
Refactored a bunch of code related to compression.
Added lzma and zstd compression methods.


x4m:
Cool! Thanks!
@CyberDem0n Alexander, Hi! As far as I remember you was interested in #97 , can you look into this? Does it bring some more performance on your clusters? Vladimir observes 3x better compression with roughly same CPU consumption. @Tinsane is these numbers correct?

Also, I'll rebase branch on a current master.

compression.go
```diff
@@ -0,0 +1,50 @@
...
+	FileExtension() string
+}
+
+type Decompressor interface {
+	Decompress(dst io.Writer, src io.Reader) error
+	FileExtension() string
+}
+
+var Compressors map[string]Compressor
+var Decompressors = []Decompressor{
```
x4m: Not a show stopper, but different styles of initializations look ugly.

compression_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	err = compareDecompressedFileWithInitial(initialFilePath)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+}
+
+func TestSmallFileCompression(t *testing.T) {
+	for _, compressingAlgorithm := range compressingAlgorithms {
```
x4m: How about adding lzop here too? If it is no a lot of work, of course. I know that it is somewhat broken #35 butI still hope to find an example of broken file.... in this case we should leave broken file even in an event of failure.

crypter.go
```diff
@@ -2,7 +2,7 @@ package walg
 
 import "io"
 
-// Crypter is responsible for makeing cryptographical pipeline parts when needed
+// Crypter is responsible for making cryptographical pipeline parts when needed
 type Crypter interface {
```
x4m: Currently, we have compressor and decompressor. How do you think, is it viable to split crypter into encryptor and decryptor? They share keyring infrastructure and we have an issue about them #92, maybe keepeing them together for now is a good idea.

extract.go
```diff
@@ -45,36 +46,36 @@ func extractOne(ti TarInterpreter, s io.Reader) error {
...
+	for _, decompressor := range Decompressors {
+		if readerMaker.Format() != decompressor.FileExtension() {
+			continue
 		}
-	} else if readerMaker.Format() == Lz4FileExtension {
-		_, err = DecompressLz4(writeCloser, r)
+		err = decompressor.Decompress(writeCloser, readCloser)
 		if err != nil {
-			return errors.Wrap(err, "ExtractAll: lz4 decompress failed. Is archive encrypted?")
+			return errors.Wrap(err, fmt.Sprintf("ExtractAll: %v decompress failed. Is archive encrypted?", decompressor.FileExtension()))
```
x4m: How about errors.Wrapf? If it's our package

s3_tar_ball.go
```diff
@@ -1,22 +1,20 @@
...
-	"github.com/aws/aws-sdk-go/aws"
-	"bytes"
 )
 
 // S3TarBall represents a tar file that is
 // going to be uploaded to S3.
 type S3TarBall struct {
-	baseDir          string
 	trim             string
 	bkupName         string
```
x4m: I'm surprised this didn't trigger you :)

s3_tar_ball.go
```diff
@@ -33,14 +31,14 @@ type S3TarBall struct {
...
-// the form `part_....tar.lz4`.
+// the form `part_....tar.[Compressor file extension]`.
 func (tarBall *S3TarBall) SetUp(crypter Crypter, names ...string) {
 	if tarBall.tarWriter == nil {
 		var name string
 		if len(names) > 0 {
 			name = names[0]
 		} else {
-			name = "part_" + fmt.Sprintf("%0.3d", tarBall.partCount) + ".tar." + Lz4FileExtension
+			name = "part_" + fmt.Sprintf("%0.3d", tarBall.partCount) + ".tar." + tarBall.tarUploader.compressor.FileExtension()
```
x4m: How about tarFileExtension? If it's not overkill

utility.go
```diff
@@ -142,11 +146,39 @@ func getMaxConcurrency(key string, default_value int) int {
...
+			con = defaultValue
 		} else {
 			con = 10
 		}
 	}
 	return max(con, 1)
 }
+
+// CheckType grabs the file extension from PATH.
+func CheckType(path string) string {
```
x4m: It's not CheckType() it's GetExtensions()
chances are regex can be part of the global state (static)

x4m:
I'll do one more pass through code tomorrow

compression.go
Tinsane: Fixed

s3_tar_ball.go
Tinsane: Fixed :)

extract.go
Tinsane: Fixed

utility.go
Tinsane: Fixed

s3_tar_ball.go
Tinsane: Fixed

x4m:
I think it's fine to merge now. But we need README update. It's good to have some advice which one user should choose. Please, add your benchmarks and suggestions when user should switch to zstd (network limitations, some spare CPU) and when to lzma (storage and network limitations, a lot of unused CPU).

x4m:
Thanks, pushed.

Issue #104, opened 2018-07-24
Do Not Overwrite WAL Files


DustinChaloupka:
Reading the PostgreSQL documentation related to archiving, we came across this (from https://www.postgresql.org/docs/10/static/continuous-archiving.html):

>The archive command should generally be designed to refuse to overwrite any pre-existing archive file. This is an important safety feature to preserve the integrity of your archive in case of administrator error (such as sending the output of two different servers to the same archive directory).
>
>It is advisable to test your proposed archive command to ensure that it indeed does not overwrite an existing file, and that it returns nonzero status in this case. The example command above for Unix ensures this by including a separate test step. On some Unix platforms, cp has switches such as -i that can be used to do the same thing less verbosely, but you should not rely on these without verifying that the right exit status is returned. (In particular, GNU cp will return status zero when -i is used and the target file already exists, which is not the desired behavior.)

So we checked to see if `wal-g` did anything that would prevent it from overwriting files, but looking at https://github.com/wal-g/wal-g/blob/master/tar_uploader.go#L96 it does not appear that it does anything, nor any configuration being available that might help.

Is there something we might be missing? If not, would it be fine for something like that to be added around the upload? Another option would be to have it as a separate command that can be used in the `archive_command` for the PostgreSQL configuration, but then that means doing some command line checks with it.

As a workaround, it's probably possible to just manually check the path with say the `aws` CLI tool, but that would be kinda brittle with [wal-g defining the destination path](https://github.com/wal-g/wal-g/blob/master/tar_uploader.go#L84).

x4m:
That's true, WAL-G overwrites archive. We (me and my colleagues) already had some trouble with this. I will fix this (or happily review PR if you have some suggestions). Thanks for taking the time to test it and report with such a careful analysis.

DustinChaloupka:
#105 Is a try at it.

PR #105, opened 2018-07-24
Check WAL file existence before upload

DustinChaloupka:
Issue: https://github.com/wal-g/wal-g/issues/104

A first pass attempt at not overwriting WAL files.

I did not see any way to set up unit tests for this. The `commands_test.go` file looked to just do parsing tests.

Manually building this and testing it did work though.

```
$ wal-g wal-push /var/lib/postgresql/10/main/pg_wal/000000010000000000000001
BUCKET: mybucket
SERVER: myserver
WAL PATH: myserver/wal_005/000000010000000000000001.lz4
$ wal-g wal-push /var/lib/postgresql/10/main/pg_wal/000000010000000000000001
BUCKET: mybucket
SERVER: myserver
2018/07/24 21:47:03 WAL file '/var/lib/postgresql/10/main/pg_wal/000000010000000000000001' already archived, not overwriting
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -604,6 +604,19 @@ func HandleWALPush(tarUploader *TarUploader, dirArc string, pre *S3Prefix, verif
 
 // UploadWALFile from FS to the cloud
 func UploadWALFile(tarUploader *TarUploader, dirArc string, pre *S3Prefix, verify bool) {
+	archive := &Archive{
+		Prefix:  pre,
+		Archive: aws.String(sanitizePath(tarUploader.server + WalPath + filepath.Base(dirArc) + "." + tarUploader.compressor.FileExtension())),
+	}
+
+	exists, err := archive.CheckExistence()
+	if err != nil {
+		log.Fatalf("FATAL %+v\n", err)
+	}
+	if exists {
+		log.Fatalf("FATAL WAL file '%s' already archived, not overwriting", dirArc)
+	}
+
 	path, err := tarUploader.UploadWal(dirArc, pre, verify)
 	if re, ok := err.(CompressingPipeWriterError); ok {
 		log.Fatalf("FATAL: could not upload '%s' due to compression error.\n%+v\n", path, re)
--- a/bguploader_test.go
+++ b/bguploader_test.go
@@ -1,9 +1,11 @@
 package walg_test
 
 import (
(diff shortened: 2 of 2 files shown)


commands.go
```diff
@@ -593,6 +593,20 @@ func DownloadAndDecompressWALFile(pre *S3Prefix, walFileName string, dstLocation
...
+		Archive: aws.String(sanitizePath(tarUploader.server + WalPath + filepath.Base(dirArc) + "." + tarUploader.compressor.FileExtension())),
+	}
+
+	exists, err := archive.CheckExistence()
+	if err != nil {
+		log.Fatalf("%+v\n", err)
+	}
+	if exists {
+		log.Printf("WAL file '%s' already archived, not overwriting", dirArc)
+		return
```
rubbish: Will this return a non-zero exit code? We'll want that so postgres knows the archive_command failed and will keep the WAL file around.

x4m:
Is there a way to say to S3: "do not allow to overwrite this file"?

commands.go
DustinChaloupka: It exits with status `1` now instead.

x4m:
Thanks for doing WAL-G better! I really appreciate your work on the project.

Things are a bit more complex.... WAL-G, invoked as archive_command can also get information from archive_status: it reduces startup cost and makes WAL archive parallel.
But in this PR you check the only file of archive_command, not those files which were stealed from archive_status by background uploader.

DustinChaloupka:
@x4m The check has been moved to be included in the path of the `BgUploader` uploading WAL files as well.

x4m:
Ugh...it breaks tests. If you have some spare time plz fix it. If you wish I can fix it myself tomorrow (UTC+5) and merge PR.

x4m:
https://travis-ci.org/wal-g/wal-g/jobs/408531439

DustinChaloupka:
@x4m I actually added a test for the `bg_uploader` to check that the overwrite does not happen as well.

x4m:
`cmd := exec.Command(os.Args[0], "-test.run=TestBackgroundNoOverwriteWALUpload")`
whoa, hacky, tricky, but works.... testing panic... OK, I'll add comment there.

Thank you, pushed!

PR #109, opened 2018-08-08
Move WAL overwrite under switch and add precaution checks

x4m:
@DustinChaloupka, @fdr plz review this PR, mainly design decisions. I'll add docs later.

Here we check WAL contents if WAL-G tries to overwrite WAL.

I've tried MD5==ETag thing, it does not work for gpg-encrypted files, because the file is different each time.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -18,6 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/pkg/errors"
 	"sync"
+	"bytes"
 )
 
 // HandleDelete is invoked to perform wal-g delete
@@ -544,7 +545,7 @@ func tryDownloadWALFile(pre *S3Prefix, walFullPath string) (archiveReader io.Rea
 	return
 }
 
-func decompressWALFile(archiveReader io.ReadCloser, dstLocation string, decompressor Decompressor) error {
+func decompressWALFile(archiveReader io.ReadCloser, file io.WriteCloser, decompressor Decompressor) error {
 	crypter := OpenPGPCrypter{}
 	if crypter.IsUsed() {
 		reader, err := crypter.Decrypt(archiveReader)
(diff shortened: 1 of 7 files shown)


commands.go
```diff
@@ -617,6 +610,42 @@ func UploadWALFile(tarUploader *TarUploader, dirArc string, pre *S3Prefix, verif
...
+		if err != nil {
+			log.Fatalf("FATAL: %+v\n", err)
+		}
+
+		if !bytes.Equal(archived.Bytes(), localBytes) {
+			log.Fatalf("FATAL: WAL file '%s' already archived, contents differ, unable to overwrite\n", dirArc)
+		} else {
+			log.Printf("WARNING: WAL file '%s' already archived, archived content equals\n", dirArc)
+		}
+	}
```
DustinChaloupka: This function looks fine design-wise, the only thing that might potentially be a worry is how much slower this makes the uploads.

archive.go
```diff
@@ -15,24 +15,29 @@ type Archive struct {
...
-				return false, nil
+				return false, headResult.ETag, nil
 			default:
-				return false, awsErr
+				return false, headResult.ETag, awsErr
 			}
 		}
 	}
-	return true, nil
+	return true, headResult.ETag, nil
```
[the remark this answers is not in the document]
x4m: yeah, that's what left from MD5 experiments, will delete these lines.

commands.go
x4m: well, uploads are reversed: you download and compare. I didn't benchmark though, I think it is not a normal situation when you try to overwrite WAL, even if with the same data.

x4m:
@Tinsane brought my attention to one more thing: we should not report error (with nonzero exit code) if overwrite is attempted from bgUploader, only if we encounter overwrite with main file.

PR #110, opened 2018-08-09
Wal delta

Tinsane:
Implemented faster delta backups using WAL files. Now, instead of full database scan, a user can create and upload to S3 special delta files, which contain all DB changes and are used to get changed pages during delta backup.


x4m:
I've merged it to enable further development and avoid a lot of merges.
There is a known problem with WAL parsing in parallel. This does not lead to broken backups but triggers a fallback to default delta-method. This will be fixed in future patches.

PR #111, opened 2018-08-10
Read config file and back WALE settings with WALG settings

x4m:


fdr:
Drop-in compatibility was not really a goal of mine, unless a compelling interest made exact compatibility a priority. I have found dual-compatible wal-e/wal-g easy to run in practice.

I have thought over the configuration situation for a while, as I've long wanted to switch to configuration files, and possibly multiple as to address multiple previous timelines to close gaps across multiple promotions...but, I also think a minimal environment variable approach is admirably accessible and easy to get started with. Barring mechanical issues, I figured both might be viable, with the idea that simple uses and early uses would be on env vars, and later on, more involved users would use (apparently) yaml.

x4m:
In this PR I propose to keep exact correspondence between env vars and yaml (and json) config key-values. Maybe, one day, we sill make config files more powerful, but I do not see necessity for this right now.

PR #116, opened 2018-09-13
Add brotli support

CyberDem0n:
Brotli libraries are linked statically with wal-g.
In order to achieve that, I am building libraries before compiling wal-g in the [build.sh](https://github.com/wal-g/wal-g/pull/116/files#diff-0b83f9dedf40d7356e5ca147a077acb4R4) script.
Besides that, build.sh changes the content of vendor/github.com/google/brotli/go/cbrotli/cgo.go file, in order to specify include and lib directories and change library names to `*-static`.

I agree this is a nasty trick, but unfortunately, we can't avoid changes in the `cgo.go`, because linking again static libraries requires specifying all dependencies together:
* `libbrotlidec-static.a` depends on `libbrotlicommon-static.a`
* `libbrotlienc-static.a` depends on `libbrotlicommon-static.a` and `libm`


Tinsane:
Cool!! Thanks for your contribution! The failed unit test has a known issue, which we found not so long ago, but still didn't fix in the master branch. You can fix it yourself by replacing -1s with 0s in this function: https://github.com/wal-g/wal-g/blob/master/walg_test/lz4_compression_test.go#L143-L154.
By the way, we have already implemented brotli in another WAL-G fork, but it depends on some .so.1 brotli libraries, so now we'd like to make some kind of WAL-G package with brotli to do an easy distribution.

x4m:
Cool! Looks good.

PR #131, opened 2018-11-15
Removed outdated garbage from source code

Tinsane:
Removed unused tools, because of testing infrastructure update.


x4m:
OK, seems like all these cases are now covered by unit tests and integration tests.

PR #136, opened 2018-11-16
Logging levels

Tinsane:
To make WAL-G debug easier added prefixes: ERROR, WARNING, INFO for every WAL-G output. Also, created loggers for printing settings possibility.


x4m:
Thanks, pushed. BTW it'd be good if logs were parsable by things for logs collecting: usually they expect timestamp and log level

PR #137, opened 2018-11-23
Made more convenient way to print log with fatal

Tinsane:
Made possible to distinguish way of error printing for developers and normal users.


x4m:
thanks, pushed

PR #145, opened 2018-11-27
Got rid of increment_base and optimized incrementing

Tinsane:


backup.go
```diff
@@ -92,17 +95,122 @@ func (backup *Backup) GetKeys() ([]string, error) {
...
 }
 
-func (backup *Backup) fetchSentinel() (sentinelDto S3TarBallSentinelDto, err error) {
+func (backup *Backup) fetchSentinel() (S3TarBallSentinelDto, error) {
+	sentinelDto := S3TarBallSentinelDto{}
 	backupReaderMaker := NewS3ReaderMaker(backup.Folder, backup.getStopSentinelPath())
 	backupReader, err := backupReaderMaker.Reader()
 	if err != nil {
-		return S3TarBallSentinelDto{}, err
+		return sentinelDto, err
```
x4m: m.b. just return nil?

Tinsane: It's a structure, so it's impossible to return nil here. Also, when I tried to do it in such a way I got several issues with value/reference type, in this function usages.

PR #146, opened 2018-11-27
Refactored S3 out of source code to storage interface

Tinsane:


testtools/mock_storage.go
```diff
@@ -5,31 +5,22 @@ import (
 	"sync"
 )
 
-type MockStorage struct {
+type InMemoryStorage struct {
```
x4m: Does in-memory store actually store anything?

Tinsane: Yes

PR #147, opened 2018-11-28
Unit tests for increment optimize

Tinsane:


storage_folder.go
```diff
@@ -22,11 +22,18 @@ func (err ObjectNotFoundError) Error() string {
 type StorageFolder interface {
 	// Path should always ends with '/'
 	GetPath() string
+
+	// Should return objects with relative paths
 	ListFolder() (objects []StorageObject, subFolders []StorageFolder, err error)
+
```
x4m: This interface is partially commented. I missed this last time I reviewed.

backup_push_handler.go
```diff
@@ -21,7 +21,7 @@ func getDeltaConfig() (maxDeltas int, fromFull bool) {
 	origin, hasOrigin := os.LookupEnv("WALG_DELTA_ORIGIN")
 	if hasOrigin {
 		switch origin {
-		case "LATEST":
+		case LatestString:
```
x4m: Is this the only LATEST literal?

Tinsane: No, but other occurrences are inside of user message strings.

PR #149, opened 2018-11-28
Refactored deleteBeforeTarget and fixed bug of garbage non deletion

Tinsane:


delete_handler.go
```diff
@@ -27,6 +27,22 @@ type DeleteCommandArguments struct {
 	dryrun     bool
 }
 
+// TODO : unit tests
+func modifyDeleteTarget(target *Backup, findFull bool) *Backup {
```
x4m: modify?
adjust?
check?

delete_handler.go
```diff
@@ -152,48 +168,44 @@ func ParseDeleteArguments(args []string, fallBackFunc func()) (result DeleteComm
...
+		return
 	}
 
-	skipLine, walSkipFileName := ComputeDeletionSkipline(backups, target)
+	for _, garbageName := range garbageToDelete {
+		dropBackup(backupFolder, garbageName)
+	}
+	skipLine, walSkipFileName := ComputeDeletionSkipline(backupToScan, target)
+	if skipLine < len(backupToScan)-1 {
+		deleteWALBefore(walSkipFileName, walFolder)
```
x4m: technically you can have only one DeleteWAL call, but yeah, it is safer this way

PR #151, opened 2018-11-28
Reorganization of the code base

Savichev-Igor:
I tried to follow this [pattern](https://github.com/golang-standards/project-layout/blob/master/README.md)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "submodules/brotli"]
```
Tinsane: How did you solve the problem with brotli internal linking?

Gopkg.toml
```diff
@@ -1,26 +1,86 @@
+# Gopkg.toml example
```
Tinsane: Is there any need to leave this example here as it is?

Savichev-Igor: It's autogenerated file from [dep](https://github.com/golang/dep)

With his help we can remove `vendors/` from repo =)

.gitmodules
Savichev-Igor: I'm working on it, but now i just move `.brotli` to `git submodule`

In my opinion is a more correct solution and expandable

PR #153, opened 2018-12-04
Fix subfolders for backwards compatibility

x4m:
Recent refactoring changed subfolder hierarchy, this commit brings back old folder trees.

### DIFF
--- a/internal/backup.go
+++ b/internal/backup.go
@@ -33,8 +33,8 @@ type Backup struct {
 	Name             string
 }
 
-func NewBackup(folder StorageFolder, name string) *Backup {
-	return &Backup{folder.GetSubFolder(BaseBackupPath), name}
+func NewBackup(baseBackupFolder StorageFolder, name string) *Backup {
+	return &Backup{baseBackupFolder, name}
 }
 
 func (backup *Backup) getStopSentinelPath() string {
--- a/internal/backup.go
+++ b/internal/backup.go
@@ -34,7 +34,7 @@ type Backup struct {
 }
 
 func NewBackup(folder StorageFolder, name string) *Backup {
-	return &Backup{folder, name}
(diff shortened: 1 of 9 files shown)


internal/s3_folder.go
```diff
@@ -65,10 +66,7 @@ func (folder *S3Folder) ReadObject(objectRelativePath string) (io.ReadCloser, er
 }
 
 func (folder *S3Folder) GetSubFolder(subFolderRelativePath string) StorageFolder {
-	if !strings.HasSuffix(subFolderRelativePath, "/") {
-		subFolderRelativePath = subFolderRelativePath + "/"
-	}
-	return NewS3Folder(folder.uploader, folder.S3API, *folder.Bucket, folder.Path+subFolderRelativePath)
+	return NewS3Folder(folder.uploader, folder.S3API, *folder.Bucket, path.Join(folder.Path, subFolderRelativePath) + "/")
```
Tinsane: Is usage of path.Join valid here? What happens here if we use some other platform, like Windows?

internal/backup.go
```diff
@@ -34,7 +34,7 @@ type Backup struct {
 }
 
 func NewBackup(folder StorageFolder, name string) *Backup {
-	return &Backup{folder, name}
+	return &Backup{folder.GetSubFolder(BaseBackupPath), name}
```
Tinsane: It doesn't work for some backup creation scenarios. E. g. in HandleDelete. Check for this change correctness in all usages of NewBackup function.

internal/s3_folder.go
x4m: I've replaced path.join with hand-written, possibly more buggy, but platfrom-independent join

internal/backup.go
x4m: Done.

PR #162, opened 2018-12-25
Fixed issue with existing files when retrying

Tinsane:
There were messages from users that WAL-G fails because of retries during backup-fetch. This commit fixes file existance error during subsequent fetches.


x4m:
Looks good to me. It will fix current retry issues.
But do you think it is a good solution to delete this unsucessful files? May be we should just recreate them during fetch? Just handle file existance in creation?

x4m:
Meh... this does not fix the problem. There might be prior files from same tar.

Issue #163, opened 2019-01-01
Allow using GCS with default service account creds (metadata server)


Dirbaio:
Currently, the GCS integration requires that `GOOGLE_APPLICATION_CREDENTIALS` is set, and errors otherwise.

https://github.com/wal-g/wal-g/blob/9d01f3a62a150724cc9ab8722326dd8dfe2768cf/internal/gcs_folder.go#L33-L41

However, when running in GCE or GKE, it's useful to let the application use the service account of the machine it's running on (via the metadata server). GCP client libraries [automatically do this](https://cloud.google.com/docs/authentication/production) if you initialize them with default settings.

wal-g shouldn't specify the `option.WithCredentialsFile` option above, and just let the GCP lib automatically get creds from `GOOGLE_APPLICATION_CREDENTIALS` or the metadata service.

Dirbaio:
I'll happily make a PR soon :)

(Wal-e also leaves credential management to the GCP SDK. Their docs wrongly state that `GOOGLE_APPLICATION_CREDENTIALS` is required, though.)

PR #164, opened 2019-01-02
Let the GCS client library handle authentication.

Dirbaio:
Follow up to #163.

This allows using other forms of application-default credentials
other than GOOGLE_APPLICATION_CREDENTIALS, such as getting them
from the metadata service.

Things: 
- My editor runs goimports on save so some imports are switched around. If this is a problem, I can resend the PR without the changes. Ideally all the code should be formatted according to goimports though :) 
- For the test to continue working in your local machine, you can do `gcloud auth application-default login`, and wal-g will pick up that auth.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -2,45 +2,45 @@ package internal
 
 import (
 	"encoding/json"
-	"github.com/go-yaml/yaml"
-	"github.com/wal-g/wal-g/internal/tracelog"
 	"io/ioutil"
 	"os"
 	"os/user"
 	"path/filepath"
 	"strings"
+
+	"github.com/go-yaml/yaml"
+	"github.com/wal-g/wal-g/internal/tracelog"
 )
 
 var (
 	WalgConfig        *map[string]string
 	allowedConfigKeys = map[string]*string{
-		"WALG_S3_PREFIX":                 nil,
-		"WALE_S3_PREFIX":                 nil,
-		"WALG_FILE_PREFIX":               nil,
-		"WALE_FILE_PREFIX":               nil,
-		"WALG_GS_PREFIX":                 nil,
-		"WALE_GS_PREFIX":                 nil,
-		"GOOGLE_APPLICATION_CREDENTIALS": nil,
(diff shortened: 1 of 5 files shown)


internal/config.go
```diff
@@ -2,45 +2,45 @@ package internal
...
 var (
 	WalgConfig        *map[string]string
 	allowedConfigKeys = map[string]*string{
-		"WALG_S3_PREFIX":                 nil,
-		"WALE_S3_PREFIX":                 nil,
-		"WALG_FILE_PREFIX":               nil,
-		"WALE_FILE_PREFIX":               nil,
-		"WALG_GS_PREFIX":                 nil,
-		"WALE_GS_PREFIX":                 nil,
-		"GOOGLE_APPLICATION_CREDENTIALS": nil,
```
x4m: Currently, the user can set up this var using not just env var, but also with yaml or json config of WAL-G. If we rely only on GCP SDK this will not work. How do you think, is this functionality useful, or we can simply drop it? Let's also ask @carolfly86 and @NikolayS on their opinion, since they were asking about GCP.

Dirbaio: Envvar + metadata server is the "standard" way to give credentials to an app using a GCP client library. The S3 integration does the same, and wal-e's GS and S3 too.

I agree with you that it's somewhat inconsistent though... We can also get `GOOGLE_APPLICATION_CREDENTIALS` via wal-g's config system, pass it to the GCS client lib if set, and if not let the lib do the default (which will check `GOOGLE_APPLICATION_CREDENTIALS` again, and fall back to metadata server). 

But then we should do the same for `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY,` `AWS_SECURITY_TOKEN`, `AWS_PROFILE` ... for consistency.

Dirbaio: And we probably don't want to do that because it gets way more complex than google's :)

x4m:
Thanks, pushed!
We will have another prerelease soon after fixing again one small bug in retries.

PR #169, opened 2019-01-22
Fix pg_control check for WAL-E backups

x4m:
Currently, WAL-E backup restoration was broken since pre-release 0.2.1+
We were checking for pg_control before understanding that it is not necessary. This commit fixes it.

### DIFF
--- a/internal/backup.go
+++ b/internal/backup.go
@@ -120,14 +120,20 @@ func (backup *Backup) unwrap(dbDataDirectory string, sentinelDto BackupSentinelD
 	if err != nil {
 		return err
 	}
+
+	// Check name for backwards compatibility. Will check for `pg_control` if WALG version of backup.
+	need_pg_control := isPgControlRequired(backup, sentinelDto)
+
+	if pgControlKey == "" && need_pg_control {
+		return NewPgControlNotFoundError()
+	}
+
 	err = ExtractAll(tarInterpreter, tarsToExtract)
 	if err != nil {
 		return err
 	}
-	// Check name for backwards compatibility. Will check for `pg_control` if WALG version of backup.
-	re := regexp.MustCompile(`^([^_]+._{1}[^_]+._{1})`)
-	match := re.FindString(backup.Name)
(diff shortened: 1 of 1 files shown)


internal/backup.go
```diff
@@ -138,6 +144,13 @@ func (backup *Backup) unwrap(dbDataDirectory string, sentinelDto BackupSentinelD
 	return nil
 }
 
+func isPgControlRequired(backup *Backup, sentinelDto BackupSentinelDto) bool {
+	re := regexp.MustCompile(`^([^_]+._{1}[^_]+._{1})`)
+	balg_basebackup_name := re.FindString(backup.Name) == ""
```
Tinsane: a typo

internal/backup.go
```diff
@@ -120,14 +120,20 @@ func (backup *Backup) unwrap(dbDataDirectory string, sentinelDto BackupSentinelD
 	if err != nil {
 		return err
 	}
+
+	// Check name for backwards compatibility. Will check for `pg_control` if WALG version of backup.
+	need_pg_control := isPgControlRequired(backup, sentinelDto)
+
+	if pgControlKey == "" && need_pg_control {
```
Tinsane: Snake case among camel case??

test/backup_test.go
```diff
@@ -29,3 +29,14 @@ func TestGetTarNames(t *testing.T) {
 	assert.NoError(t, err)
 	assert.ElementsMatch(t, []string{"1", "2", "3"}, tarNames)
 }
+
+func TestIsPgControlRequired(t *testing.T) {
```
Tinsane: Lets split it into test cases.

internal/backup.go
```diff
@@ -62,8 +62,7 @@ func (backup *Backup) GetTarNames() ([]string, error) {
 	return result, nil
 }
 
-// TODO : unit tests
```
Tinsane: There are no tests for this function so it's better to return TODO back.

x4m: It is called from TestIsPgControlRequired, all successful code paths are executed

PR #173, opened 2019-01-26
Reading GPG keys from files using environment variables

Savichev-Igor:
Variant of solution for #92


x4m:
@fdr, @secwall, @davidfetter can I ask your opinions too?
In this PR we want to remove the dependency on external gpg.
Currently, gpg is used to read keys from gpg storage, real encryption is done by compiled-in OpenPGP. This PR allows to read keys from a file or directly from config (thus, reducing startup time and easing deployment). Also, this PR allows specifying the passphrase for PGP key. All this functionality seems good to me.

Also, this PR emits WARNING in case of usage of old-style keys invoked from gpg storage.
This implies that somewhere in future we will have a major release stating "WALG_GPG_KEY_ID is not supported, please use WALG_PGP_KEY or WALG_PGP_KEY_PATH". Does this seem good decision in terms of future WAL-G development?

fdr:
Does WAL-G have so many GPG users that it's worth adding too much expense to perform extraordinary compatibility maintenane? I think not: I'd provide the service of a message only. I'm sure if WARNING is more wise than an ERROR: warnings tend to get ignored.

PR #174, opened 2019-01-28
bugfix - missing error handling

g0djan:
when an error occurred in the listFolder, "No backups found" was returned instead of the error itself


x4m:
Good catch, thanks!

PR #179, opened 2019-02-05
Fix WALG_GPG_KEY_ID env variable usage

bn0ir:
WAL-G will use both WALG_GPG_KEY_ID and WALE_GPG_KEY_ID env variables for gpg key id now.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,6 +1,7 @@
 Adam Shannon <adamkshannon@gmail.com>
 Andrey Borodin <amborodin@acm.org>
 Barthelemy Dagenais <barthelemy@infobart.com>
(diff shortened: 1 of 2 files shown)


x4m:
That looks like a serious bug! Thanks!

I've adjusted PR slightly, can I ask you to verify that everything ok?
For some reason we support contributors list in the source code, plz check the spelling of your name there.

bn0ir:
Fix PR to support both WALE_GPG_KEY_ID and WALG_GPG_KEY_ID env vars.
With `getSettingValue("WALG_GPG_KEY_ID")` it miss WALE_GPG_KEY_ID value: getSettingValue perform only WALE -> WALG forward check.

My name spelling is correct, thanks!

PR #188, opened 2019-02-24
yet an other makefile

alijkl:
- pakages dependencies defined in the makefile
 - work only with one make job at the moment
 - allow to build and link (static) with lib brotli
 - permit to set the same flags for wal-g and the test
   shared, static-libgo, static

Note:
 On my system I recive a signal SIGSEGV: segmentation violation code=2
 when running the test using -static
 But I don't know if it is related to my system or wal-g code yet.

### DIFF
--- a/Makefile.gccgo
+++ b/Makefile.gccgo
@@ -1,0 +1,115 @@
+# -*- mode: makefile -*-. –
+
+GOPATH=$(HOME)/go
+goarch := gccgo_linux_amd64
+
+pkgwal-g := github.com/wal-g/wal-g
+
+pkgdep += "cloud.google.com/go/storage"
+pkgdep += "github.com/DataDog/zstd"
+pkgdep += "github.com/RoaringBitmap/roaring"
+pkgdep += "github.com/aws/aws-sdk-go/aws"
+#pkgdep += "github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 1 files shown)


Makefile.gccgo
```diff
@@ -0,0 +1,115 @@
...
+pkgdep += "github.com/DataDog/zstd"
+pkgdep += "github.com/RoaringBitmap/roaring"
+pkgdep += "github.com/aws/aws-sdk-go/aws"
+#pkgdep += "github.com/aws/aws-sdk-go/aws/awserr"
+#pkgdep += "github.com/aws/aws-sdk-go/aws/defaults"
+#pkgdep += "github.com/aws/aws-sdk-go/aws/session"
+#pkgdep += "github.com/aws/aws-sdk-go/service/s3"
+#pkgdep += "github.com/aws/aws-sdk-go/service/s3/s3iface"
+#pkgdep += "github.com/aws/aws-sdk-go/service/s3/s3manager"
+#pkgdep += "github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface"
```
g0djan: A lot of commented lines. Is it necessary here?

Makefile.gccgo
```diff
@@ -0,0 +1,115 @@
+# -*- mode: makefile -*-. –
+
```
g0djan: Should it be used with gccgo?
Could you give examples of using this Makefile?

PR #194, opened 2019-03-01
add extensions logic

g0djan:


g0djan:
@Tinsane I marked comments from https://github.com/wal-g/wal-g/pull/190, which I fixed

internal/extensions.go
```diff
@@ -0,0 +1,63 @@
...
+	"plugin"
+)
+
+var Extensions []Extension
+
+type Extension interface {
+	TryPrintHelp(command, firstArgument string) bool
+	HasCommand(command string) bool
+	Execute(command string, uploader *Uploader, folder storage.Folder, args []string)
+	HasAllowedConfigKey(key string) bool
```
Tinsane: Maybe it's better to replace this function with function returning all allowed config keys for this extension.

internal/extensions.go
```diff
@@ -0,0 +1,63 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"plugin"
+)
+
+var Extensions []Extension
+
+type Extension interface {
+	TryPrintHelp(command, firstArgument string) bool
```
Tinsane: Let's improve TryPrintHelp interface, passing here all arguments instead of only first one.

internal/extensions.go
```diff
@@ -0,0 +1,63 @@
...
+	HasCommand(command string) bool
+	Execute(command string, uploader *Uploader, folder storage.Folder, args []string)
+	HasAllowedConfigKey(key string) bool
+	Flush(time BackupTime, folder storage.Folder)
+}
+
+func LoadExtensions() {
+	if _, err := os.Stat("./extensions"); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
g0djan: Now user can set a directory for reading ".so" files in program arguments. But by default, it is "./extensions/build".

internal/extensions.go
```diff
@@ -0,0 +1,63 @@
...
+		return
+	}
+	files, err := ioutil.ReadDir("./extensions")
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
g0djan: Done. But I'm not sure that handling error in "init" of the main package is the best way.

internal/extensions.go
g0djan: Fixed

internal/extensions.go
g0djan: Fixed

PR #199, opened 2019-03-05
Moved backup finishing function from TarBall

Tinsane:
This function definitely shouldn't be a part of TarBall interface.

### DIFF
--- a/internal/backup_push_handler.go
+++ b/internal/backup_push_handler.go
@@ -3,12 +3,26 @@ package internal
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
+	"github.com/pkg/errors"
 	"github.com/wal-g/wal-g/internal/tracelog"
 	"os"
 	"path/filepath"
 	"strconv"
 	"time"
 )
+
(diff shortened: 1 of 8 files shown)


internal/backup_push_handler.go
```diff
@@ -134,31 +136,64 @@ func HandleBackupPush(archiveDirectory string, uploader *Uploader) {
...
-			currentBackupSentinelDto.IncrementFrom = &previousBackupName
-			if previousBackupSentinelDto.isIncremental() {
-				currentBackupSentinelDto.IncrementFullName = previousBackupSentinelDto.IncrementFullName
-			} else {
-				currentBackupSentinelDto.IncrementFullName = &previousBackupName
-			}
-			currentBackupSentinelDto.IncrementCount = &incrementCount
-		}
+	if timelineChanged {
+		uploader.finish() // TODO : is it needed here?
```
x4m: yes we should try to make consistent backup. The only problem with this timelineChanged thin is backup is named improperly. It still can be restored, but only on new timeline.
I think we should also upload that json sentinel, just make it unreadable....

PR #205, opened 2019-03-09
Introduced cobra CLI

Tinsane:
Improved CLI code quality.


cmd/root.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var cfgFile string
+
+var RootCmd = &cobra.Command{
+	Use:   "wal-g",
+	Short: "PostgreSQL backup tool",
+	Long: `A longer description that spans multiple lines and likely contains
```
x4m: We need at least some text here

Tinsane: This PR is devoted to refactoring. Let's make an issue about the appropriate description to solve it in another PR?

x4m: OK

PR #213, opened 2019-03-20
Fix/race condition private key

femaref:


femaref:
87cb612 fixes a deadlock. both code paths need to RUnlock.

Tinsane:
Your issue is clear and the code seems really nice to me. Thank you for your contribution!

PR #215, opened 2019-03-22
fix delete retention length warning

pohzipohzi:
This fixes the help message when deleting backups with `retain`

Prior
```
➜  wal-g git:(master) ✗ wal-g delete retain 1000
WARNING: 2019/03/23 06:30:21.803143 Have only 1000 backups.
```

After
```
➜  wal-g git:(master) ✗ wal-g delete retain 1000
WARNING: 2019/03/23 06:31:32.969621 Have only 3 backups.
```


x4m:
Thanks! The information was quite misleading.

PR #217, opened 2019-03-25
refactoring logic of delete before [FIND_FULL] backup_name

g0djan:
it's part of big refactoring and a lot of code will be deleted at the end of it

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -5,13 +5,15 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/internal/storages/storage"
 	"github.com/wal-g/wal-g/internal/tracelog"
+	"regexp"
 	"strconv"
 	"time"
 )
 
 const (
-	ConfirmFlag = "confirm"
+	ConfirmFlag            = "confirm"
 	DeleteShortDescription = "Clears old backups and WALs"
 
 	DeleteRetainExamples = `  retain 5                      keep 5 backups
@@ -25,6 +27,8 @@ const (
 var StringModifiers = []string{"FULL", "FIND_FULL"}
 
 var confirmed = false
+var regexpLSN = regexp.MustCompile("[0-9A-F]{24}")
+var maxCountOfLSN = 2
 
 func extractDeleteModifierFromArgs(args []string) (int, string) {
 	if len(args) == 1 {
@@ -114,7 +118,14 @@ func runDeleteBefore(cmd *cobra.Command, args []string) {
 	}
 	before, err := time.Parse(time.RFC3339, beforeStr)
 	if err != nil {
-		internal.HandleDeleteBeforeBackup(folder, beforeStr, modifier, !confirmed)
+		target, err := internal.FindTargetBeforeName(folder, beforeStr, modifier)
+		// after updating delete before time it will be made common
+		if err != nil {
+			tracelog.ErrorLogger.FatalError(err)
+		}
+		if confirmed {
+			err = internal.DeleteBeforeTarget(folder, target, earlierCreated)
+		}
 	} else {
 		internal.HandleDeleteBeforeTime(folder, before, modifier, !confirmed)
 	}
@@ -133,4 +144,13 @@ func init() {
 	deleteCmd.AddCommand(deleteRetainCmd, deleteBeforeCmd)
 
 	deleteCmd.PersistentFlags().BoolVar(&confirmed, ConfirmFlag, false, "Confirms backup deletion")
+}
+
+// it's here because it's part of postgres logic and in future it will be placed in postgres part of wal-g
+func earlierCreated(object1 storage.Object, object2 storage.Object) bool {
+	return fetchLSN(object1) < fetchLSN(object2)
+}
+
+func fetchLSN(object storage.Object) string {
+	return regexpLSN.FindAllString(object.GetName(), maxCountOfLSN)[0]
 }
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -124,10 +124,9 @@ func runDeleteBefore(cmd *cobra.Command, args []string) {
 			tracelog.ErrorLogger.FatalError(err)
 		}
 		if confirmed {
-			isYoungerEnough := func(object storage.Object) bool {
+			err = storage.DeleteObjectsWhere(folder, func(object storage.Object) bool {
 				return earlierCreated(object, target)
-			}
-			err = internal.DeleteOldObjects(folder, isYoungerEnough)
+			})
 		}
 	} else {
 		internal.HandleDeleteBeforeTime(folder, before, modifier, !confirmed)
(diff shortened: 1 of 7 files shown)


internal/storages/storage/folder.go
```diff
@@ -24,3 +25,32 @@ type Folder interface {
...
+	for i, object := range objects {
+		// meaning: now it's objects from root of folder tree.
+		// I think that []storage.Object is right returning type for ListFolderRecursively
+		objects[i] = NewLocalObject(GetRelativePath(object, folder), object.GetLastModified())
+	}
+	return objects
+}
+
+func GetRelativePath(object Object, folder Folder) string {
+	return path.Join(folder.GetPath(), object.GetName())
```
Tinsane: strings.TrimPrefix instead of path.Join?

internal/delete_handler.go
```diff
@@ -294,3 +294,40 @@ func getGarbageFromPrefix(folders []storage.Folder, nonGarbage []BackupTime) []s
 	}
 	return garbage
 }
+
+func DeleteBeforeTarget(
+	folder storage.Folder,
+	target storage.Object,
```
Tinsane: Let's take as a parameter one variable predicate - object in a store instead of a target and before function.

internal/storages/storage/folder.go
g0djan: That was GetFullPath, I removed it because that was redundant

internal/delete_handler.go
g0djan: Done

test/folder_test.go
```diff
@@ -0,0 +1,36 @@
...
+		"subfolder1/subfolder11/c",
+		"subfolder2/d",
+	}
+	for _, relativePath := range paths {
+		err := folder.PutObject(relativePath, &bytes.Buffer{})
+		assert.NoError(t, err)
+	}
+	fullPathObjects, err := storage.ListFolderRecursively(folder)
+	assert.NoError(t, err)
+	found := false
```
Tinsane: Extra declaration.

test/delete_test.go
```diff
@@ -88,3 +91,67 @@ func TestSkiplineComputationAfterUpgrade(t *testing.T) {
...
+
+func TestFindTargetBeforeName_ReturnsFullBackup_Without_Modifier(t *testing.T) {
+	targetDelta := "base_000000010000000000000007"
+	folder := createMockStorageFolderWithDeltaBackups(t)
+	object, err := internal.FindTargetBeforeName(folder, targetDelta, internal.NoDeleteModifier)
+	assert.NoError(t, err)
+	assert.Equal(t, targetDelta + internal.SentinelSuffix, object.GetName())
+}
+
+func TestFindTargetBeforeName_ReturnsErrorForDeltaBackup_With_FIND_FULL(t *testing.T) {
```
Tinsane: Test name doesn't match with its code.

test/delete_test.go
```diff
@@ -88,3 +91,67 @@ func TestSkiplineComputationAfterUpgrade(t *testing.T) {
...
+	savedObjects, err := storage.ListFolderRecursively(folder)
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(savedObjects))
+	assert.Equal(t, expectedOnlyOneSavedObjectName, savedObjects[0].GetName())
+}
+
+func TestFindTargetBeforeName_ReturnsErrorForDeltaBackup_Without_Modifier(t *testing.T) {
+	targetDelta := "base_000000010000000000000009_D_000000010000000000000007"
+	folder := createMockStorageFolderWithDeltaBackups(t)
+	expectedErrorMessage := " is incremental and it's predecessors cannot be deleted. Consider FIND_FULL option."
```
Tinsane: You didn't check the error message in this code. Anyway, it's better to check the error type or to check only error existence.

internal/delete_handler.go
```diff
@@ -91,7 +88,7 @@ func HandleDeleteBeforeTime(folder storage.Folder, before time.Time, modifier in
 
 // TODO : unit tests
 func deleteBeforeTarget(folder storage.Folder, target *Backup, findFull, dryRun bool) {
-	target = adjustDeleteTarget(target, findFull)
+	target, _ = adjustDeleteTarget(target, findFull)
```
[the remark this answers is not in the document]
g0djan: ok, but in the end, this function will be deleted

cmd/delete.go
```diff
@@ -124,7 +124,10 @@ func runDeleteBefore(cmd *cobra.Command, args []string) {
 			tracelog.ErrorLogger.FatalError(err)
 		}
 		if confirmed {
-			err = internal.DeleteBeforeTarget(folder, target, earlierCreated)
+			isYoungerEnough := func(object storage.Object) bool {
```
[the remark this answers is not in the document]
g0djan: done

internal/delete_handler.go
```diff
@@ -295,20 +292,20 @@ func getGarbageFromPrefix(folders []storage.Folder, nonGarbage []BackupTime) []s
 	return garbage
 }
 
-func DeleteBeforeTarget(
-	folder storage.Folder,
-	target storage.Object,
-	before func(object1 storage.Object, object2 storage.Object) bool) error {
-	relativePathObjects, err := storage.ListFolderRecursively(folder)
+func DeleteOldObjects(folder storage.Folder, isElderEnough func(object1 storage.Object) bool) error {
```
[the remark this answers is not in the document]
g0djan: done

test/folder_test.go
g0djan: done

test/delete_test.go
g0djan: done

internal/delete_handler.go
```diff
@@ -15,19 +16,20 @@ const (
...
-		tracelog.ErrorLogger.FatalError(err)
+		return nil, err
 	}
 	if sentinelDto.isIncremental() {
 		if findFull {
 			target.Name = *sentinelDto.IncrementFullName
 		} else {
-			tracelog.ErrorLogger.Fatalf("%v is incremental and it's predecessors cannot be deleted. Consider FIND_FULL option.", target.Name)
+			errorMessage := "%v is incremental and it's predecessors cannot be deleted. Consider FIND_FULL option."
+			return nil, NewWrongTypeError(fmt.Sprintf(errorMessage, target.Name))
```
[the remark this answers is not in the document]
g0djan: done. ForbiddenActionError

test/delete_test.go
g0djan: Why? If you don't check the message, you can't be sure which case you test.

internal/utility.go
```diff
@@ -182,3 +183,11 @@ func stripWalFileName(path string) string {
 	}
 	return ""
 }
+
+type ForbiddenActionError interface {
```
Tinsane: Add here tracelog printing (function Error), like in other errors.

test/delete_test.go
Tinsane: Then, let's do a separate type for it and check the error type. It is better because of the flexibility of this variant (If you change error message than you shouldn't change test. Also, the test shouldn't test the exact  error message, it should test logic behind the error).

internal/utility.go
g0djan: done

test/delete_test.go
g0djan: done

PR #219, opened 2019-03-26
Moved InMemoryStorage to storages

Tinsane:
Refactoring.


internal/storages/memory/folder.go
```diff
@@ -0,0 +1,83 @@
...
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+type Folder struct {
+	path    string
+	Storage *Storage
+}
+
```
x4m: Let's add here some notes about intended usage. Maybe warn users about possible /dev/null effect of process termination.

PR #225, opened 2019-03-28
Refactoring logic of delete before [FIND_FULL] time

g0djan:


internal/storages/storage/object.go
```diff
@@ -1,8 +1,30 @@
...
+import (
+	"time"
+)
 
 type Object interface {
 	GetName() string
 	GetLastModified() time.Time
 }
+
+type SortableObjectsSlice []Object
```
Tinsane: https://golang.org/pkg/sort/#Slice

cmd/delete.go
```diff
@@ -117,19 +120,22 @@ func runDeleteBefore(cmd *cobra.Command, args []string) {
...
-		if confirmed {
-			err = storage.DeleteObjectsWhere(folder, func(object storage.Object) bool {
-				return earlierCreated(object, target)
-			})
-		}
-	} else {
-		internal.HandleDeleteBeforeTime(folder, before, modifier, !confirmed)
+		beforeStr = fetchBackupName(potentialTarget)
+	}
+	target, err := internal.FindTargetBeforeName(folder, beforeStr, modifier)
```
Tinsane: What about replacing this function with a function FindTargetLSNBeforeName? It seems to me that such a substitution will simplify the code.

g0djan: I thought about it, but it is postgres specified logic.

internal/storages/storage/object.go
g0djan: Done.

x4m:
Thanks, pushed!

PR #231, opened 2019-04-02
refactoring delete handler

g0djan:


g0djan:
3 first commits should be squashed and 4th is apart

PR #232, opened 2019-04-04
fix lsn parser for compatibility with wal-e

g0djan:


internal/utility.go
```diff
@@ -174,13 +175,14 @@ func stripPrefixName(path string) string {
...
+var regexpLSN = regexp.MustCompile(patternLSN)
+
 // Strips the backup WAL file name.
 func stripWalFileName(path string) string {
-	name := stripBackupName(path)
-	name = strings.SplitN(name, "_D_", 2)[0]
-
-	if strings.HasPrefix(name, backupNamePrefix) {
-		return name[len(backupNamePrefix):]
+	founded_lsn := regexpLSN.FindAllString(path, 2)
```
x4m: just found_lsn, not founded_lsn

internal/utility.go
```diff
@@ -174,13 +175,14 @@ func stripPrefixName(path string) string {
...
+	if len(founded_lsn) > 0 {
+		return founded_lsn[0]
 	}
 	return ""
```
x4m: Can we return something line +inf, not -inf?
So that we will not delete unknown files...

x4m:
Thanks, pushed

PR #240, opened 2019-04-09
Tests for tar_interpreter.go -> TarInterpreter -> Interpret

dimastark:


test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"bytes"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/wal-g/wal-g/internal"
+)
+
+type testInterpretOptions struct {
```
g0djan: What is the reason for creating this struct? it doesn't simplify the code for reading and hide dependencies.
Use KISS principle here.

test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+func testInterpret(t *testing.T, options testInterpretOptions) (os.FileInfo, os.FileInfo) {
+	tarInterpreter := &internal.FileTarInterpreter{
+		DBDataDirectory: options.DBDataDirectory,
+	}
+
+	err := options.Create(options.Name)
+
+	assert.NoError(t, err)
+
+	defer options.Delete(options.Name)
```
g0djan: It looks like an unhandled error. Try to replace it with something similar to this https://pastebin.com/7xAFJtpX

test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+		&bytes.Buffer{},
+		&tar.Header{
+			Name:     options.Name,
+			Typeflag: options.Typeflag,
+		},
+	)
+
+	assert.NoError(t, err)
+
+	defer options.Delete(path.Join(options.DBDataDirectory, options.Name))
```
g0djan: It looks like an unhandled error

test/tar_interpreter_test.go
g0djan: Passing keyword arguments is redundant here.

test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+		DBDataDirectory: dbDataDirectory,
+
+		Name:     "test_dir",
+		Typeflag: tar.TypeDir,
+
+		Create: mkdirAll,
+		Delete: os.RemoveAll,
+	})
+
+	defer os.RemoveAll(dbDataDirectory)
```
g0djan: Unhandled error

test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+	DBDataDirectory string
+
+	Name     string
+	Typeflag byte
+
+	Create func(string) error
+	Delete func(string) error
+}
+
+func testInterpret(t *testing.T, options testInterpretOptions) (os.FileInfo, os.FileInfo) {
```
g0djan: You don't cover cases when an error occurs.

g0djan: You can use assert.IsType for it.

test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+	assert.NoError(t, err)
+
+	return srcFileInfo, dstFileInfo
+}
+
+func mkdirAll(path string) error {
+	return os.MkdirAll(path, 0766)
+}
+
+func justCreate(path string) error {
```
g0djan: Rename it to createFile and comment that it's needed for unification returning type with createDir.

test/tar_interpreter_test.go
```diff
@@ -0,0 +1,156 @@
...
+	assert.NoError(t, err)
+
+	dstFileInfo, err := os.Lstat(path.Join(options.DBDataDirectory, options.Name))
+
+	assert.NoError(t, err)
+
+	return srcFileInfo, dstFileInfo
+}
+
+func mkdirAll(path string) error {
```
g0djan: Rename to createDir for a unique name with createFile

test/tar_interpreter_test.go
dimastark: I think a comment is not needed, the purpose of the function becomes obvious if you look at the place of use.

dimastark: I think a comment is not needed, the purpose of the function becomes obvious if you look at the place of use.

test/tar_interpreter_test.go
dimastark: I write this function only for **happy path** cases. I'll deal with **bad path** later.

test/tar_interpreter_test.go
```diff
@@ -11,48 +11,38 @@ import (
...
-
-	Name     string
-	Typeflag byte
-
-	Create func(string) error
-	Delete func(string) error
-}
-
-func testInterpret(t *testing.T, options testInterpretOptions) (os.FileInfo, os.FileInfo) {
+func testInterpret(t *testing.T, dbDataDirectory, name string, typeflag byte, create, delete func(string) error) (os.FileInfo, os.FileInfo) {
```
g0djan: Break this line into several

g0djan: Nice work. Fix it and I'll merge it.

test/tar_interpreter_test.go
g0djan: Ok, thanks. We will wait for your next PR :)

PR #241, opened 2019-04-09
Added GetMaxConcurrency unit tests

BurningMarshmallow:
* Exporting GetMaxConcurrency so it could be tested outside package

* Added GetMaxConcurrency unit tests


internal/utility.go
```diff
@@ -108,7 +108,6 @@ func getMaxUploadDiskConcurrency() int {
 	return GetMaxConcurrency("WALG_UPLOAD_DISK_CONCURRENCY", 1)
 }
 
-// TODO : unit tests
```
g0djan: There left one case you don't cover. os.SetEnv("WALG_UPLOAD_CONCURRENCY", "incorrect number").
But interface of this function is sick because we shouldn't to Fatal there with an error.
Let's don't do such things as fatal or os.Exit here and just return an error to the main.

g0djan:
I'll see it tomorrow

BurningMarshmallow:
Tried to add returning error, however it looks ugly at some places (tried to use panic, if it was already there). I will be glad to discuss any changes in handling of these situations (if they're not too big for this small task, of course)

internal/wal_push_handler.go
```diff
@@ -26,10 +26,14 @@ func (err CantOverwriteWalFileError) Error() string {
 // HandleWALPush is invoked to perform wal-g wal-push
 func HandleWALPush(uploader *Uploader, walFilePath string) {
 	uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(WalPath)
-	bgUploader := NewBgUploader(walFilePath, int32(getMaxUploadConcurrency(16)-1), uploader)
+	concurrency, err := getMaxUploadConcurrency(16)
+	if err != nil {
+		panic(err)
```
g0djan: You don't need to panic here. Error is in a handler already. So there you can use tracelog.ErrorLogger.Fatalf

cmd/pg/wal_prefetch.go
```diff
@@ -20,7 +20,10 @@ var walPrefetchCmd = &cobra.Command{
 		if err != nil {
 			tracelog.ErrorLogger.FatalError(err)
 		}
-		internal.HandleWALPrefetch(uploader, args[0], args[1])
+		err = internal.HandleWALPrefetch(uploader, args[0], args[1])
```
g0djan: You should handle error in handler not in cobra command.

internal/prefetch.go
```diff
@@ -221,14 +226,18 @@ func GetPrefetchLocations(location string, walFileName string) (prefetchLocation
 
 // TODO : unit tests
 func forkPrefetch(walFileName string, location string) {
+	concurrency, err := getMaxDownloadConcurrency(16)
+	if err != nil {
+		tracelog.ErrorLogger.Println("WAL-prefetch failed: ", err)
```
g0djan: Okay

internal/wal_push_handler.go
BurningMarshmallow: I just copied it from below :). Your words seem correct, so I also changed it there, after UploadWALFile.

cmd/pg/wal_prefetch.go
BurningMarshmallow: Fine remark, fixed it

PR #244, opened 2019-04-15
Upload metadata file with backup

x4m:
We store json sentinel, which is functional now and is under quite
hard requirements. Usually it is big, maybe few MB or so. In
metadata file we want to have easily retrivable, frequently used
information.

### DIFF
--- a/internal/backup_push_handler.go
+++ b/internal/backup_push_handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/pkg/errors"
+	"github.com/wal-g/wal-g/internal/storages/storage"
 	"github.com/wal-g/wal-g/internal/tracelog"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


internal/backup_push_handler.go
```diff
@@ -180,18 +186,34 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string) {
...
 	if err != nil {
 		tracelog.ErrorLogger.Printf("Failed to upload sentinel file for backup: %s", backupName)
 		tracelog.ErrorLogger.FatalError(err)
 	}
 }
 
 // TODO : unit tests
-func UploadSentinel(uploader *Uploader, sentinelDto *BackupSentinelDto, backupName string) error {
+func UploadSentinel(uploader *Uploader, sentinelDto *BackupSentinelDto, backupName string, meta ExtendedMetadataDto) error {
+	meta.FinishTime = time.Now()
```
Tinsane: Let's move metadata creation and uploading to a separate function: "UploadMetadata". This will save us from using incomplete metadata object and separate its logic from logic related to Sentinel.

PR #245, opened 2019-04-15
makefile and pg targets

g0djan:


x4m:
Hi! The main concern here is about leaving the existing API intact. Does this expectation hold?

g0djan:
@x4m API is not changed, it's only about the building process.

Issue #250, opened 2019-04-24
Add support for archiving to minio over SSL using custom CA 


iamrz1:
Currently, wal-g doesn't support archiving to Minio using HTTPs endpoints. It will be nice to be able to use Minio servers secured with self-signed CA.

Regards,
Rezoan

x4m:
This is merged, right?

PR #251, opened 2019-04-24
Support for archiving to minio over ssl using custom CA

iamrz1:
xref:https://github.com/wal-g/wal-g/issues/250


iamrz1:
@x4m , @g0djan PTAL

internal/storages/s3/session.go
```diff
@@ -6,10 +6,15 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/pkg/errors"
+	"os"
 	"strconv"
 	"strings"
 )
 
+const (
+	S3_CA_CERT_FILE = "WALG_S3_CA_CERT_FILE"
```
g0djan: Add this key to internal/config.go in allowedConfigKeys

internal/storages/s3/session.go
```diff
@@ -79,5 +84,15 @@ func createSession(bucket string, settings map[string]string) (*session.Session,
 	}
 	config = config.WithRegion(region)
 
+	caFilePath := os.Getenv(S3_CA_CERT_FILE)
+	if caFilePath != "" {
+		if file, err := os.Open(caFilePath); err == nil {
+			defer file.Close()
```
g0djan: use loggedClose from utility

internal/storages/s3/session.go
iamrz1: Do we still use os.Getenv(S3_CA_CERT_FILE) ?

g0djan: Try to use internal.GetSettingValue

iamrz1: I initially tried to use `internal.GetSettingValue(S3_CA_CERT_FILE)` which required me to import `github.com/wal-g/wal-g/internal`, that in turn gave me an import cycle error during build:
```
import cycle not allowed
package github.com/wal-g/wal-g/main/pg
        imports github.com/wal-g/wal-g/cmd/pg
        imports github.com/wal-g/wal-g/internal
        imports github.com/wal-g/wal-g/internal/storages/s3
        imports github.com/wal-g/wal-g/internal
Makefile:16: recipe for target 'pg_build' failed
```
So I kept getEnv() for the time being.

internal/storages/s3/session.go
```diff
@@ -96,3 +98,10 @@ func createSession(bucket string, settings map[string]string) (*session.Session,
 
 	return session.NewSession(config)
 }
+
+func loggedClose(c io.Closer) {
```
g0djan: If you want to pass a message to LoggedCLose then change the signature of the function, don't copy it. https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

internal/storages/s3/session.go
iamrz1: I initially tried to use `internal.GetSettingValue(S3_CA_CERT_FILE)` which required me to import `github.com/wal-g/wal-g/internal`, that in turn gave me an import cycle error during build:
```
import cycle not allowed
package github.com/wal-g/wal-g/main/pg
        imports github.com/wal-g/wal-g/cmd/pg
        imports github.com/wal-g/wal-g/internal
        imports github.com/wal-g/wal-g/internal/storages/s3
        imports github.com/wal-g/wal-g/internal
Makefile:16: recipe for target 'pg_build' failed
```
Basically internal/storage_adapter.go imports internal/storages/s3, and now I have to import internal in internal/storages/s3, which become a circular thing.
So I kept getEnv() for the time being.

iamrz1: I found a better way: getting values from settingNames.

iamrz1:
Hello @g0djan,
 I have updated the pull. Can you please have a look again?

g0djan:
Yep. Now you can't import loggedClose from internal because it has cyclic error. Try to move utility to new package "utility" from "internal" and use loggedClose from this package instead of copy-paste.

iamrz1:
> Yep. Now you can't import loggedClose from internal because it has cyclic error. Try to move utility to new package "utility" from "internal" and use loggedClose from this package instead of copy-paste.

Okay, I am moving the function to another folder (wal-g/utility) outside wal-g/internal.

g0djan:
@iamrz1 I wait for the fix. I guess tests crashed because you update not all of the imports of the utility package

iamrz1:
@g0djan, fixed. It was due to how test script copies a few select directories.

PR #257, opened 2019-05-03
Compression refactoring

Tinsane:
Refactored compression logic to a separate folder.
Removed ReaderFromWriteCloser to adhere DRY.
Removed CompressingPipeWriter antipattern.


x4m:
Hi! While refactoring looks good and useful, CI is unhappy.

internal/compress_and_encrypt.go
```diff
@@ -0,0 +1,72 @@
...
+	return CompressAndEncryptError{errors.New(reason)}
+}
+
+func (err CompressAndEncryptError) Error() string {
+	return fmt.Sprintf(tracelog.GetErrorFormatter(), err.error)
+}
+
+// CompressAndEncrypt compresses input to a pipe reader. Output must be used or
+// pipe will block.
+func CompressAndEncrypt(source io.Reader, compressor compression.Compressor, crypter Crypter) io.Reader {
```
x4m: Eventually we will have to stop passing crypter here and there...

testtools/util.go
```diff
@@ -75,6 +66,14 @@ func AssertReaderIsEmpty(t *testing.T, reader io.Reader) {
 	assert.Equal(t, io.EOF, err)
 }
 
+type NopCloserWriter struct {
```
x4m: Don't we have this in ioutils?

Tinsane: Nop

PR #261, opened 2019-05-06
Refactoring

Tinsane:
Refactored redundant BackupTimes sorting method.
Replaced DelayedWriteCloser with buffering.
Refactored crypter creation logic.
Refactored limiter's logic.


internal/uploader.go
```diff
@@ -59,7 +62,7 @@ func (uploader *Uploader) Clone() *Uploader {
 		uploader.Compressor,
 		&sync.WaitGroup{},
 		uploader.deltaFileManager,
-		uploader.Success,
```
g0djan: What is the reason for this change?

PR #271, opened 2019-05-13
Added race detection to CI tests and fixed Padding Byte race condition

Tinsane:


docker/pg_tests/scripts/tests/delete_end_to_end_test.sh
```diff
@@ -50,7 +52,8 @@ done
 diff /tmp/dump4 /tmp/dump${FIRST}
 diff /tmp/dump9 /tmp/dumpLATEST
 
-pkill -9 postgres
+pkill -9 postgres || true
+pkill -9 wal-g || true
```
g0djan: Should you kill wal-g in the end in all tests?

Tinsane: I should kill it before removing pgdata.

g0djan: It means that you should kill it in all tests where you do "rm -rf ${PGDATA}".

PR #272, opened 2019-05-13
add backup label to info log of backup-push

HeikoOnnebrink:
In order to know the label of a wal-g backup-push operation the name of the written backup set should be written to stdout. This is required in automated scenarios where wal-g is called from external backup solutions / scheduler / etc.

From the stdout of wal-g backup-psuh we can grep the label name by cutting out the string after "Wrote backup set with name" 

This can serve as a fast "workaround" solution until a proper used defined backup label feature is available within wal-g (and later on can stay inside the code as a useful info field)


x4m:
Thanks, pushed!

PR #275, opened 2019-05-13
moved preventWalOverwrite from Uploader to BgUploader

AlSoEdit:


internal/wal_push_handler.go
```diff
@@ -23,18 +24,40 @@ func (err CantOverwriteWalFileError) Error() string {
 	return fmt.Sprintf(tracelog.GetErrorFormatter(), err.error)
 }
 
+func  GetWalOverwriteSetting() (preventWalOverwrite bool, err error) {
```
g0djan: Move it to configure.go and rename similar to other ConfigureSomething in this file.

AlSoEdit: Done

Issue #274, opened 2019-05-13
Properties not working from config file


Tri0L:
```
$ whoami
postgres

$ ls -la
-rw-rw-r--  1 postgres postgres  329 May 13 15:14 .walg.json

$ cat ./.walg.json 
{
	"WALG_LOG_LEVEL": "DEVEL",
	"USER": "postgres",
	"PGPORT": "5432",
	"PGUSER": "postgres",
	"OS_PASSWORD": "password",
	"PGHOST": "/var/run/postgresql",
	"OS_AUTH_URL": "authurl",
	"OS_USERNAME": "user",
	"OS_TENANT_NAME": "user",
	"WALG_SWIFT_PREFIX": "swift://wal-g/somefolder",
	"WALG_DELTA_MAX_STEPS": "5"
}
```

Time to time different errors:
```
$ wal-g backup-list
ERROR: 2019/05/13 16:17:03.799099 Settings OS_PASSWORD is unknown
panic: Settings OS_PASSWORD is unknown

goroutine 1 [running]:
log.(*Logger).Panic(0xc0000fe0f0, 0xc0002a3cd0, 0x1, 0x1)
        /home/travis/.gimme/versions/go1.11.9.linux.amd64/src/log/log.go:212 +0xbe
github.com/wal-g/wal-g/internal.verifyConfig()
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/config.go:62 +0x2f7
github.com/wal-g/wal-g/internal.init.1()
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/config.go:48 +0x21d


$ wal-g backup-list
ERROR: 2019/05/13 16:17:14.140655 Settings OS_AUTH_URL is unknown
panic: Settings PGUSER is unknown

goroutine 1 [running]:
log.(*Logger).Panic(0xc0000fe0f0, 0xc0002a3cd0, 0x1, 0x1)
        /home/travis/.gimme/versions/go1.11.9.linux.amd64/src/log/log.go:212 +0xbe
github.com/wal-g/wal-g/internal.verifyConfig()
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/config.go:62 +0x2f7
github.com/wal-g/wal-g/internal.init.1()
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/config.go:48 +0x21d

```

g0djan:
[...]

Issue #273, opened 2019-05-13
--config not working


Tri0L:
```
$ whoami
postgres

$ cd ~

$ ls -l
-rwxrwxr-x 1 postgres postgres  481 May 13 13:56 wal-g.config

$ wal-g --config ./wal-g.config backup-list 
Using config file: /var/lib/postgresql/wal-g.config
ERROR: 2019/05/13 15:39:30.429076 No storage is configured now, please set one of following settings: [WALE_S3_PREFIX WALE_FILE_PREFIX WALE_GS_PREFIX WALE_AZ_PREFIX WALE_SWIFT_PREFIX]

$ wal-g --config /var/lib/postgresql/wal-g.config backup-list 
Using config file: /var/lib/postgresql/wal-g.config
ERROR: 2019/05/13 15:39:30.429076 No storage is configured now, please set one of following settings: [WALE_S3_PREFIX WALE_FILE_PREFIX WALE_GS_PREFIX WALE_AZ_PREFIX WALE_SWIFT_PREFIX]

$ mv ./wal-g.config .walg.json

$ wal-g backup-list
name                                                     last_modified        wal_segment_backup_start
```

vmihailenco:
@x4m the problem seems to be that https://github.com/wal-g/wal-g/blob/master/internal/config.go#L113-L125 is called before viper is able to read anything in https://github.com/wal-g/wal-g/blob/master/main/configure.go#L29-L56

Tri0L:
Working only if `--config` is last argument.

```
#not working
wal-g --config ./wal-g.config backup-list

#working OK
wal-g backup-list --config ./wal-g.config
```

Issue #267, opened 2019-05-10
wal-g reading wrong config file


dragomirr:
When running wal-g it should be using `$HOME/.wal-g.yaml` as configuration file, as stated in help. Instead `$HOME/.walg.json` is used.
When running `wal-g --help`  this is output about config `--config string   config file (default is $HOME/.wal-g.yaml)`
Also when running wal-g it says it is using `$HOME/.wal-g.yaml` while actually does not read config file.
Link to code https://github.com/wal-g/wal-g/blob/master/internal/config.go#L74 thanks to @x4m

Also yaml syntax does not work, config file must be json formatted.
When using `.walg.json` config file with json syntax everything works as expected.

Issue #216, opened 2019-03-25
home directory detection


femaref:
postgres calls the archive/restore command with a very limited env, notably, `$PATH` and `$HOME` are missing. As such, the home dir detection of `github.com/mitchellh/go-homedir` fails, as it can't find `getent`. 

There is a fix for that (https://github.com/mitchellh/go-homedir/issues/18) but that's on in the v1.1.0 release.

Looking for the code, there are two other places where the homedir is used - `internal/crypto.go` and `internal/config.go`. Both use `os/user.Current`. Is there a reason why `cmd/root.go` uses the other way? Nonetheless, a workaround right now is to 
 use `HOME=/path/to/dir /usr/bin/wal-g`.

x4m:
Hi! Sorry for so long delay...
Yes, this go-home thing certainly  is a problem: usages should be unified. And if we really need go-homedir, it should be in gopkg.toml?

x4m:
@femaref can you please send a PR with behavior that you think is optimal?
I mean to choose  between go-homedir and os/user.Current?

x4m:
Should be fixed now, please reopen if this will reappear in 0.2.10

PR #276, opened 2019-05-14
Use viper config reader

x4m:
This should help to  #274 #267 #273 
Also added some allowed env vars.

Do we pass variables from config to env vars? We definitely should do that.

Also, we need integration tests for config files.


x4m:
Also this will help #216 since we get rid of homedir

PR #283, opened 2019-05-20
add --detail switch to backup-list to print meta

HeikoOnnebrink:
WAL-G creates and saves a metadata.json file during each `backup-push operation` 
The metadata file contains lots of valuable data that can be used by e.g. external automation of backup/restore processes.

This PR should add the functionality to retrieve this metadata during a `backup-list` operation 

For this purpose a new flag `--detail` is added to the signature of backup-list
It supports existing flags `--pretty` and `--json` to control the output style

In this way e.g. the retrieval of the PostgreSQL version of a backup from a shell script will be  possible (with a little help from jq command line util)

wal-g backup-list --detail --json | jq '.[] | select( .backup_name == "<some backupname>" ) | .pg_version  '

### DIFF
--- a/internal/backup_detail.go
+++ b/internal/backup_detail.go
@@ -1,0 +1,25 @@
+package internal
+
+// BackupDetails is used to append ExtendedMetadataDto details to BackupTime struct
+type BackupDetail struct {
+	BackupTime
+	ExtendedMetadataDto
+}
+
+// writeBackupTime sets matching elements in BackupDetail with values from BackupTime
+func (backupDetail *BackupDetail) writeBackupTime(backupTime BackupTime) {
+	backupDetail.BackupName = backupTime.BackupName
+	backupDetail.Time = backupTime.Time
+	backupDetail.WalFileName = backupTime.WalFileName
+}
+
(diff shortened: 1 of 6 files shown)


internal/backup_detail.go
```diff
@@ -0,0 +1,25 @@
+package internal
+
+// BackupDetails is used to append ExtendedMetadataDto details to BackupTime struct
+type BackupDetail struct {
+	BackupTime
+	ExtendedMetadataDto
+}
+
+// writeBackupTime sets matching elements in BackupDetail with values from BackupTime
+func (backupDetail *BackupDetail) writeBackupTime(backupTime BackupTime) {
```
Tinsane: Let's remove these two functions. Instead you can construct such an object with 'BackupDetail{backupTimeObj, ExtendedMetadataDtoObj}' statement.

HeikoOnnebrink: Thanks for feedback.. still learning GO.. fixed it as requested ..

Tinsane: Nice, thanks!

PR #285, opened 2019-05-21
Save decompressor when downloading WAL files

dimastark:


internal/databases/mysql/stream_fetch_handler.go
```diff
@@ -86,6 +86,7 @@ func fetchBinlogs(folder storage.Folder, sentinel StreamSentinelDto, binlogsAreD
 		return
 	}
 	var fetchedLogs []storage.Object
+	var possibleDecompressors []compression.Decompressor
```
dimastark: Is it possible that multiple compression formats will be used when fetching WAL files?

x4m: For any single WAL file only one decompressor will be used. For multiple WALs this does not hold.

PR #286, opened 2019-05-23
Refactored wal-g settings and added default config values where it is possible

Tinsane:


internal/configure.go
```diff
@@ -198,28 +240,71 @@ func ConfigureUploader() (uploader *Uploader, err error) {
...
+	if err != nil {
+		return 1, err
+	}
+	if con < 1 {
+		return 1, NewInvalidConcurrencyValueError(con)
+	}
+	return con, nil
+}
+
+func GetSentinelUserData() interface{} {
```
g0djan: I guess the format of sentinel is defined. So there could be a better way than returning interface{}.

internal/config.go
```diff
@@ -4,42 +4,116 @@ import (
...
+	CompressionMethodSetting     = "WALG_COMPRESSION_METHOD"
+	DiskRateLimitSetting         = "WALG_DISK_RATE_LIMIT"
+	NetworkRateLimitSetting      = "WALG_NETWORK_RATE_LIMIT"
+	UseWalDeltaSetting           = "WALG_USE_WAL_DELTA"
+	LogLevelSetting              = "WALG_LOG_LEVEL"
+	TarSizeThresholdSetting      = "WALG_TAR_SIZE_THRESHOLD"
+	GpgKeyIDSetting              = "GPG_KEY_ID"
+	PgpKeySetting                = "WALG_PGP_KEY"
+	PgpKeyPathSetting            = "WALG_PGP_KEY_PATH"
+	PgpKeyPassphraseSetting      = "WALG_PGP_KEY_PASSPHRASE"
```
g0djan: "Setting" suffix looks spare. You use it to getting information that it comes from env variables.
It would be more explicit to create a struct Environment {}. 
For example, use it as env.DownloadConcurrency.

internal/databases/mysql/mysql.go
```diff
@@ -19,12 +19,13 @@ import (
...
-	BinlogDst    = "WALG_MYSQL_BINLOG_DST"
-	BinlogSrc    = "WALG_MYSQL_BINLOG_SRC"
-	SslCa        = "WALG_MYSQL_SSL_CA"
+	StreamPrefix          = "stream_"
+	BinlogPath            = "binlog_" + utility.VersionStr + "/"
+	DatasourceNameSetting = "WALG_MYSQL_DATASOURCE_NAME"
+	BinlogEndTsSetting    = "WALG_MYSQL_BINLOG_END_TS"
+	BinlogDstSetting      = "WALG_MYSQL_BINLOG_DST"
+	BinlogSrcSetting      = "WALG_MYSQL_BINLOG_SRC"
+	SslCaSetting          = "WALG_MYSQL_SSL_CA"
```
g0djan: "Setting" suffix again

internal/delete_handler.go
```diff
@@ -17,10 +17,9 @@ const (
 	NoDeleteModifier = iota
 	FullDeleteModifier
 	FindFullDeleteModifier
-	ConfirmFlag = "confirm"
+	ConfirmFlag            = "confirm"
```
g0djan: Spare line in diff.

internal/delete_handler.go
```diff
@@ -17,10 +17,9 @@ const (
...
 	DeleteShortDescription = "Clears old backups and WALs"
 
-
```
g0djan: Spare line in diff

test/crypto_comp_test.go
```diff
@@ -13,16 +14,20 @@ import (
...
 
 	"github.com/stretchr/testify/assert"
-	"github.com/wal-g/wal-g/internal"
 	"github.com/wal-g/wal-g/internal/compression/lzo"
 	"github.com/wal-g/wal-g/internal/crypto/openpgp"
 )
 
 const waleWALfilename = "testdata/000000010000000000000024.lzo"
 
+func noPassphrase() (string, bool) {
```
g0djan: Move it after constants.

internal/config.go
```diff
@@ -4,42 +4,116 @@ import (
...
+		CompressionMethodSetting:     nil,
+		DiskRateLimitSetting:         nil,
+		NetworkRateLimitSetting:      nil,
+		UseWalDeltaSetting:           nil,
+		LogLevelSetting:              nil,
+		TarSizeThresholdSetting:      nil,
+		"USER":                       nil, // TODO : do something with it
+		"PGPORT":                     nil, // TODO : do something with it
+		"PGUSER":                     nil, // TODO : do something with it
+		"PGHOST":                     nil, // TODO : do something with it
```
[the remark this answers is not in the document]
Tinsane: We don't use them now, so it's better to wait for usage because there is a possibility that we will just delete them.

PR #290, opened 2019-05-27
fix wal-g delete and no backups found exit code

HeikoOnnebrink:
When `WAL-G DELETE <options>` is called and no backup matches the delete criteria nothing gets deleted but WAL-G returns with **exit code 1** and "nil" in the output

In order to distinguish these "nothing to delete errors" from real errors WAL-G should return in this situation with **exit code 0** and some proper INFO output instead

```
wal-g backup-list
name                          last_modified        wal_segment_backup_start
base_000000010000000000000008 2019-05-27T09:21:33Z 000000010000000000000008
base_00000001000000000000000A 2019-05-27T09:23:25Z 00000001000000000000000A
base_00000001000000000000000E 2019-05-27T09:49:47Z 00000001000000000000000E

wal-g delete retain FULL 5
INFO: 2019/05/27 15:20:17.155260 No backup found for deletion
echo $? 
0

```

### DIFF
--- a/internal/backup_fetch_handler.go
+++ b/internal/backup_fetch_handler.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"fmt"
+
 	"github.com/pkg/errors"
 	"github.com/wal-g/wal-g/internal/storages/storage"
 	"github.com/wal-g/wal-g/internal/tracelog"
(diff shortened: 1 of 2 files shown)


Tinsane:
Hi! Good PR, seems logical to me. Thank you for contribution!

PR #294, opened 2019-05-29
Fixed config initialization. Also fixed initialization from config for S3 storage.

Tinsane:


internal/config.go
```diff
@@ -104,46 +76,46 @@ func GetWaleCompatibleSetting(key string) (value string, exists bool) {
...
-func readConfig() {
-	cfg := make(map[string]string)
-	WalgConfig = &cfg
-	for _, key := range viper.AllKeys() {
-		cfg[key] = viper.GetString(key)
+	// If a config file is found, read it in.
+	err := viper.ReadInConfig()
+	//if err != nil {
+	//	tracelog.WarningLogger.Printf("Failed to use config file because of error: %v\n", err)
+	//}
```
g0djan: Looks like forgotten to delete

internal/config.go
```diff
@@ -104,46 +76,46 @@ func GetWaleCompatibleSetting(key string) (value string, exists bool) {
...
+// initConfig reads in config file and ENV variables if set.
+func InitConfig() {
+	if CfgFile != "" {
+		// Use config file from the flag.
+		viper.SetConfigFile(CfgFile)
+	} else {
+		// Find home directory.
+		usr, err := user.Current()
+		if err != nil {
+			fmt.Println(err)
```
g0djan: tracelog.ErrorLogger.FatalError(err)

internal/configure.go
```diff
@@ -70,8 +69,8 @@ type InvalidConcurrencyValueError struct {
 	error
 }
 
-func NewInvalidConcurrencyValueError(value int) InvalidConcurrencyValueError {
-	return InvalidConcurrencyValueError{errors.Errorf("Concurrency value is expected to be positive but is: %v", value)}
+func NewInvalidConcurrencyValueError(concurrency string, value int) InvalidConcurrencyValueError {
```
g0djan: Is a type of setting "concurency" is a string? Or there should be the name of any setting?

g0djan: In the first case, let's make a comment.

g0djan: In the second case. It should be more common Error and be placed in a different package.

PR #295, opened 2019-05-29
fix to avoid different behavior on mac and linux when we using time.T…

g0djan:
…ime for backups metafile

### DIFF
--- a/internal/backup_push_handler.go
+++ b/internal/backup_push_handler.go
@@ -110,7 +110,7 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string) {
 	bundle := NewBundle(archiveDirectory, crypter, previousBackupSentinelDto.BackupStartLSN, previousBackupSentinelDto.Files)
 
 	var meta ExtendedMetadataDto
-	meta.StartTime = ceilTimeUpToMicroseconds(time.Now())
+	meta.StartTime = utility.CeilTimeUpToMicroseconds(time.Now())
 	meta.Hostname, _ = os.Hostname()
 
 	// Connect to postgres and start/finish a nonexclusive backup.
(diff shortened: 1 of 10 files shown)


internal/backup_sentinel_dto.go
```diff
@@ -23,13 +23,14 @@ type BackupSentinelDto struct {
...
-	StartTime  time.Time `json:"start_time"`
-	FinishTime time.Time `json:"finish_time"`
-	Hostname   string    `json:"hostname"`
-	DataDir    string    `json:"data_dir"`
-	PgVersion  int       `json:"pg_version"`
-	StartLsn   uint64    `json:"start_lsn"`
-	FinishLsn  uint64    `json:"finish_lsn"`
+	StartTime      time.Time `json:"start_time"`
+	FinishTime     time.Time `json:"finish_time"`
+	DatetimeFormat string    `json:"date_fmt"`
```
Tinsane: date_fmt -> datetime_fmt?

internal/backup_push_handler.go
```diff
@@ -232,3 +233,10 @@ func UploadSentinel(uploader *Uploader, sentinelDto *BackupSentinelDto, backupNa
 
 	return uploader.Upload(sentinelName, bytes.NewReader(dtoBody))
 }
+
+// This function is needed for being cross-platform
+func ceilTimeUpToMicroseconds(timeToCeil time.Time) time.Time {
```
Tinsane: 1. Let's unit test it
2. It doesn't work correctly when timeToCeil.Nanosecond() % 1000 == 0

Tinsane:
I looked through your changes. You fixed not all occurrences of time.Now(). Is it ok?

internal/backup_push_handler.go
```diff
@@ -110,15 +110,16 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string) {
 	bundle := NewBundle(archiveDirectory, crypter, previousBackupSentinelDto.BackupStartLSN, previousBackupSentinelDto.Files)
 
 	var meta ExtendedMetadataDto
-	meta.StartTime = time.Now()
+	meta.StartTime = utility.CeilTimeUpToMicroseconds(time.Now())
```
Tinsane: What about a single function for:
utility.CeilTimeUpToMicroseconds(time.Now())
?

PR #296, opened 2019-05-30
WIP: Implemented Redis backup-push

BurningMarshmallow:
Added support of push command for Redis database:
Using go-redis to connect to Redis and execute BGSAVE command, after that taking dump.rdb(when it changes) and archiving it
Made changes similar to MySQL support - extracted cmd, docker containers and wrote simple integration test in Docker environment


BurningMarshmallow:
Note: I marked debatable places of code with //DISCUSS comment, you could write what you think about them as code comments

cmd/redis/redis.go
```diff
@@ -0,0 +1,35 @@
...
+
+// These variables are here only to show current version. They are set in makefile during build process
+var WalgVersion = "devel"
+var GitRevision = "devel"
+var BuildDate = "devel"
+
+var RedisCmd = &cobra.Command{
+	Use:     "redis",
+	Short:   RedisShortDescription, // TODO : improve description
+	Version: WalgVersion + "\t" + GitRevision + "\t" + BuildDate + "\t" + "Redis",
```
Tinsane: https://golang.org/pkg/strings/#Join

internal/databases/redis/redis.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func getRedisConnection() *redis.Client {
+	redisAddr := internal.GetSettingWithLocalDefault("WALG_REDIS_HOST", "localhost")
+	redisPort := internal.GetSettingWithLocalDefault("WALG_REDIS_PORT", "6379")
+	redisPassword := internal.GetSettingWithLocalDefault("WALG_REDIS_PASSWORD", "") // no password set
+	redisDbStr, ok := internal.GetSetting("WALG_REDIS_DB")
+	redisDb := 0 // use default DB
+	if ok {
+		redisDbValue, err := strconv.Atoi(redisDbStr) // DISCUSS: could redisDb changed on success without additional variable redisDbValue?
```
Tinsane: Don't mind. I'll rewrite it with the new config in parallel PR.

internal/databases/redis/backup_push_handler.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func uploadBackup(uploader *Uploader, redisDataFoler string) {
+	f, err := os.Open(redisDataFoler + "/dump.rdb")
+	if err != nil {
+		tracelog.ErrorLogger.Fatalf("%+v\n", err)
+	}
+	defer utility.LoggedClose(f, "")
+	reader := bufio.NewReader(f)
+	namedReader := internal.NewNamedReaderImpl(reader, "base_"+time.Now().Format(time.RFC3339)) // DISCUSS: Is it correct backup name?
```
Tinsane: We prefer to name backups, thinking of DB inner structure. For PG we use LSN - logical sequence number (logical time for DB). For MySQL, we have to use ordinary time just because we can't use anything else. So, if for Redis there is some simple analog for LSN, then let's use it in backup names instead.

internal/databases/redis/backup_push_handler.go
```diff
@@ -0,0 +1,80 @@
...
+func initializeBackupSaving(client *redis.Client) {
+	err := client.BgSave().Err()
+	if err != nil {
+		tracelog.ErrorLogger.Fatalf("%+v\n", err)
+	}
+}
+
+func waitForNewBackup(currentTime time.Time, redisDataFoler string) {
+	for {
+		if _, err := os.Stat(redisDataFoler + "/dump.rdb"); !os.IsNotExist(err) {
```
Tinsane: Let's add some sleeps to this cycle to not waste CPU.

internal/databases/redis/backup_push_handler.go
```diff
@@ -0,0 +1,80 @@
...
+			break
+		}
+	}
+
+	for {
+		stat, err := os.Stat(redisDataFoler + "/dump.rdb")
```
Tinsane: Let's add some sleeps to this cycle to not waste CPU.

internal/databases/redis/backup_push_handler.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			tracelog.ErrorLogger.Fatalf("%+v\n", err)
+		}
+
+		if math.Abs(currentTime.Sub(stat.ModTime()).Seconds()) < 10 {
```
Tinsane: Let's name this "magical constant" somehow

internal/databases/redis/backup_push_handler.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Get Redis connection
+	client := getRedisConnection()
+
+	// Init backup process
+	currentTime := time.Now()
+	initializeBackupSaving(client)
+
+	// Wait for new backup on disk.
+	// DISCUSS: Maybe add Timer to stop trying after some time?
```
Tinsane: 1. We discussed Redis dumping to stdout, is there something wrong with stream-push method?
2. Can't we await for this event using redis library?
3. If everything abovementioned doesn't work, then let's do it.

docker/redis_tests/walg.env
```diff
@@ -0,0 +1,12 @@
+PGSSLMODE=allow
```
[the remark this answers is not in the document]
BurningMarshmallow: Hm, if I understand correctly this variable needed only by pg_tests and overwritten if it is not set in other environment files? In that case, we can specify environment variable in docker-compose file https://docs.docker.com/compose/environment-variables/
It will be like
```
    environment:
      - PGSSLMODE=allow
```
I tested it with removing PGSSLMODE variable from redis_tests - it works.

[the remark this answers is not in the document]
Tinsane: It was a comment related to the whole file :)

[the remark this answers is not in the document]
BurningMarshmallow: Oh, I see :). Yeah, sure, we can use common environment file, I implemented it in my PR, it really looks better that way.

internal/databases/redis/backup_push_handler.go
BurningMarshmallow: As far as i'm concerned, there is no analog of LSN in Redis, I suppose it is connected with overall simplicity of Redis as plain key-value storage. 
By the way, is RFC3339 format appropriate for backup suffix, we don't need to have naming with precision to nanoseconds, right?

Tinsane: Not so long ago @g0djan resolved this issue with the precision of time. Let's wait, what he will tell here.

PR #299, opened 2019-05-31
[WIP] Filter unnecessary regular files during backup fetch

tonyfresher:


internal/backup_sentinel_dto.go
```diff
@@ -14,6 +14,7 @@ type BackupSentinelDto struct {
 	IncrementCount    *int    `json:"DeltaCount,omitempty"`
 
 	Files BackupFileList `json:"Files"`
+	TarFileSets map[string][]string `json:"TarFileSets"`
```
g0djan: option+cmd+L in GoLand

g0djan: option+cmd+L in GoLand

PR #302, opened 2019-06-07
add permanent check for wals in delete

pohzipohzi:


x4m:
Hi! Thanks for the PR, this is an important feature.
I want to note that you have a very inefficient search in isPermanent
Can you improve it?
1. use array of LSN starts and ends
2. use hashtable instead of an array

Tinsane:
Seems good to me.

PR #303, opened 2019-06-07
mongodb extension

akulin:
implemented extension to manage mongodb backups. Yes, commands are very similar to MySQL commands.


cmd/mongo/delete.go
```diff
@@ -0,0 +1,82 @@
...
+		time1, ok1 := tryFetchTimeRFC3999(object1)
+		time2, ok2 := tryFetchTimeRFC3999(object2)
+		if !ok1 || !ok2 {
+			return object2.GetLastModified().After(object1.GetLastModified())
+		}
+		return time1 < time2
+	}
+}
+
+func tryFetchTimeRFC3999(object storage.Object) (string, bool) {
```
Tinsane: This function seems to repeat cmd/mysql/delete.go:tryFetchTimeRFC3999
Let's move both of them to utility or to some other common library, but not to repeat their code.

internal/databases/mongo/stream_fetch_handler.go
```diff
@@ -2,141 +2,57 @@ package mongo
...
-}
-
 func fetchOplogs(folder storage.Folder, startTime time.Time, oplogAreDone chan error) {
-	oplogFolder := folder.GetSubFolder(OplogPath)
-	endTS, oplogDstFolder := getOplogConfigs()
-	if oplogDstFolder == "" {
-		oplogAreDone <- errors.New("WALG_MONGO_OPLOG_DST is not configured")
+	endTS, oplogDstFolder, err := getOplogConfigs()
+	if err != nil {
+		oplogAreDone <- nil
```
Tinsane: Why are we suppressing error here?

internal/databases/mongo/stream_fetch_handler.go
```diff
@@ -2,141 +2,57 @@ package mongo
...
 		return
 	}
-	oplogFiles, _, err := oplogFolder.ListFolder()
+	oplogFolder := folder.GetSubFolder(OplogPath)
+	logsToFetch, err := internal.GetOperationLogsCoveringInterval(oplogFolder, startTime, endTS)
 	if err != nil {
-		oplogAreDone <- err
+		oplogAreDone <- nil
```
Tinsane: Why are we suppressing error here?

akulin: Wow! Unbelievable bug! Fixed now

Tinsane:
Thank you for your contribution!

PR #305, opened 2019-06-08
add permanent flag in backup push

pohzipohzi:


Tinsane:
Hi!
As you add a new flag to delete handler here, let's do two more things in this PR:
1. Let's write about permanent backups feature in README.
2. Let's add an integration test for it.

docker/pg_tests/scripts/tests/delete_before_permanent_test.sh
```diff
@@ -0,0 +1,46 @@
...
+
+/usr/lib/postgresql/10/bin/initdb ${PGDATA}
+
+echo "archive_mode = on" >> /var/lib/postgresql/10/main/postgresql.conf
+echo "archive_command = '/usr/bin/timeout 600 /usr/bin/wal-g wal-push %p'" >> /var/lib/postgresql/10/main/postgresql.conf
+echo "archive_timeout = 600" >> /var/lib/postgresql/10/main/postgresql.conf
+
+/usr/lib/postgresql/10/bin/pg_ctl -D ${PGDATA} -w start
+
+# push first backup as permanent
```
Tinsane: This way you check only full backup permanence. Add another test scenario that will check delta backup permanence.
Let it be something like:
1. Make several non-permanent backups.
2. Make permanent delta backup.
3. Add some data to DB.
4. Make another non-permanent delta backup.
5. Run 'delete' that will delete the latest backup.
6. Restore LATEST backup and check that DB has the same state that it had before permanent backup.

internal/uploader.go
```diff
@@ -103,3 +109,17 @@ func (uploader *Uploader) Upload(path string, content io.Reader) error {
 	tracelog.ErrorLogger.Printf(tracelog.GetErrorFormatter()+"\n", err)
 	return err
 }
+
+// UploadMultiple uploads multiple objects, returning the first error if any.
+// Note that this is not atomic thus potentially dangerous
+// TODO : unit tests
```
Tinsane: Let's unit test it.

internal/backup_mark.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/wal-g/wal-g/internal/storages/storage"
+	"github.com/wal-g/wal-g/internal/tracelog"
+	"github.com/wal-g/wal-g/utility"
+)
+
+// MarkSelfAndPreviousBackupsPermanent marks all previous related backups
+// permanent, including itself, any previous delta backups and initial full
+// backup
+// TODO: unit tests
```
Tinsane: Let's unit test it.

internal/backup_push_handler.go
```diff
@@ -192,6 +192,13 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
 		tracelog.ErrorLogger.Printf("Failed to upload sentinel file for backup: %s", backupName)
 		tracelog.ErrorLogger.FatalError(err)
 	}
+
+	if isPermanent {
```
Tinsane: Does it work properly in the case when the current backup is full?

internal/backup_mark.go
```diff
@@ -0,0 +1,81 @@
...
+			toUpload = append(toUpload, uploadObject)
+		}
+	}
+
+	tracelog.InfoLogger.Printf("Marking permanent backups: %v\n", toUpload)
+	return uploader.UploadMultiple(toUpload)
+}
+
+func getMetadataUploadObject(backupName string, meta ExtendedMetadataDto) (UploadObject, error) {
+	metaFile := storage.JoinPath(backupName, utility.MetadataFileName)
```
Tinsane: metaFile -> metaFilePath

internal/backup_mark.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// MarkSelfAndPreviousBackupsPermanent marks all previous related backups
+// permanent, including itself, any previous delta backups and initial full
+// backup
+// TODO: unit tests
+func MarkSelfAndPreviousBackupsPermanent(uploader *Uploader, baseBackupFolder storage.Folder, backupName string) error {
+	toUpload := []UploadObject{}
+
+	// mark self if impermanent
```
Tinsane: 'if -> as'?

internal/backup_mark.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/wal-g/wal-g/internal/storages/storage"
+	"github.com/wal-g/wal-g/internal/tracelog"
+	"github.com/wal-g/wal-g/utility"
+)
+
+// MarkSelfAndPreviousBackupsPermanent marks all previous related backups
+// permanent, including itself, any previous delta backups and initial full
+// backup
+// TODO: unit tests
+func MarkSelfAndPreviousBackupsPermanent(uploader *Uploader, baseBackupFolder storage.Folder, backupName string) error {
```
Tinsane: Let's make this function recursive to simplify it.

internal/backup_mark.go
pohzipohzi: not really, i wanted to mean "mark if not already permanent" to avoid uploading objects unnecessarily. This is also useful when we eventually add the `makeBackupPermanent` command.

internal/backup_push_handler.go
pohzipohzi: actually it doesn't, let me fix that...

internal/backup_mark.go
pohzipohzi: changing the function to recursive will depend on all previous backups to be uncorrupted in order to work, and that may or may not cause some undesirable behavior. On the other hand doing this iteratively limits the number of iterations to `currentSentinel.IncrementCount`, which is not great either but i think can be slightly better. What do you think?

internal/backup_mark.go
Tinsane: Yes, I misread this, it's ok.

internal/backup_push_handler.go
```diff
@@ -193,7 +193,7 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
 		tracelog.ErrorLogger.FatalError(err)
 	}
 
-	if isPermanent {
+	if isPermanent && previousBackupSentinelDto.BackupStartLSN != nil {
```
Tinsane: 'previousBackupSentinelDto.BackupStartLSN != nil' means only that previous backup was done by the version which doesn't support increments. We shouldn't enter this branch at all if the current backup is full.

internal/backup_mark.go
Tinsane: If one or more of previous backups is corrupted, then the current backup is useless too. So recursion seems to be OK here.

internal/backup_mark.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/json"
+
+	"github.com/wal-g/wal-g/internal/storages/storage"
+	"github.com/wal-g/wal-g/utility"
+)
+
+// GetImpermanentBackupsBefore marks all previous related backups permanent,
+// including itself, any previous delta backups and initial full backup
+// TODO: unit tests
+func GetImpermanentBackupsBefore(baseBackupFolder storage.Folder, backupName string, toUpload *[]UploadObject) error {
```
Tinsane: toUpload -> impermanentBackupMetadatas
Let's also return this list instead of passing it by the pointer.

internal/backup_push_handler.go
```diff
@@ -193,6 +193,21 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string) {
 		tracelog.ErrorLogger.Printf("Failed to upload sentinel file for backup: %s", backupName)
 		tracelog.ErrorLogger.FatalError(err)
 	}
+	// If pushing permanent delta backup, mark all previous backups permanent as well
+	if isPermanent && currentBackupSentinelDto.IsIncremental() {
```
Tinsane: There are several thoughts about metadata files:
1. Now metadata contains vital info about backup permanence property so we should fail if we can't upload it.
2. We should upload metadata files in order of backups. It's gonna be a fail if we mark as permanent only last delta backup but don't mark as permanent its base.

internal/backup_push_handler.go
```diff
@@ -193,18 +193,19 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
...
-	// If pushing permanent delta backup, mark all previous backups permanent as well
+	// If pushing permanent delta backup, mark all previous backups permanent.
 	if isPermanent && currentBackupSentinelDto.IsIncremental() {
 		tracelog.InfoLogger.Printf("Retrieving previous related backups to be marked as permanent")
-		toUpload := []UploadObject{}
-		err := GetImpermanentBackupsBefore(basebackupFolder, previousBackupName, &toUpload)
+		impermanentBackupMetadata, err := GetImpermanentBackupMetadataBefore(basebackupFolder, previousBackupName)
 		if err != nil {
 			tracelog.ErrorLogger.Printf("Failed to get previous backups: %v", err)
+			return
```
Tinsane: Here Fatalf should be used instead of Printf to have non-zero exit code.

internal/backup_push_handler.go
```diff
@@ -193,18 +193,19 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
...
 		} else {
-			tracelog.InfoLogger.Printf("Retrieved backups to mark as permanent, marking: %v", toUpload)
-			err = uploader.UploadMultiple(toUpload)
+			tracelog.InfoLogger.Printf("Retrieved backups to mark as permanent, marking: %v", impermanentBackupMetadata)
+			err = uploader.UploadMultiple(impermanentBackupMetadata)
 			if err != nil {
 				tracelog.ErrorLogger.Printf("Failed to mark previous backups as permanent: %v", err)
+				return
```
Tinsane: Here also Fatalf should be used instead of Printf to have non-zero exit code.

PR #306, opened 2019-06-08
Add unit-tests for utility.FastCopy

kadukm:
Add unit-tests for FastCopy method in utility package. Add auxiliary structs (ErrorWriter and ErrorReader) into testtools package for this purpose.
Also apply some refactoring in utility package: use CopiedBlockMaxSize const instead of CompressedBlockMaxSize in FastCopy method.

### DIFF
--- a/test/utility_test.go
+++ b/test/utility_test.go
@@ -1,7 +1,9 @@
 package test
 
 import (
+	"bytes"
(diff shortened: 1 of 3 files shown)


Tinsane:
Nice PR! Thank you for your contribution.

PR #313, opened 2019-06-19
Refactoring of tests

kadukm:
Made a refactoring of tests, namely:
- moved some general functionality from test package to testtools or utility
- deleted duplicate code (such as function `Min` and struct `ErrorReader`)

I've made this PR because I want to fix coverage for tests (now coverage described in README doesn't work). It will be simpler for you to review (and merge) this two PRs separetely.
My next PR will contain only changes regarding that fix.

### DIFF
--- a/test/backup_fetch_handler_test.go
+++ b/test/backup_fetch_handler_test.go
@@ -3,39 +3,40 @@ package test
 import (
 	"github.com/stretchr/testify/assert"
 	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/testtools"
 	"testing"
 )
 
 func TestGetRestoredBackupFilesToUnwrap_SimpleFile(t *testing.T) {
 	sentinelDto := internal.BackupSentinelDto{
-		Files: NewBackupFileListBuilder().WithSimple().Build(),
+		Files: testtools.NewBackupFileListBuilder().WithSimple().Build(),
 	}
 
 	files := internal.GetRestoredBackupFilesToUnwrap(sentinelDto)
-	assert.Contains(t, files, SimplePath)
+	assert.Contains(t, files, testtools.SimplePath)
 }
 
 func TestGetRestoredBackupFilesToUnwrap_IncrementedFile(t *testing.T) {
 	sentinelDto := internal.BackupSentinelDto{
-		Files: NewBackupFileListBuilder().WithIncremented().Build(),
+		Files: testtools.NewBackupFileListBuilder().WithIncremented().Build(),
 	}
 
 	files := internal.GetRestoredBackupFilesToUnwrap(sentinelDto)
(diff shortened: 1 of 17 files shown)


test/backup_fetch_handler_test.go
```diff
@@ -3,39 +3,40 @@ package test
...
 
 func TestGetRestoredBackupFilesToUnwrap_SimpleFile(t *testing.T) {
 	sentinelDto := internal.BackupSentinelDto{
-		Files: NewBackupFileListBuilder().WithSimple().Build(),
+		Files: testtools.NewBackupFileListBuilder().WithSimple().Build(),
 	}
 
 	files := internal.GetRestoredBackupFilesToUnwrap(sentinelDto)
-	assert.Contains(t, files, SimplePath)
+	assert.Contains(t, files, testtools.SimplePath)
```
g0djan: SimplePath used only in tests of backup_fetch_handler. What is the reason to move it to testtools package? Also "testtools" is strange package because package should be independent and reusable entity. https://medium.com/rungo/everything-you-need-to-know-about-packages-in-go-b8bac62b74cc

test/backup_fetch_handler_test.go
```diff
@@ -3,39 +3,40 @@ package test
...
 }
 
 func TestGetRestoredBackupFilesToUnwrap_IncrementedFile(t *testing.T) {
 	sentinelDto := internal.BackupSentinelDto{
-		Files: NewBackupFileListBuilder().WithIncremented().Build(),
+		Files: testtools.NewBackupFileListBuilder().WithIncremented().Build(),
```
g0djan: NewBackupFileListBuilder used only in backup_fetch_handler. I think you can find all such places in your code by yourself.

kadukm:
@g0djan, yes, I wanted to do what poezi wrote but as you said testtools package now depends on a lot of other internal packages. If move tests from test package to other packages then import cycle appears (such as `internal->testtools->internal` or `utility->testtools->utility`). So i decided move code from testtools package into `*packagename*_test` packages. It won't allow white-box testing but command `go test -coverprofile=c.out` will work properly and the rest of the project structure will remain unchanged.
Of course, problem with import cycle can be solved by moving code from testtools into appropriate packages. I didn't want to do it because it's not obviously where to place the code from testtools. I can place general functionality somewhere in different packages in `*someentity*_test.go` files but I'm afraid that after that action new contributors won't be able to find this general functionality and start write their own duplicate code. Also in this case we have to place mock objects in internal packages (otherwise we have import cycle such as `internal->mock->internal`), what seemed to me not an attractive way.

Can you please advise how to do better?

g0djan:
@kadukm your goal is fixing coverage now so advise - just fix coverage :)
Also if you want you can try to surf github golang repositories and try to find some cool solutions for tests packages.

PR #314, opened 2019-06-20
Make TOTAL_BG_UPLOADER_LIMIT custom variable

reshke:


internal/configure.go
```diff
@@ -271,3 +271,9 @@ func GetSentinelUserData() interface{} {
 	}
 	return out
 }
+
+func GetTotalBgUploadedLimit() int32 {
```
Tinsane: We don't need a whole function for casting...

test/configure_test.go
```diff
@@ -55,3 +55,8 @@ func TestGetSentinelUserData(t *testing.T) {
 	t.Log(data)
 	assert.NotNilf(t, data, "Unable to parse WALG_SENTINEL_USER_DATA")
 }
+
+func TestGetTotalBgUploadLimit(t *testing.T) {
```
Tinsane: Viper testing is a work of viper developers, don't you think so?

internal/bguploader.go
```diff
@@ -117,7 +116,7 @@ func (bgUploader *BgUploader) scanOnce() {
 }
 
 func (bgUploader *BgUploader) shouldKeepScanning() bool {
-	return atomic.LoadInt32(&bgUploader.maxParallelWorkers) > 0 && atomic.LoadInt32(&bgUploader.totalUploaded) < TotalBgUploadedLimit
+	return atomic.LoadInt32(&bgUploader.maxParallelWorkers) > 0 && atomic.LoadInt32(&bgUploader.totalUploaded) < int32(viper.GetInt(TotalBgUploadedLimit))
```
Tinsane: https://godoc.org/github.com/spf13/viper#Viper.GetInt32

PR #315, opened 2019-06-20
Add aws/kms + minio/sio symmetric client side encryption.

bn0ir:
Add aws/kms + minio/sio symmetric client side encryption.

`WALG_CSE_KMS_ID` env var specifies KMS key id.
`WALG_CSE_KMS_REGION` env var specifies KMS key region.

Encryption:

- Generate unique symmetric key for every backup;
- Encrypt key with AWS KMS;
- Encrypt backup with symmetric key (minio/sio DARE 32 bytes key);
- Add encrypted symmetric key as backup file prefix (184 bytes);

Decryption:

- Read encrypted symmetric key from file prefix (184 bytes);
- Decrypt symmetric key with AWS KMS;
- Decrypt file with symmetric key (minio/sio DARE 32 bytes key);


README.md
```diff
@@ -116,6 +116,13 @@ To enable S3 server-side encryption, set to the algorithm to use when storing th
 
 If using S3 server-side encryption with `aws:kms`, the KMS Key ID to use for object encryption.
 
+* `WALG_CSE_KMS_ID`
```
x4m: These settings, they are S3-only, aren't they? If they are, I think we should use WALG_S3_CSE_...

internal/crypto/awskms/crypter.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return ioextensions.NewOnCloseFlusher(encryptedWriter, bufferedWriter), nil
+}
+
+// Decrypt creates decrypted reader from ordinary reader
+func (crypter *Crypter) Decrypt(reader io.Reader) (io.Reader, error) {
+	encryptedSymmetricKey := make([]byte, crypter.SymmetricKey.GetEncryptedKeyLen())
+	_, err := reader.Read(encryptedSymmetricKey)
+	if err != nil {
+		tracelog.ErrorLogger.Printf("Can't read encryption key from s3: %v", err)
```
x4m: I think we should better wrap error with our description...

internal/crypto/awskms/crypter.go
```diff
@@ -0,0 +1,71 @@
...
+		return reader, err
+	}
+
+	crypter.SymmetricKey.SetEncryptedKey(encryptedSymmetricKey)
+
+	err = crypter.SymmetricKey.Decrypt()
+	if err != nil {
+		tracelog.ErrorLogger.Printf("Can't decrypt symmetric key: %v", err)
```
x4m: Is this fatal? Looks like we continue anyway....

README.md
bn0ir: In most cases you want to use it with s3, but they can be used with any storage backend (you can use AWS KMS encryption and GCP storage)

internal/crypto/awskms/crypter.go
bn0ir: In most cases this error will be fatal, minio/sio will fail with error

internal/crypto/awskms/crypter.go
bn0ir: Fix error message in https://github.com/wal-g/wal-g/pull/315/commits/360633709839d3e0adda77fc760237d8fe924fb4

internal/crypto/awskms/crypter.go
bn0ir: Fix in https://github.com/wal-g/wal-g/pull/315/commits/f36b26391cf3aa6c3c467411da9d33af699641a7

x4m:
Thanks! Pushed

PR #321, opened 2019-07-05
Fix AWS_SECURITY_TOKEN vs AWS_SESSION_TOKEN thinko

fdr:
The naming of AWS_SECURITY_TOKEN was an educated guess in the WAL-E
era, but it was wrong: more references began to appear that preferred
AWS_SESSION_TOKEN. And indeed, WAL-G *only* respects the latter as-is,
for that reason:

    $ git grep AWS_SESSION_TOKEN internal/
    internal/storages/s3/folder.go: SessionTokenSetting      = "AWS_SESSION_TOKEN"

Rectify the documentation and one other reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ To connect to Amazon S3, WAL-G requires that this variable be set:
 
 * `WALG_S3_PREFIX` (eg. `s3://bucket/path/to/folder`) (alternative form `WALE_S3_PREFIX`)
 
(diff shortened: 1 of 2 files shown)


fdr:
Also, hi, long time no see. WAL-G has really grown!

x4m:
Hi! Thanks! Documentation is our weak point number 1. I'll see what's with CI and merge PR.

BTW, this code which copes with AWS settings is somewhat legacy, actually, many storage providers like GCP maintain their own env vars parsing. So does AWS, as far as I know.

fdr:
Regrettably, "make" also hangs for me, after getting quite a few dependencies installed, in the integration tests. I advise not using those too heavily, relative to getting faster and more precise mocks in.

PR #322, opened 2019-07-08
Temporary test speedup

Tinsane:
Now tests are too slow. It is a temporary solution.


x4m:
Can we revert it now? I've stuck on it in #361

PR #327, opened 2019-07-08
Selective backup-fetch for file mask

Tinsane:
Ability to fetch only files that satisfy given file mask.

### DIFF
--- a/utility/utility.go
+++ b/utility/utility.go
@@ -194,3 +194,21 @@ func ConcatByteSlices(a []byte, b []byte) []byte {
 	copy(result[len(a):], b)
 	return result
 }
+
+func SelectMatchingFiles(fileMask string, filePaths map[string]bool) (map[string]bool, error) {
+	if fileMask == "" {
+		return filePaths, nil
+	}
+	fileMask = "/" + fileMask
+	result := make(map[string]bool)
+	for filePath := range filePaths {
+		matches, err := filepath.Match(fileMask, filePath)
+		if err != nil {
+			return nil, err
+		}
+		if matches {
+			result[filePath] = true
+		}
+	}
+	return result, nil
+}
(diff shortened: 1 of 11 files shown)


utility/utility.go
```diff
@@ -194,3 +194,21 @@ func ConcatByteSlices(a []byte, b []byte) []byte {
...
+		matches, err := filepath.Match(fileMask, filePath)
+		if err != nil {
+			return nil, err
+		}
+		if matches {
+			result[filePath] = true
+		}
+	}
+	return result, nil
+}
```
g0djan: newline)

utility/utility.go
```diff
@@ -194,3 +194,21 @@ func ConcatByteSlices(a []byte, b []byte) []byte {
...
 }
+
+func SelectMatchingFiles(fileMask string, filePaths map[string]bool) (map[string]bool, error) {
+	if fileMask == "" {
+		return filePaths, nil
+	}
+	fileMask = "/" + fileMask
+	result := make(map[string]bool)
+	for filePath := range filePaths {
+		matches, err := filepath.Match(fileMask, filePath)
```
g0djan: Where have you found Match? I don't see it documentation and it's absent when i try reproduce it with play.golang.

utility/utility_test.go
```diff
@@ -199,3 +199,39 @@ func TestFastCopy_ReturnsError_WhenWriterFails(t *testing.T) {
...
+	files := map[string]bool{
+		"/a":   true,
+		"/b/c": true,
+		"d":    true,
+	}
+	_, err := utility.SelectMatchingFiles("[a-c", files)
+	assert.Error(t, err)
+}
+
+func TestSelectMatchingFiles_ValidMask(t *testing.T) {
```
g0djan: Does it work with */c for example? It's not obvious from SelectMatchingFiles. Add some description to mask format

utility/utility.go
g0djan: Rename it explicitly, pls (:

PR #341, opened 2019-07-23
Parallel tests

munakoiso:


docker/pg_tests/Dockerfile_config_test
```diff
@@ -0,0 +1,22 @@
...
+
+RUN cd main/pg &&     go build -race -o wal-g -tags lzo -ldflags "-s -w -X main.BuildDate=`date -u +%Y.%m.%d_%H:%M:%S`"
+
+FROM wal-g/pg:latest
+
+COPY --from=build /go/src/github.com/wal-g/wal-g/main/pg/wal-g /usr/bin
+
+COPY docker/pg_tests/scripts/ /tmp
+
+CMD su postgres -c "/tmp/tests/config_test.sh"
```
g0djan: All these files are different only in the last line. So it would be better to place a common part to Dockerfile_test. And place "FROM ..." in all Dockerfile_some_test files

Makefile
```diff
@@ -22,14 +22,72 @@ pg_test: install deps pg_build lint unittest unlink_brotli pg_integration_test
...
-	docker-compose build pg_tests
-	docker-compose up --exit-code-from pg_tests pg_tests
-
 pg_integration_test:
 	docker-compose build $(DOCKER_COMMON) pg pg_tests
 	docker-compose up --exit-code-from pg_tests pg_tests
 
+pg_unittests: install deps lint unittest
+
+pg_integration_delete_before_name_find_full_test: install deps pg_build unlink_brotli
```
g0djan: You can create a target with parameters in Makefile. So there won't be necessary to create all these similar targets.

docker-compose.yml
```diff
@@ -68,6 +68,215 @@ services:
     links:
       - s3
 
+  pg_delete_before_name_find_full_test:
```
g0djan: Try to find out: is it possible to create targets with parameters in .yml files?

docker/pg_tests/Dockerfile_several_delta_backups_test
```diff
@@ -0,0 +1,24 @@
...
+COPY main/ main/
+COPY utility/ utility/
+
+RUN cd main/pg &&     go build -race -o wal-g -tags lzo -ldflags "-s -w -X main.BuildDate=`date -u +%Y.%m.%d_%H:%M:%S`"
+
+FROM wal-g/pg:latest
+
+COPY --from=build /go/src/github.com/wal-g/wal-g/main/pg/wal-g /usr/bin
+
+COPY docker/pg_tests/scripts/ /
```
g0djan: How you use this line?

docker/pg_tests/Dockerfile_full_backup_test
```diff
@@ -0,0 +1,22 @@
...
+COPY main/ main/
+COPY utility/ utility/
+
+RUN cd main/pg &&     go build -race -o wal-g -tags lzo -ldflags "-s -w -X main.BuildDate=`date -u +%Y.%m.%d_%H:%M:%S`"
+
+FROM wal-g/pg:latest
+
+COPY --from=build /go/src/github.com/wal-g/wal-g/main/pg/wal-g /usr/bin
+
+COPY docker/pg_tests/scripts/ /tmp
```
g0djan: Let's try to "COPY docker/pg_tests/scripts /tmp" instead of "COPY docker/pg_tests/scripts/ /tmp"

docker/pg_tests/Dockerfile_several_delta_backups_test
munakoiso: in that test "./scripts/drop_pg.sh" is used , but in code I can't find that link
That's why I decide to copy scripts to ./

docker/pg_tests/Dockerfile_full_backup_test
munakoiso: Ok, it works

Makefile
munakoiso: Done

PR #342, opened 2019-07-23
Fix WALG_PGP_KEY variable support.

### LINKED ISSUE #320: WALG_PGP_KEY support is broken
Hello!

wal-g expects ASCII-armored pgp key which is a multiline file. I may be wrong but I do not recall a way to assign a multiline env var without losing line breaks which are crucial for `openpgp/armor` Decode function.

The only option to assign a pgp pub key to the `WALG_PGP_KEY` variable I see is to concatenate a key using `\n`-s and pass it as a single string, but it does not work:

> upload: encryption error openpgp: invalid argument: no armored data found

I have a proof-of-concept solution which looks like that: https://github.com/dragonsmith/wal-g/commit/873132cab91a58d7ce2fd7ed54ef09367d79a309

I've tested the way with `WALG_PGP_KEY_FILE` plus a separate file with a key and it works, but I think it will be easier for somebody (including me) to have an option to use `WALG_PGP_KEY` directly.

What do you think?

Thank you very much!

dragonsmith:
add a support to pass the PGP key as a one line with `\n`-s

Fixes #320


x4m:
Thanks! Pushed

dragonsmith:
@x4m you are welcome!

Please, can you provide some time estimations regarding the next release or pre-release build?

Thank you!

x4m:
@dragonsmith hi!
There's new pre-release

PR #344, opened 2019-07-24
Reduce DB size in integration tests to make CI faster

x4m:


Tinsane:
It seems to me that after such a reduction DB is to small for proper wal-delta testing.

x4m:
ISTM that wal-delta is not correct in terms of segment picking
look
https://travis-ci.org/wal-g/wal-g/jobs/562904103#L5855
why it wants WALs with zero segment, when backup was done in 3rd?

PR #345, opened 2019-07-24
Update Readme for PostgreSQL

ilicmilan:
* Minor fixes in Readme for PostgreSQL.
* Update the default value of `WALG_UPLOAD_CONCURRENCY`.
According to `UploadConcurrencySetting` in [internal/config.go](https://github.com/wal-g/wal-g/blob/master/internal/config.go#L40) default value is 16 instead of specified 10 streams.


x4m:
Thanks! Pushed.

PR #350, opened 2019-07-25
add backup mark command

pohzipohzi:
this adds a utility for users to mark backups as permanent or impermanent


g0djan:
Already merged the same PR.

PR #355, opened 2019-07-26
support MySQL 8.0

mialinx:
MySQL 8 has different fields in some queries,
so we need to read query results with different
columns count

### DIFF
--- a/internal/databases/mysql/mysql.go
+++ b/internal/databases/mysql/mysql.go
@@ -37,6 +37,23 @@ type Backup struct {
 	*internal.Backup
 }
 
+func scanToMap(rows *sql.Rows, dst map[string]interface{}) error {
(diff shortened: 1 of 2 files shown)


internal/databases/mysql/mysql.go
```diff
@@ -37,6 +37,23 @@ type Backup struct {
 	*internal.Backup
 }
 
+func scanToMap(rows *sql.Rows, dst map[string]interface{}) error {
```
x4m: Please, use StructScan http://jmoiron.github.io/sqlx/#advancedScanning

mialinx: Wal-g doesn't depend on sqlx now.  It doesn't make sense to bring a new dependency for just one call.

x4m:
Thanks, pushed

PR #361, opened 2019-07-31
Fix WAL overwrite prevention [WiP]

x4m:


docker/pg_tests/scripts/tests/full_backup_test.sh
```diff
@@ -35,3 +35,16 @@ psql -f tmp/scripts/amcheck.sql -v "ON_ERROR_STOP=1" postgres
...
+# Also we test here WAL overwrite prevention as a part of regular backup functionality
+
+export WALG_PREVENT_WAL_OVERWRITE=true
+
+echo test > /tmp/test_file
+
+wal-g wal-push /tmp/test_file test_file
+wal-g wal-push /tmp/test_file test_file
+echo test1 > /tmp/test_file
+!wal-g wal-push /tmp/test_file test_file
```
g0djan: What is the reason for the exclamation mark? I think that it executes the last command that starts with wal-g. Does it check that rewriting was failed?

x4m: This should negate WAL-G exit code.

g0djan: I checked it. It works on ubuntu if you have space between quotation mark and command.

like there
! wal-g wal-push something

x4m:
Currently WIP.
Test is not testing correcntly

dyson:
I just discovered this issue in testing and I'm super keen to see this merged. Is there anything I can do to help get this in? What else needs to be done? I looks like both the fix and the test should be ok to me?

x4m:
@dyson we need to create test that fails if the feature is not working.
Currently tests pass, that's the problem.

x4m:
Without a test, we will break it quite soon, anyway.

dyson:
Ah okay no problem. I noticed the tests are broken locally after the https://github.com/wal-g/wal-g/pull/341 change so I'll get them working again so this can be tested locally without having to use CI to check.

dyson:
I had a look at this @x4m and there were a few little things here and there to resolve so I put a bunch of commits together is a separate PR: https://github.com/wal-g/wal-g/pull/410

Would be great to get your feedback.

serprex:
@x4m can this PR be closed now that #410 was merged to master?

x4m:
Thanks @serprex !

PR #362, opened 2019-08-02
Support RIPEMD-160

x4m:
Consider supporting https://ru.wikipedia.org/wiki/RIPEMD-160
See https://godoc.org/golang.org/x/crypto/ripemd160
It is somewhat deprecated but in some cases PGP needs it.


x4m:
Users can experience with some keys generated by opengpg, because is picks bad algorithm RIPEMD-160 as default (oldest one). I do not think that currently, we need to support it.

PR #366, opened 2019-08-07
Link statically with liblzo2.a

CyberDem0n:
It adds only 40kb to the binary but allows to get rid from liblzo2 dependency.
In addition to that, execute tests against binary statically linked with liblzo2 and libbrotli.


x4m:
Thanks! Pushed

PR #372, opened 2019-08-10
Set swift settings as env variables for swift client

Tri0L:
We add swift settings to ENV as swift client should read these parameters from environment variables.

Should help with #274


x4m:
Hi! Thanks! Looks useful.
Please adjust tests so that ConfigureFolder tests check that env var are set.
Also, plz add some information about you to contributors list.

x4m:
Thanks, pushed.

PR #375, opened 2019-08-12
Backup mark

munakoiso:


internal/backup_mark.go
```diff
@@ -49,6 +103,101 @@ func GetImpermanentBackupMetadataBefore(baseBackupFolder storage.Folder, backupN
...
+	return backupMetadata, nil
+
+}
+
+func backupHasPermanentInFuture(reverseLinks *map[string][]string, backupName string, permanentBackups *map[string]bool) (bool) {
+	if _, ok := (*permanentBackups)[backupName[len(utility.BackupNamePrefix):len(utility.BackupNamePrefix)+24]]; ok {
+		return true
+	}
+
+	if _, ok := (*reverseLinks)[backupName]; !ok {
```
Tinsane: It appears to be the only needed check. Because if there is permanent backup somewhere in future, then one of next backups is permanent. So you can remove other if's in this function, but leave the comment.

internal/backup_mark.go
```diff
@@ -49,6 +103,101 @@ func GetImpermanentBackupMetadataBefore(baseBackupFolder storage.Folder, backupN
 	return previousImpermanentBackupMetadata, nil
 }
 
+func getMarkedImpermanentBackupMetadata(folder storage.Folder, backupName string) ([]UploadObject, error){
+	backupMetadata := []UploadObject{}
+	baseBackupFolder := folder.GetSubFolder(utility.BaseBackupPath)
```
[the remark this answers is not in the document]
munakoiso: baseBackupFolder?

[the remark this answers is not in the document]
Tinsane: Yep

internal/backup_mark_test.go
```diff
@@ -10,275 +10,263 @@ import (
...
+			},
+		},
 	}
+	toMark := "base_000000010000000000000002"
+	expectUploadObjectLen := 0
+	expectUploadObjectPaths := map[int]string{}
+	testCase(backups, false, true, toMark, expectUploadObjectLen, expectUploadObjectPaths, t)
+}
 
+func testCase(
```
Tinsane: testCase -> testGetBackupMetadataToUpload

internal/backup_mark.go
```diff
@@ -133,23 +132,26 @@ func getMarkedImpermanentBackupMetadata(folder storage.Folder, backupName string
...
 
+//backup has permanent in future only when one of the next backups is permanent
+func backupHasPermanentInFuture(reverseLinks *map[string][]string, backupName string, permanentBackups *map[string]bool) (bool) {
+	//if there is no next backups
 	if _, ok := (*reverseLinks)[backupName]; !ok {
 		return false
 	}
 
+	//if one of the next backups is permanent
 	for _, b := range (*reverseLinks)[backupName]{
```
Tinsane: make fmt!

PR #383, opened 2019-08-15
Backup mark int tests

munakoiso:


Gopkg.lock
```diff
@@ -269,12 +269,12 @@
   version = "v1.8.0"
 
 [[projects]]
+  branch = "master"
```
Tinsane: Don't change it. Just leave it as it is.

docker/pg_tests/scripts/tests/backup_mark_impermanent_test.sh
```diff
@@ -0,0 +1,63 @@
...
+    wal-g --config=${TMP_CONFIG} backup-push ${PGDATA}
+done
+wal-g --config=${TMP_CONFIG} backup-list
+
+# delete all backups
+last_backup_name=`wal-g --config=${TMP_CONFIG} backup-list | tail -n 1 | cut -f 1 -d " "`
+wal-g --config=${TMP_CONFIG} delete before $last_backup_name --confirm
+wal-g --config=${TMP_CONFIG} backup-list
+
+# check that permanent backup still exists
```
Tinsane: Let's fix this comment

Gopkg.lock
Tinsane: I mean. Rollback the change of this file.

PR #390, opened 2019-08-23
Refactoring downloading delta map

g0djan:


internal/delta_map_downloader.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/pkg/errors"
+	"github.com/tinsane/storages/storage"
+	"github.com/wal-g/wal-g/internal/walparser"
+)
+
+func GetDeltaMap(folder storage.Folder, timeline uint32, firstUsedLSN, firstNotUsedLSN uint64) (PagedFileDeltaMap, error) {
+	deltaMap := NewPagedFileDeltaMap()
+	firstUsedDeltaNo, firstNotUsedDeltaNo := GetDeltaRange(firstUsedLSN, firstNotUsedLSN)
+	// Get locations from [firstUsedDeltaNo, lastUsedDeltaNo). We use lastUsedDeltaNo in next step
+	locationsFromDeltas, err := GetLocationsFromDeltas(folder, timeline, firstUsedDeltaNo, firstNotUsedDeltaNo.Previous())
```
Tinsane: Amount of memory used here can exceed RAM size. Delta map should be built online.

g0djan:
@Tinsane just tell when it will be ok to merge. I want to check the comparison of delta maps after review fixes.

g0djan:
It works.

PR #394, opened 2019-08-28
Run xtrabackup in exec command

munakoiso:


cmd/mysql/stream_push.go
```diff
@@ -12,12 +13,14 @@ const StreamPushShortDescription = ""
 
 // streamPushCmd represents the streamPush command
 var streamPushCmd = &cobra.Command{
-	Use:   "stream-push",
+	Use:   "stream-push command\\ [command\\ args]",
```
Tinsane: It seems to be unextendable. Also, it will harden flags addition for "stream-push" command. Let's move inner command to env variable

internal/databases/mysql/stream_push_handler.go
```diff
@@ -1,30 +1,50 @@
...
 
 	"github.com/tinsane/tracelog"
 	"github.com/wal-g/wal-g/internal"
 	"github.com/wal-g/wal-g/utility"
 )
 
-func HandleStreamPush(uploader *Uploader) {
+func HandleStreamPush(uploader *Uploader, command []string) {
+	if len(command) == 0{
+		tracelog.ErrorLogger.Println("WARNING: command not specified")
```
Tinsane: ErrorLogger.Fatal

internal/databases/mysql/stream_push_handler.go
```diff
@@ -1,30 +1,50 @@
...
-	} else {
-		tracelog.ErrorLogger.Println("WARNING: stdin is terminal: operating in test mode!")
-		stream = strings.NewReader("testtesttest")
-	}
 	err = uploader.UploadStream(db, stream)
 	tracelog.ErrorLogger.FatalOnError(err)
+	err = waitFunc()
+	tracelog.ErrorLogger.FatalOnError(err)
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(stderr)
```
Tinsane: No error handling here

internal/databases/mysql/stream_push_handler.go
```diff
@@ -1,30 +1,50 @@
...
+	s := buf.String()
+	if s != ""{
+		tracelog.ErrorLogger.Println("ERROR: Stderr of the command is not empty.")
```
Tinsane: Fatal

internal/databases/mysql/stream_push_handler.go
```diff
@@ -1,30 +1,50 @@
...
+		os.Exit(1)
+	}
+}
+
+func startCommand(command []string) (waitFunc func() error, stdout, stderr io.ReadCloser) {
+	c := exec.Command(command[0], command[1:]...)
+	stdout, err := c.StdoutPipe()
```
Tinsane: It seems to me that it'll create a new "stdout" variable, shadowing existing one (which actually returns in the end). So it's better to avoid using ":=" here or write return statement explicitly.

PR #408, opened 2019-09-13
WALG_DELTA_ORIGIN=LATEST_FULL fix

kruftik:
Fixes wal-g#406


phumpal:
The parent issue persists in v0.2.15. It would be nice to see this reviewed+merged.

x4m:
@kruftik @phumpal Sorry, I've totally forgot about this PR. Let's get it merged.
But there are few problems:
1. It will still error out if we don't have any backups and try to run delta from full backup. I think we should just make a full backup in this case.
2. We do not need to fetch list of backups again. We already have a name of previous backup, it must be there in a storage, unless user messed with the storage.

I propose checking for that previousBackupSentinelDto == nil and fallback to full backup in this case.

phumpal:
Awesome, thanks for the quick response @x4m 

> previousBackupSentinelDto == nil and fallback to full backup in this case.

:+1: for usability; this seems like a good approach.

>  must be there in a storage, unless user messed with the storage

Gentle ping @kruftik otherwise I'm happy try in a new PR.

kruftik:
It was not an easy task to catch up with the issue itself and PR after almost an year has passed :) 


@x4m,

> It will still error out if we don't have any backups and try to run delta from full backup. I think we should just make a full backup in this case.

I can't figure out, how such an event can happen, given the code:

```go
		previousBackupName, err = getLatestBackupName(folder)
		if err != nil {
			if _, ok := err.(NoBackupsFoundError); ok {
				tracelog.InfoLogger.Println("Couldn't find previous backup. Doing full backup.")
			} else {
			}
```

> I propose checking for that previousBackupSentinelDto == nil and fallback to full backup in this case.

I believe you meant checking for nil of previousBackupSentinelDto.IncrementFullName since previousBackupSentinelDto can't be nil at all

> We do not need to fetch list of backups again. We already have a name of previous backup, it must be there in a storage, unless user messed with the storage.

reworked the PR a bit, now previousBackupName used as the delta source if previous nil-check was true

x4m:
Looks good to me. @kruftik , thanks!

PR #413, opened 2019-09-18
Check that pg_version and pg_control are exist

munakoiso:


internal/backup_push_handler.go
```diff
@@ -108,6 +109,10 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
 	backupName, backupStartLSN, pgVersion, dataDir, err := bundle.StartBackup(conn,
 		utility.CeilTimeUpToMicroseconds(time.Now()).String())
 	meta.DataDir = dataDir
+	if dataDir != archiveDirectory {
+		warning := fmt.Sprintf("Data directory '%s' is not equal to archive directory '%s'", dataDir, archiveDirectory)
```
x4m: despite the name, this is not archive directory, but rather wal-push argument

PR #414, opened 2019-09-19
Fix stream-push

munakoiso:


internal/databases/mongo/stream_push_handler.go
```diff
@@ -3,17 +3,26 @@ package mongo
...
+	waitFunc, stream, errorStream := internal.StartCommand(command)
 	uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(utility.BaseBackupPath)
 	err := uploader.UploadStream(stream)
 	tracelog.ErrorLogger.FatalOnError(err)
+	var errorString string
+	if b, err := ioutil.ReadAll(errorStream); err == nil {
+		errorString = string(b)
+	}
 	err = waitFunc()
+	if err == nil {
```
reshke: Here everything is ok, but it’s better to do like:

if err != nil {
   // handle error
}

// do the rest

internal/databases/mysql/stream_push_handler.go
```diff
@@ -3,21 +3,30 @@ package mysql
...
 	tracelog.ErrorLogger.FatalOnError(err)
 	defer utility.LoggedClose(db, "")
 	err = uploader.UploadStream(db, stream)
 	tracelog.ErrorLogger.FatalOnError(err)
+	var errorString string
+	if b, err := ioutil.ReadAll(errorStream); err == nil {
+		errorString = string(b)
+	}
 	err = waitFunc()
+	if err == nil {
```
reshke: same as above

internal/databases/redis/stream_push_handler.go
```diff
@@ -5,18 +5,26 @@ import (
...
 	backupName := "dump_" + time.Now().Format(time.RFC3339)
 	err := uploader.UploadStream(backupName, stream)
 	tracelog.ErrorLogger.FatalOnError(err)
+	var errorString string
+	if errorBytes, err := ioutil.ReadAll(errorStream); err == nil {
+		errorString = string(errorBytes)
+	}
 	err = waitFunc()
-	tracelog.ErrorLogger.FatalOnError(err)
+	if err == nil {
```
reshke: !=

internal/databases/mongo/stream_push_handler.go
```diff
@@ -3,18 +3,27 @@ package mongo
...
 )
 
 func HandleStreamPush(uploader *Uploader, command []string) {
-	waitFunc, stream:= internal.StartCommand(command)
+	waitFunc, stream, errorStream := internal.StartCommand(command)
 	uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(utility.BaseBackupPath)
 	err := uploader.UploadStream(stream)
 	tracelog.ErrorLogger.FatalOnError(err)
+	var errorString string
+	if errorBytes, err := ioutil.ReadAll(errorStream); err == nil {
```
reshke: What if err != nil ? We need to handle all errs explicitly

PR #416, opened 2019-09-22
Set compiled config to ENV

Tri0L:
Main problem that wal-g can't work with config file only. It still needed some ENV vars for swift/pg/etc.

This change set compiled config to ENV variables to have transparent config for all libs in wal-g.
And for now wal-g can work even only config file available, without ENV.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -3,7 +3,9 @@ package internal
 import (
 	"github.com/spf13/viper"
 	"github.com/tinsane/tracelog"
+	"os"
 	"os/user"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


x4m:
This looks very useful, thanks!

Maybe just a few lines of documentation would improve overall impact.
Also, plz consider case when user mistyped some env vars, we should inform him that options from config were not interpreted correctly

x4m:
Wow, this looks like a lot of work on collection every single property! Thanks!

x4m:
I've made some corrections, can you please revisit my cahnges?

Tri0L:
Yep! Looks good from my side.

x4m:
pushed, thanks!

Issue #409, opened 2019-09-14
wal-g backup-push hangs while pushing to Swift storage


skvav:
Caught this problem second time. Again it happened right after pg_stop_backup() according to log.
backup-push runs on production server in archive mode with wal-g wal-push archive command. backup-push starts right after pg_repack finished its work. Maybe it is a cause of this freeze? When I run backup-push manually it usually works successfully.

```INFO: 2019/09/14 03:58:55.640727 Starting part 124 ...
INFO: 2019/09/14 03:58:55.742759 Finished writing part 124.
INFO: 2019/09/14 03:58:55.750918 Starting part 125 ...
INFO: 2019/09/14 03:58:55.750957 /global/pg_control
INFO: 2019/09/14 03:58:55.790923 Finished writing part 125.
INFO: 2019/09/14 03:58:55.804163 Calling pg_stop_backup()
SIGABRT: abort
PC=0x968d21 m=0 sigcode=0

goroutine 0 [idle]:
runtime.futex(0x22c36c0, 0x80, 0x0, 0x0, 0x0, 0xc001ab2800, 0x0, 0x0, 0x7fff38a58200, 0x919c72, ...)
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/sys_linux_amd64.s:531 +0x21
runtime.futexsleep(0x22c36c0, 0xc000000000, 0xffffffffffffffff)
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/os_linux.go:46 +0x4b
runtime.notesleep(0x22c36c0)
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/lock_futex.go:151 +0xa2
runtime.stopm()
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/proc.go:2016 +0xe3
runtime.findrunnable(0xc000054f00, 0x0)
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/proc.go:2487 +0x4dc
runtime.schedule()
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/proc.go:2613 +0x13a
runtime.park_m(0xc000385380)
	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/runtime/proc.go:2676 +0xae
runtime.mcall(0x13aa5d0)
[...]

PR #417, opened 2019-09-23
Fix background uploader race condition

x4m:


x4m:
I think this is a root cause of #409

PR #423, opened 2019-09-26
Readme for mongo

munakoiso:


MongoDB.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+wal-g stream-fetch LATEST | mongorestore --archive --oplogReplay
+```
+
+* ``stream-push``
+
+Command for compressing, encrypting and sending backup from stream to storage.
+
+```
```
g0djan: I guess ther should be an example with WALG_STREAM_CREATE_COMMAND.

PR #427, opened 2019-09-30
[WIP]Little refactoring: try to remove dependency on postgres in configure process 

reshke:


internal/config.go
```diff
@@ -28,10 +28,6 @@ const (
 	PgpKeyPathSetting            = "WALG_PGP_KEY_PATH"
 	PgpKeyPassphraseSetting      = "WALG_PGP_KEY_PASSPHRASE"
 	PgDataSetting                = "PGDATA" // TODO : do something with it
-	UserSetting                  = "USER"   // TODO : do something with it
```
reshke: because we don't use them

Issue #171, opened 2019-01-25
"Invalid page header" warnings during backup-push


vstax:
Hello, I'm trying out wal-g on barely loaded database and sometimes getting warnings like this (here `wal-g backup-push /var/lib/pgsql/11/data` was started at 2:02 AM:
```
Path:  
INFO: 2019/01/25 02:02:01.649327 Delta backup from base_000000010000005900000010_D_00000001000000590000000C with LSN 5910000028. 
INFO: 2019/01/25 02:02:01.780709 Walking ...
INFO: 2019/01/25 02:02:01.780992 Starting part 1 ...
WARNING: 2019/01/25 02:02:27.105408 Invalid page  0  page header  &{0 0 30859 0 64 8176 8176 8196}
WARNING: 2019/01/25 02:02:27.105722 failed to read file '/base/27057/880656' as incremented
WARNING: 2019/01/25 02:02:27.109879 Invalid page  0  page header  &{0 0 30859 0 64 8176 8176 8196}
WARNING: 2019/01/25 02:02:27.109894 failed to read file '/base/27057/880660' as incremented
INFO: 2019/01/25 02:02:27.480395 Finished writing part 1.
INFO: 2019/01/25 02:02:27.864803 Starting part 2 ...
INFO: 2019/01/25 02:02:27.864846 /global/pg_control
INFO: 2019/01/25 02:02:27.869397 Finished writing part 2.
INFO: 2019/01/25 02:02:28.901469 Starting part 3 ...
INFO: 2019/01/25 02:02:28.901515 backup_label
INFO: 2019/01/25 02:02:28.901523 tablespace_map
INFO: 2019/01/25 02:02:28.904414 Finished writing part 3.
INFO: 2019/01/25 02:02:29.078008 Uploaded 3 compressed tar Files.
```

Wal-g 0.2.4, postgresql 11.1 running on EL 7.6, settings are
```
WALG_UPLOAD_CONCURRENCY=4
WALG_DELTA_MAX_STEPS=6
WALG_COMPRESSION_METHOD=brotli
```
Backup is done from local connection (PGHOST=/var/run/postgresql). Nothing unusual in postgresql.conf.

postgresql log file shows that 3 WAL files were pushed at the same time:
```
Path:  
INFO: 2019/01/25 02:02:03.088336 FILE PATH: 000000010000005900000033.br
Path:  
INFO: 2019/01/25 02:02:28.821981 FILE PATH: 000000010000005900000034.00000028.backup.br
[...]

PR #428, opened 2019-09-30
issue 171 stop printing Invalid page header warnings

AlexSvalov99:
#171


internal/incremental_page_reader.go
```diff
@@ -156,7 +155,6 @@ func (pageReader *IncrementalPageReader) SelectNewValidPage(pageBytes []byte, bl
 			isNew = true
 			valid = true
 		} else {
-			tracelog.WarningLogger.Println("Invalid page ", blockNo, " page header ", pageHeader)
```
g0djan: I think a better solution is managing this message with a debug level.

PR #429, opened 2019-10-01
Аdd db size to userData

munakoiso:
- fix race in tarBall.size
- some refactoring in stream push
- add dataSize and dataTarSize to userData
- fix flap (wait while pg not ready)

### DIFF
--- a/internal/backup_push_handler.go
+++ b/internal/backup_push_handler.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync/atomic"
 	"time"
 
 	"github.com/spf13/viper"
@@ -132,6 +133,7 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
 	tracelog.ErrorLogger.FatalOnError(err)
 	err = bundle.FinishQueue()
 	tracelog.ErrorLogger.FatalOnError(err)
+	dataSize := bundle.TarBall.Size()
 	err = bundle.UploadPgControl(uploader.Compressor.FileExtension())
 	tracelog.ErrorLogger.FatalOnError(err)
 	// Stops backup and write/upload postgres `backup_label` and `tablespace_map` Files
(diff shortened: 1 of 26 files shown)


internal/databases/mongo/stream_push_handler.go
```diff
@@ -3,29 +3,17 @@ package mongo
...
 	"io"
-	"io/ioutil"
 
 	"github.com/tinsane/tracelog"
 	"github.com/wal-g/wal-g/utility"
 )
 
 func HandleStreamPush(uploader *Uploader, command []string) {
-	waitFunc, stream, errorStream := internal.StartCommand(command)
+	waitAndFatalOnError, stream := internal.StartCommand(command)
```
reshke: nice

internal/start_command.go
```diff
@@ -15,5 +16,18 @@ func StartCommand(command []string) (waitFunc func() error, stdout, stderr io.Re
 	err = c.Start()
 	waitFuncResult := c.Wait
 	tracelog.ErrorLogger.FatalOnError(err)
-	return waitFuncResult, stdoutResult, stderrResult
+	return func() {
```
reshke: here we lose code readability. I think this function is too big to declare it and call it in one place of the code

internal/uploader.go
```diff
@@ -22,6 +22,7 @@ type Uploader struct {
 	deltaFileManager     *DeltaFileManager
 	archiveStatusManager DataFolder
 	Failed               atomic.Value
+	tarSize              *int64
```
reshke: is there any reason why we use a pointer, is it possible to simply pass by value?

internal/start_command.go
munakoiso: I don't know other ways to declare this function with variables from StartCommand

internal/uploader.go
munakoiso: It's not possible because of races.

internal/backup_push_handler.go
```diff
@@ -171,7 +173,12 @@ func HandleBackupPush(uploader *Uploader, archiveDirectory string, isPermanent b
 
 	currentBackupSentinelDto.setFiles(bundle.GetFiles())
 	currentBackupSentinelDto.BackupFinishLSN = &finishLsn
-	currentBackupSentinelDto.UserData = GetSentinelUserData()
+	userData := map[string]interface{}{"backupSize": dataSize, "backupTarSize": atomic.LoadInt64(uploader.tarSize)}
```
x4m: Please create typed field in dto classes

internal/storage_tar_ball.go
```diff
@@ -15,7 +16,7 @@ import (
 type StorageTarBall struct {
 	backupName  string
 	partNumber  int
-	size        int64
+	size        *int64
```
x4m: technically, this is not size of tarball. This is a size of all tarballs.
I'm not sure this is a good place to store size of all tarballs, but let's leave it aside, just rename it correctly.

PR #436, opened 2019-10-04
Consistent formatting for commands in README

tbartelmess:


x4m:
Cool! Thanks! We really need to improve the documentation.
I'd send a beer for every PR to docs, but Russian Post refuses to ship alcohol.

PR #437, opened 2019-10-04
Move wal-g build into mongo docker image

perekalov:


tests_func/images/base/Dockerfile
```diff
@@ -18,16 +18,25 @@ RUN cp --force --recursive --verbose \
...
+    netcat-openbsd iputils-ping \
+    net-tools dnsutils \
+    nginx nginx-full nginx-common \
+	&& rm -rf /var/lib/apt/lists/*
+
+ENV GOLANG_URL "https://golang.org/dl/go1.13.1.linux-amd64.tar.gz"
+RUN wget -O go.tgz "https://golang.org/dl/go1.13.1.linux-amd64.tar.gz" && \
+	tar -C /usr/local -xzf go.tgz && \
+	rm go.tgz && \
+	export PATH="/usr/local/go/bin:$PATH" && \
```
reshke: ok, you can also try to use image that already has golang in future
like here
https://github.com/wal-g/wal-g/blob/master/docker/pg_tests/Dockerfile_prefix#L1
or here
https://github.com/wal-g/wal-g/blob/master/docker/golang/Dockerfile#L1
or maybe it is essential to use such a version?

PR #443, opened 2019-10-15
Add unit tests to utility.TryFetchTimeRFC3999; minor refactor;

GSKryachko:


GSKryachko:
I decided do create another PR with my commit because after master was force-pushed back it all messed up.
https://github.com/wal-g/wal-g/pull/447

PR #447, opened 2019-10-17
Add unit tests to utility.TryFetchTimeRFC3999; minor refactor;

GSKryachko:
Attempt number 2
I closed previous PR because master was force pushed back after I've cloned it

### DIFF
--- a/utility/utility_test.go
+++ b/utility/utility_test.go
@@ -200,6 +200,44 @@ func TestFastCopy_ReturnsError_WhenWriterFails(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestTryFetchTimeRFC3999_Valid(t *testing.T) {
+	var testCases = []struct {
+		input    string
+		expected string
+	}{
+		{"20191015T211200Z", "20191015T211200Z"},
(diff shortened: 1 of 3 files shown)


utility/utility_test.go
```diff
@@ -200,6 +200,44 @@ func TestFastCopy_ReturnsError_WhenWriterFails(t *testing.T) {
...
+func TestTryFetchTimeRFC3999_Valid(t *testing.T) {
+	var testCases = []struct {
+		input    string
+		expected string
+	}{
+		{"20191015T211200Z", "20191015T211200Z"},
+		{"20191015T211200Z22221015T211200Z", "20191015T211200Z"},
+		{"         20191015T211200Z", "20191015T211200Z"},
+		{"000000020191015T211200Z", "20191015T211200Z"},
+		{"20191015T211200ZZZZZ", "20191015T211200Z"},
```
g0djan: Hmm, would it work with 201920191015T211200Z ? I consider it as invalid string

GSKryachko: It will be treated as a valid string and return time 20191015T211200Z. You think we need to change that behavior? That may break compatibility.

PR #451, opened 2019-10-20
Get backup finish time from metadata.json instead of using sentinel m…

GSKryachko:
…odification time.

### DIFF
--- a/internal/delete_handler.go
+++ b/internal/delete_handler.go
@@ -81,8 +81,8 @@ func getBackupsAndGarbage(folder storage.Folder) (backups []BackupTime, garbage 
 
 // TODO : unit tests
 func getBackupTimeSlices(backups []storage.Object, folder storage.Folder) []BackupTime {
-	sortTimes := make([]BackupTime, len(backups))
-	for i, object := range backups {
+	sortTimes := make([]BackupTime, 0)
+	for _, object := range backups {
 		key := object.GetName()
 		if !strings.HasSuffix(key, utility.SentinelSuffix) {
 			continue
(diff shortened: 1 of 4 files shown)


testtools/util.go
```diff
@@ -45,11 +47,17 @@ func NewStoringMockUploader(storage *memory.Storage, deltaDataFolder internal.Da
...
+	meta000 := fmt.Sprintf("{\"finish_time\":\"%s\"}", time.Now().Format(time.RFC3339))
 	subFolder := folder.GetSubFolder(utility.BaseBackupPath)
 	subFolder.PutObject("base_123_backup_stop_sentinel.json", &bytes.Buffer{})
+	subFolder.PutObject("base_123/metadata.json", strings.NewReader(meta123))
 	subFolder.PutObject("base_456_backup_stop_sentinel.json", strings.NewReader("{}"))
+	subFolder.PutObject("base_456/metadata.json", strings.NewReader(meta456))
 	subFolder.PutObject("base_000_backup_stop_sentinel.json", &bytes.Buffer{}) // last put
-	subFolder.PutObject("base_123312", &bytes.Buffer{})                        // not a sentinel
+	subFolder.PutObject("base_000/metadata.json", strings.NewReader(meta000))
+	subFolder.PutObject("base_123312", &bytes.Buffer{}) // not a sentinel
```
g0djan: You should test backup-list to check the difference with the old behavior.

internal/delete_handler.go
```diff
@@ -73,26 +73,38 @@ func getBackupsAndGarbage(folder storage.Folder) (backups []BackupTime, garbage
...
 		}
-		time := object.GetLastModified()
-		sortTimes[i] = BackupTime{utility.StripBackupName(key), time, utility.StripWalFileName(key)}
+		name := utility.StripBackupName(key)
+		backup, err := GetBackupByName(name, folder)
+		if err != nil {
+			continue
+		}
+		var finishTime time.Time
+		meta, err := backup.FetchMeta()
```
g0djan: Some users have huge amount of backups. So this type of listing could create high-load reading for storage.
We think it would be better to use this type of listing only with explicit flag --details.

testtools/util.go
GSKryachko: What exactly do I test? There may be a valid difference. The new listing order may be different from the old one if, for example, for some reason sentinel creation took a long time and it's modification time is significantly different from the backup finish time written in metadata.

g0djan: You should check that the time that you get from getBackupTimeSlices should be equal to values in metafiles and different from the time of last modified in storage. Also time in metafiles has better precision than a time of last-modified. So you need to update the integration test of backup-list to check it.

PR #457, opened 2019-10-24
Added possibility to dump compiled ENV vars

Tri0L:
Dump all ENV vars of wal-g runtime when DEVEL log level is set.
Can be useful in case of configuration problems in config file/env.


g0djan:
Yeah, you're right

PR #458, opened 2019-10-25
Split stream fetch command

reshke:


internal/databases/mysql/stream_fetch_handler.go
```diff
@@ -2,6 +2,7 @@ package mysql
```
g0djan: Why this file still has name "stream_fetch_handler"?

PR #461, opened 2019-11-01
Essential functionality for working with wal-g in godog feature tests

Aallzz:
Changes:

1. We don't generate docker-compose file anymore, environment variables are used instead. 
2. We stop using python behave and now go godog is used. That's why functions for work with mongo and minio

To run tests:
```
cd tests_func/
godog 
```

The next steps: 

- Add feature tests from genbackup and new functions for testing mongo if needed.


Aallzz:
In [e5c78ea](https://github.com/wal-g/wal-g/pull/461/commits/e5c78eade944b362b9dfc3d38b89991919f91087)  few things were added:

- In python tests minio is ignored. Now all backups are stored in minio01 container.
- No backups are not empty, but store all dbs. For dumping and restoring mongodump and mongorestore are used.
- Update in feature tests: restoring backup to mongodb and comparing the data in two databases

PR #465, opened 2019-11-07
Add --apply option

reshke:


internal/config.go
```diff
@@ -39,6 +39,8 @@ const (
 	PgSslModeSetting             = "PGSSLMODE"
 	TotalBgUploadedLimit         = "TOTAL_BG_UPLOADED_LIMIT"
 	NameStreamCreateCmd          = "WALG_STREAM_CREATE_COMMAND"
+	NameStreamRestoreCmd         = "WALG_STREAM_RESTORE_COMMAND"
+	NameLogApplyCmdPath          = "WALG_MYSQL_BINLOG_APPLY_COMMAND_PATH"
```
mialinx: May be WALG_LOG_APPLY_COMMAND is better name ?  It's more db neutral and may be suitable for mongodb (oplog)

internal/databases/mysql/binlog_fetch_handler.go
```diff
@@ -42,7 +43,7 @@ func (handlers BinlogFetchHandlers) DownloadLogTo(logFolder storage.Folder, logN
 		tracelog.ErrorLogger.Print(err)
 		return err
 	}
-	return nil
+	return handlers.applier(dstLogFilePath)
```
mialinx: We do not need to save binlog to local file at all.
Instead we can stream binlog content directly to apply command via pipe.

For instance we could use following WALG_LOG_APPLY_COMMAND:

mysqlbinog - | mysql 

Command should be executed as shell command ($SHELL env variable) here.

internal/databases/mysql/binlog_fetch_handler.go
```diff
@@ -1,23 +1,24 @@
...
 type BinlogFetchSettings struct {
-	dt time.Time
+	startTs   time.Time
+	endTS     *time.Time
+	needApply bool
 }
 
-func (settings BinlogFetchSettings) GetEndTS() (*time.Time, error) {
-	return &settings.dt, nil
+func (settings BinlogFetchSettings) GetLogsStartEndTs() (time.Time, *time.Time) {
```
mialinx: EndTS is much clearer

internal/start_command.go
```diff
@@ -31,3 +31,35 @@ func StartCommand(command []string) (waitFunc func(), stdout io.ReadCloser) {
 		}
 	}, stdoutResult
 }
+
+func ApplyCommand(path string, args []string, stdin io.ReadCloser) error {
```
mialinx: stdin should be io.Reader,  as we don't call Close method

internal/start_command.go
```diff
@@ -31,3 +31,35 @@ func StartCommand(command []string) (waitFunc func(), stdout io.ReadCloser) {
...
+		tracelog.ErrorLogger.Println(errorString)
+		return err
+	}
+	err = c.Start()
+	tracelog.ErrorLogger.FatalOnError(err)
+	if err != nil {
+		return err
+	}
+	tracelog.InfoLogger.Println("waiting on command")
+	if err := c.Wait(); err != nil {
```
mialinx: We should read from stderrResult, otherwise Command may be blocked forever
Consider using https://golang.org/pkg/os/exec/#Cmd.CombinedOutput

internal/start_command.go
```diff
@@ -31,3 +31,35 @@ func StartCommand(command []string) (waitFunc func(), stdout io.ReadCloser) {
...
+func ApplyCommand(path string, args []string, stdin io.ReadCloser) error {
+	c := exec.Command(path, args...)
+	if stdin != nil {
+		c.Stdin = stdin
+	}
+	stderrResult, err := c.StderrPipe()
+	if err != nil {
+		tracelog.ErrorLogger.PrintError(err)
+		var errorString string
+		if errorBytes, err := ioutil.ReadAll(stderrResult); err == nil {
```
mialinx: We failed to create stderrResult pipe here. There is no sense reading it.

PR #472, opened 2019-11-18
fix prefetch lzo

munakoiso:


internal/configure.go
```diff
@@ -194,17 +194,12 @@ func ConfigureArchiveStatusManager() (DataFolder, error) {
 // ConfigureUploader connects to storage and creates an uploader. It makes sure
 // that a valid session has started; if invalid, returns AWS error
 // and `<nil>` values.
-func ConfigureUploader() (uploader *Uploader, err error) {
+func ConfigureUploader(configurateCompressionMethod bool) (uploader *Uploader, err error) {
```
x4m: This configuration API looks bad. I think we can cope with it for now, but lets at least mark it with TODO.
I think that TODO could describe that in one case Uploader is used not like uploader.

PR #482, opened 2019-11-30
Adding unit-tests

devyatka:


testtools/mock_logging.go
```diff
@@ -0,0 +1,43 @@
...
+	infoStats := callingPrintFuncStats{
+		PrintLnCallsCount: 0,
+		PrintMsg: "",
+	}
+	errorStats := callingFatalOnErrorFuncStats{
+		FatalOnErrorCallsCount: 0,
+		Err: nil,
+	}
+
+	return infoLoggerMock{ Stats: &infoStats },
```
munakoiso: May be it's better to return Logging struct (infoLogger, errorLogger)?

internal/backup_list_handler.go
```diff
@@ -13,17 +13,49 @@ import (
...
+	}
+	logging := Logging{
+		InfoLogger: tracelog.InfoLogger,
+		ErrorLogger: tracelog.ErrorLogger,
+	}
+
+	HandleBackupList(getBackupsFunc, writeBackupListFunc, logging)
+}
+
 // TODO : unit tests
```
munakoiso: Delete TODO?

testtools/mock_logging.go
devyatka: Well, inside a test we can't get infoLogger and errorLogger as mock structs (with usefull Stats fields for tests) because they are interfaces and have only Println and FatalOnError methods, so I think it wouldn't be better...

testtools/mock_logging.go
```diff
@@ -0,0 +1,43 @@
...
+type callingPrintFuncStats struct {
+	PrintLnCallsCount int
+	PrintMsg string
+}
+
+type infoLoggerMock struct {
+	Stats *callingPrintFuncStats
+}
+
+func (loggerMock infoLoggerMock) Println(v ...interface{}) {
```
reshke: https://github.com/golang/mock

devyatka: Why can't it be a matter of opinion? :)

PR #486, opened 2019-12-03
Add unit tests

OptimumDev:
Added unit tests for `IsDirectoryEmpty` function in `internal/backup.go`
___

### Tests

 - `true` when:
   - empty
   - doesn't exist (looks strange, I would treat this like error, but program uses this logic)
- `false` when:
   - one file
   - several files
   - nested directory


internal/backup_test.go
```diff
@@ -134,4 +134,4 @@ func TestFetchSentinelReturnErrorWhenSentinelUnmarshallable(t *testing.T) {
 
 	assert.Error(t, err)
 	assert.Equal(t, errorMessage, err.Error()[:len(errorMessage)])
-}
+}
```
reshke: it is unnecessary

internal/backup_private_test.go
```diff
@@ -0,0 +1,106 @@
...
+	assert.False(t, actual)
+}
+
+func TestIsDirectoryEmpty_ReturnsTrue_WhenDirectoryDoesntExist(t *testing.T) {
+	dir, err := createTempDir("not_existing")
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = os.Remove(dir)
+	if err != nil {
```
reshke: assert.NoError(t, err)

internal/backup_private_test.go
```diff
@@ -0,0 +1,106 @@
...
+func TestIsDirectoryEmpty_ReturnsFalse_WhenNestedDirectoryIsInDirectory(t *testing.T) {
+	dir, err := createTempDir("not_empty")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(dir)
+
+	nested, err := ioutil.TempDir(dir, "nested")
+	if err != nil {
+		t.Fatal(err)
```
reshke: assert.NoError(t, err)

internal/backup_private_test.go
```diff
@@ -0,0 +1,106 @@
...
+	dir, err := createTempDir("not_empty")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(dir)
+
+	for i := 0; i < 3; i++ {
+		file, err := ioutil.TempFile(dir, "file")
+		if err != nil {
+			t.Fatal(err)
```
reshke: assert.NoError(t, err)

PR #487, opened 2019-12-04
apply public to private refactoring;

romutchio:
Fixes in packages interfaces. Huge amount of methods do export outside.


romutchio:
Fixes in packages interfaces. Huge amount of methods do export outside.

PR #496, opened 2019-12-10
Support for deletion satisfying two retention policies

Aallzz:
We want to be able to delete backups satisfying one or two criteria at the same time: 
1) Retain given number of backups 
2) Retain commits created after given backup or time

In the base version API looks like `wal-g delete retain <retain_count> --after <backup_name|timestamp>`


internal/delete_handler.go
```diff
@@ -154,6 +155,84 @@ func FindTargetRetain(folder storage.Folder,
...
+
+	choiceFuncRetain := getRetainChoiceFunc(retentionCount, modifier, isFullBackup)
+	if choiceFuncRetain == nil {
+		return nil, utility.NewForbiddenActionError("Not allowed modifier for 'delete retain'")
+	}
+	choiceFuncAfter := func(object storage.Object) bool {
+		timeCheck := timeLine.Before(object.GetLastModified()) || timeLine.Equal(object.GetLastModified())
+		if modifier == NoDeleteModifier {
+			return timeCheck
+		} else {
```
reshke: you can make this easier(remove else):

if cond {
     return val1
}
return val2

PR #497, opened 2019-12-10
Added restart support for mongodb oplog-push command

perekalov:
- Added start archiving point discovery to oplog-push
- Added initial oplog doc validator
- Fixed archiving timer reset
- Added oplog module


internal/databases/mongo/oplog/oplog.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+const (
+	TimestampDelimiter = "."
+	TimestampRegexp    = `[0-9]+\` + TimestampDelimiter + `[0-9]+`
```
x4m: Well, go way is just place the dot into literal :) But I do not think this is an issue.

PR #509, opened 2019-12-18
Mongo load

Aallzz:
Tool for creating load on mongo for testing purposes. It gets commands from json patron (which can be generated by patron generator with config) and executes them with given level of concurrency

### DIFF
--- a/tests_func/load/load_test.go
+++ b/tests_func/load/load_test.go
@@ -1,0 +1,72 @@
+package mongoload
+
+import (
+	"context"
+	"fmt"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"os"
+	"testing"
+	"time"
+)
+
+func prepareTestLoad(configFile, patronFile, mongoUri string, ctx context.Context) (*mongo.Client, *os.File, error) {
+	err := generatePatronsFromFile(configFile)
+	if err != nil {
+		return nil, nil, err
+	}
+	cli, err := mongo.NewClient(options.Client().ApplyURI(mongoUri))
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot create client to mongodb: %v", err)
+	}
+	err = cli.Connect(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot connect to mongodb: %v", err)
+	}
+	err = cli.Ping(ctx, nil)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot ping mongodb basse: %v", err)
+	}
+	f, err := os.Open(patronFile)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot read patron: %v", err)
+	}
+	return cli, f, nil
+
+}
+
+func TestLoadRange(t *testing.T) {
+	ctx, cancel:= context.WithTimeout(context.Background(), 250 * time.Millisecond)
+	defer cancel()
+
+	cli, f, err := prepareTestLoad("mongo_load_config.json",
+		"mongo_load_patron_1.json",
+		"mongodb://localhost:27017", ctx)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	var errcList []<-chan error
+	roc, errc0, err := ReadRawMongoOps(ctx, f, 3)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	defer f.Close()
+	errcList = append(errcList, errc0)
+	cmdc, errc1 := MakeMongoOps(ctx, cli, roc)
+	errcList = append(errcList, errc1)
+	rsc := RunMongoOpFuncs(ctx, cmdc, 3, 3)
+	PrintMongoOpRes(ctx, rsc)
+	err = WaitForPipeline(errcList...)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	time.Sleep(1 * time.Second)
+	err = os.Remove("mongo_load_patron_1.json")
+	if err != nil {
+		t.Error(err)
+		return
+	}
+}
--- a/tests_func/load/mongo_load.go
+++ b/tests_func/load/mongo_load.go
@@ -1,0 +1,264 @@
+package mongoload
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"io"
+	"sync"
+)
+
+type RawMongoOp = map[string]json.RawMessage
+
+func ReadRawMongoOps(ctx context.Context, reader io.Reader, mxCmdsSize int) (<-chan RawMongoOp, <-chan error, error) {
+	cmds := make(chan RawMongoOp, mxCmdsSize)
+	errc := make(chan error, 1)
+	dec := json.NewDecoder(reader)
+
+	t, err := dec.Token()
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot parse patron json: %v", err)
+	}
+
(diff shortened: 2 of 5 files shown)


tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func main() {
+
+	err := generatePatronsFromFile("./tests_func/load/mongo_load_config.json")
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	cli, err := mongo.NewClient(options.Client().ApplyURI("mongodb://localhost:27017"))
```
perekalov: we need custom uri support here

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+	for cnt > 0 {
+		res = append(res, value)
+		cnt--
+	}
+	return res
+}
+
+func main() {
+
+	err := generatePatronsFromFile("./tests_func/load/mongo_load_config.json")
```
perekalov: so we're developing 2 different tools here: "bullet generator" and "load generator"
they should be separated (different files and binaries), let's move it to different files
after that we need to make convenient interface between them:
-   "bullet generator" writes bullets to stdout or to file
-   "load generator" reads bullets from stdin or from file

tests_func/load/mongo_load.go
perekalov: we need custom uri support here
consider using 'cobra' lib, e.g. https://github.com/wal-g/wal-g/blob/master/cmd/mongo

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+		return
+	}
+
+	dat, err := ioutil.ReadFile("./tests_func/load/mongo_load_patron_1.json")
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	mw := CreateMongoWorker(3)
```
perekalov: Workers count should be passed as argument, (see comment about 'cobra' above)

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+	err = mw.AddMongoOp(cli, string(dat))
```
perekalov: Our test documents can have a very large size (up to 16mb each), so we can not hold all "bullets" in memory.
I see an option: 

1. read bullets one by one from JSON file using stream interface (https://golang.org/pkg/encoding/json/#Decoder.Decode)
2. apply them one by one to worker's pool via buffered channel

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+
+func main() {
+
+	err := generatePatronsFromFile("./tests_func/load/mongo_load_config.json")
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	cli, err := mongo.NewClient(options.Client().ApplyURI("mongodb://localhost:27017"))
+	if err != nil {
```
perekalov: can we move these statements into a dedicated function (say "connectDB")?

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+		return err
+	}
+	for _, ropdata := range arr {
+		var opdata map[string]json.RawMessage
+		err = json.Unmarshal(ropdata, &opdata)
+		if err != nil {
+			return fmt.Errorf("command expected to be a map, but found %s", testUtils.GetStringType(opdata))
+		}
+		switch string(opdata["op"]) {
+		case `"c"`:
```
perekalov: Are ""-quotes necessary here?

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, err
+	}
+	return &res, nil
+}
+
+func mongoRunCommandOp(client *mongo.Client, op *MongoRunCommandOp) {
+	db := client.Database(op.DbName)
+	var result bson.M
+	err := db.RunCommand(contextByName[op.Ctx], op.Doc).Decode(&result)
+	if op.MakeLogs {
```
perekalov: Do we need this flag? We can log anything using "DebugLogger" and set logging level during initialization

tests_func/load/mongo_load.go
Aallzz: This is an example of usage now, I think I'll remove it when it is ready to merge with master, because there will be usage of this somewhere else.

tests_func/load/mongo_load.go
Aallzz: This is an example, we are going to pass the client from outside the lib

tests_func/load/mongo_load.go
Aallzz: Fixed

tests_func/load/mongo_load.go
Aallzz: Yes, because in patrons command has "op": "c" and because of using RawMessage we need the case like this

tests_func/load/mongo_load.go
Aallzz: Fixed, now all logs are gone in error channel or result channel

tests_func/load/mongo_load.go
```diff
@@ -0,0 +1,275 @@
...
+		})
+		if err != nil {
+			return fmt.Sprintf("Failed execution of transaction with id %d: %+v", op.Id, err)
+		} else {
+			return fmt.Sprintf("Successful execution of transaction with id %d%s", op.Id, tres)
+		}
+	}
+}
+
+func MongoRunFuncs(ctx context.Context, fc <-chan func(ctx context.Context) string, resSize int) (<-chan string, <-chan struct{}) {
```
perekalov: Is it possible to rewrite as mentioned in "Bounded parallelism" https://blog.golang.org/pipelines ?

tests_func/load/load_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"os"
+	"testing"
+	"time"
+)
+
+func prepareTestLoad(configFile, patronFile, mongoUri string) (*mongo.Client, context.Context, *os.File, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		time.Sleep(800 * time.Millisecond)
```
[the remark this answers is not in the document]
Aallzz: That's better, done

tests_func/load/load_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	cmdc, errc1 := MongoOpsFromRawChannel(ctx, cli, roc)
+	rsc, done := MongoRunFuncs(ctx, cmdc, 3)
+Loop:
+	for {
+		select {
+		case e, ok := <-errc0:
+			if ok {
+				fmt.Println("E: ", e)
+				break Loop
```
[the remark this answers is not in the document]
Aallzz: Fixed

tests_func/load/load_test.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+		case c, ok := <-rsc:
+			if ok {
+				fmt.Println("C: ", c)
+			}
+		case <-done:
+			fmt.Println("D")
+			break Loop
+		default:
```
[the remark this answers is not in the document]
Aallzz: Otherwise it's waiting, I think, but it's no longer in use

tests_func/load/mongo_load.go
Aallzz: Done

tests_func/load/load_test.go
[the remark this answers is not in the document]
Aallzz: Unfortunately go vet is complaining if cancel function is not used, that's why I came back with context.WithCancel and sleep in goroutine approach

PR #510, opened 2019-12-18
Add mongodb oplog initial replay procedure

perekalov:


cmd/mongo/oplog_replay.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/internal/databases/mongo"
+	"github.com/wal-g/wal-g/internal/databases/mongo/oplog"
+
+	"github.com/spf13/cobra"
+	"github.com/wal-g/tracelog"
+)
+
+func SetupSignalHandler(ctx context.Context, cancel func()) func() {
```
x4m: Is there any analogue to Closeables in Java or Disposable in .Net?
Returning cleanup function is not a widely used idiom.
Here you have state of closing func as local variable ch

internal/databases/mongo/oplog/applier.go
```diff
@@ -0,0 +1,67 @@
...
+type DBApplier struct {
+	uri string
+	mc  *internal.MongoClient
+}
+
+func NewDBApplier(uri string) *DBApplier {
+	return &DBApplier{uri: uri}
+}
+
+func (dba *DBApplier) Apply(ctx context.Context, ch chan Record, wg *sync.WaitGroup) (chan struct{}, chan error, error) {
```
x4m: These number of arguments need names

internal/databases/mongo/oplog/applier.go
```diff
@@ -0,0 +1,67 @@
...
+		defer mc.Close(ctx)
+		defer close(errc)
+		defer close(done)
+		for {
+			select {
+			case op, ok := <-ch:
+				tracelog.DebugLogger.Printf("Applyer receieved op %s (%s on %s), open: %t", op.TS, op.OP, op.NS, ok)
+
+				if !ok {
+					done <- struct{}{}
```
x4m: This will stuck until someone read from chan. Is it intended behavior?
Maybe you want to use buffered chan?

perekalov: Yes it's "work is done, no error acquired" indicator.

x4m: But you are doing it in a loop for bunch of works?

x4m: Oh, I see, it's not a loop, actually

PR #518, opened 2020-01-01
Fix incorrect wal-g binlog fetching

reshke:
Suppose s3 was unavailable from t1 to t2. If we try to push some binlogs while s3 is unavalable and then perform PITR on t3 such that t1 < t3 < t2,we will fail, because of incorrect condition in func logFileShouldBeFetched (let's not pay attention to end Timestamp - parse binlog first TS from binlog header is enough)

### DIFF
--- a/docker/mysql_tests/scripts/tests/binlog_push_while_s3_down_test.sh
+++ b/docker/mysql_tests/scripts/tests/binlog_push_while_s3_down_test.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+set -e -x
+
+. /usr/local/export_common.sh
+export WALE_S3_PREFIX=s3://mysqlbinlogpushwhiles3downbucket
+export WALG_COMPRESSION_METHOD=lz4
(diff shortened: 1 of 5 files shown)


docker/mysql_tests/scripts/tests/binlog_push_while_s3_down_test.sh
```diff
@@ -0,0 +1,48 @@
...
+
+mkdir "${MYSQLDATA}"
+wal-g backup-fetch LATEST | xbstream -x -C "${MYSQLDATA}"
+
+
+# start mysql && try to apply binlogs
+chown -R mysql:mysql "${MYSQLDATA}"
+service mysql start && wal-g binlog-fetch --since LATEST --until "${WALG_MYSQL_BINLOG_END_TS}" --apply
+
+check_test_additional_data
```
reshke: fails on current master

mialinx:
not relevant

PR #467, opened 2019-11-08
Add unit tests for utility.TrimFileExtension

Denchick:

PR #519, opened 2020-01-08
Fix problem with dot directory wal-g/issues/258

Denchick:


Denchick:
I'll resend.

Denchick:
Its a bit strange that in this PR there are commits 74bd9ac and 5394a25. They was merged in https://github.com/wal-g/wal-g/pull/467.

So, i just want send fix to wal-g/issues/258.
It could be cool to write tests for

g0djan:
Actually we can test it completely in integration tests, but it will be too long or not clear, so just make a test that you described.

osterboss86:
Fr

PR #522, opened 2020-01-09
Add unit tests to DecryptAndDecompress

GSKryachko:


osterboss86:
Dope

internal/extract.go
```diff
@@ -82,10 +82,9 @@ func extractOne(tarInterpreter TarInterpreter, source io.Reader) error {
 	return nil
 }
 
-// TODO : unit tests
 // Ensures that file extension is valid. Any subsequent behavior
 // depends on file type.
-func decryptAndDecompressTar(writer io.Writer, readerMaker ReaderMaker, crypter crypto.Crypter) error {
```
g0djan: Instead of extracting this method, change extract_test package to internal package.
It's "go way" to test in the same package

internal/extract_test.go
```diff
@@ -60,17 +65,152 @@ func TestTar(t *testing.T) {
...
+
+	decompressed := &bytes.Buffer{}
+	err = internal.DecryptAndDecompressTar(decompressed, brm, nil)
+
+	if err != nil {
+		t.Logf("%+v\n", err)
+	}
+
+	assert.Error(t, err)
+	assert.EqualError(t, err, "DecryptAndDecompressTar: lz4 decompress failed. Is archive encrypted?: DecompressLz4: lz4 write failed: invalid lz4 data")
```
g0djan: Instead of check message error in tests, it would be better to check the error type. And you can create new if you need it.

g0djan: And fix it in other tests too

internal/extract.go
GSKryachko: This file imports testtools, and testtools import internal package. If I change the file package to internal, there will be a cyclic import. I am not sure that we could/should address that issue here, but instead it should be a part of a big task of changing all tests to go way.

internal/extract.go
```diff
@@ -111,7 +117,11 @@ func decryptAndDecompressTar(writer io.Writer, readerMaker ReaderMaker, crypter
 			continue
 		}
 		err = decompressor.Decompress(writer, readCloser)
-		return errors.Wrapf(err, "DecryptAndDecompressTar: %v decompress failed. Is archive encrypted?", decompressor.FileExtension())
+		if err == nil {
+			return nil
+		}
+		decompressionError := newDecompressionError(err)
+		return errors.Wrapf(decompressionError, "DecryptAndDecompressTar: %v decompress failed. Is archive encrypted?", decompressor.FileExtension())
```
g0djan: Too long line

PR #533, opened 2020-01-15
Fix mongodb func tests

perekalov:


tests_func/Dockerfile
```diff
@@ -18,19 +19,22 @@ RUN apt-get update -qq && \
...
 COPY wal-g ${WALG_REPO}/
 WORKDIR ${WALG_REPO}
-RUN make install && \
-    make deps && \
-    make mongo_build && \
-    cp main/mongo/wal-g /usr/bin/wal-g
\ No newline at end of file
+
+RUN make unlink_brotli && make unlink_libsodium && make link_brotli && make link_libsodium
+RUN make mongo_build && \
```
g0djan: So, after adding these tests to run full time of running takes almost 50 minutes. If it will exceed 50 minutes tests will start flap. We should think how to resolve it.

PR #550, opened 2020-02-02
Add unit tests for parseTs function

GSKryachko:


internal/stream_fetch_helper.go
```diff
@@ -27,7 +27,6 @@ type LogFetchHandlers interface {
 	AfterFetch(logs []storage.Object) error
 }
 
-// TODO : unit tests
 func ParseTS(endTSEnvVar string) (endTS *time.Time, err error) {
 	endTSStr, ok := GetSetting(endTSEnvVar)
 	if ok {
```
g0djan: If there is no such environment variable, the function will return nil instead of an error.

GSKryachko: Yes, and I test it in TestParseTs_shouldReturnNilOnNoTime. Is something wrong with that?

PR #564, opened 2020-02-15
Fix leaking abstraction in uploader

GSKryachko:


internal/wal_uploader.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/wal-g/storages/storage"
+	"github.com/wal-g/wal-g/internal/compression"
+	"github.com/wal-g/wal-g/utility"
+	"io"
+	"path"
+)
+
+// Uploader contains fields associated with uploading tarballs.
+// Multiple tarballs can share one uploader.
+type WalUploader struct {
```
g0djan: Just uploader is "wal uploader" too. DeltaFileManger is needed for use Wal-Delta. So it should be WalDeltaUploader.

I think you should have parent class or interface Uploader and 2 implementations WalUploader which doesn't use Wal-delta and WalDeltaUploader which use.

In the best case Uploader should be created once on initialization of wal-g  where it's possible  to choose implementation of Uploader in runtime depending on environment variable WALG_USE_WAL_DELTA(find it in config) and

internal/configure.go
```diff
@@ -203,14 +203,34 @@ func ConfigureUploader() (uploader *Uploader, err error) {
...
-	uploader = NewUploader(compressor, folder, deltaFileManager)
+	uploader = NewUploader(compressor, folder)
+	return uploader, err
+}
+
+// ConfigureWalUploader connects to storage and creates an uploader. It makes sure
+// that a valid session has started; if invalid, returns AWS error
+// and `<nil>` values.
+func ConfigureWalUploader() (uploader *WalUploader, err error) {
+	uploader, err = ConfigureMockWalUploaderWithoutCompressMethod()
```
g0djan: ConfigureMockWalUploader looks like a function for testing something with mock, but you use it in here

internal/wal_uploader.go
```diff
@@ -0,0 +1,64 @@
...
+			defer utility.LoggedClose(recordingReader, "")
+		}
+	} else {
+		walFileReader = file
+	}
+
+	return uploader.UploadFile(newNamedReaderImpl(walFileReader, file.Name()))
+}
+
+func (uploader *WalUploader) FlushFiles() {
```
g0djan: uploader -> walUploader

internal/wal_uploader.go
```diff
@@ -0,0 +1,64 @@
...
+	uploader.DeltaFileManager.FlushFiles(uploader.Uploader.clone())
```
g0djan: Hmm, still looks strange enough but there no loops and it's good. Actually, why is there necessary to clone the Uploader?

internal/configure.go
```diff
@@ -220,6 +240,16 @@ func ConfigureMockUploaderWithoutCompressMethod() (uploader *Uploader, err error
 		return nil, errors.Wrap(err, "failed to configure folder")
 	}
 
+	uploader = NewUploader(nil, folder)
+	return uploader, err
+}
+
+func ConfigureMockWalUploaderWithoutCompressMethod() (uploader *WalUploader, err error) {
```
g0djan: Is this a function only for testing?

internal/databases/mysql/mysql.go
```diff
@@ -27,7 +27,7 @@ const (
 )
 
 type Uploader struct {
-	*internal.Uploader
+	*internal.WalUploader
```
g0djan: Mysql doesn't have WAL

internal/wal_uploader.go
```diff
@@ -0,0 +1,64 @@
...
+	uploader := NewUploader(compressor, uploadingLocation)
+
+	return &WalUploader{
+		uploader,
+		deltaFileManager,
+	}
+}
+
+// Clone creates similar WalUploader with new WaitGroup
+func (uploader *WalUploader) clone() *WalUploader {
```
g0djan: I can't find any usage of this function. Delete if this function unnecessary. If it is using somewhere I should take a better look at this function, because it could be really dangerous to copy deltafilemanager

cmd/mysql/backup_push.go
```diff
@@ -14,10 +14,10 @@ var streamPushCmd = &cobra.Command{
 	Use:   "backup-push",
 	Short: StreamPushShortDescription,
 	Run: func(cmd *cobra.Command, args []string) {
-		uploader, err := internal.ConfigureUploader()
+		uploader, err := internal.ConfigureWalUploader()
 		tracelog.ErrorLogger.FatalOnError(err)
 		command := internal.GetStreamCreateCmd()
-		mysql.HandleBackupPush(&mysql.Uploader{Uploader: uploader}, command)
+		mysql.HandleBackupPush(&mysql.Uploader{WalUploader: uploader}, command)
```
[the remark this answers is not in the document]
GSKryachko: I know, but it __does__ use `UploadWalFile` method, so it needs WalUploader. Check method `tryArchiveBinLog` in `mysql/binlog_push_handler`.  I can only suggest renaming WalUploader and related methods to HistoryUploader or something.

internal/configure.go
GSKryachko: This is a typo, thanks for spotting.

internal/wal_uploader.go
GSKryachko: It was this way in `internal/wal_push_handler.go`, where I moved it from, so I do not know.

internal/configure.go
GSKryachko: No, it should not have 'Mock' in name

internal/databases/mysql/mysql.go
GSKryachko: But it needs WalUploader, see other comment.

internal/wal_uploader.go
GSKryachko: It is used in bgUploader. It does not copy deltafilemanager, just returns the same one.

cmd/mysql/backup_push.go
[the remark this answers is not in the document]
g0djan: Renaming will be okay

internal/wal_uploader.go
g0djan: Try to pass it without clone, and check tests

PR #575, opened 2020-02-20
Added optional lzo condition

usernamedt:
Make installing lzo optional, disabled by default


usernamedt:
> Looks pretty good, but you also should add USE_LZO for building release in .travis.yml
Done

Also, fixed an issue with excess double quotes appeared in build tags along with a small refactoring of the Makefile - now it is possible to set USE_LZO for any target database build.

PR #594, opened 2020-03-18
add storages 2

devyatka:


usernamedt:
Done in https://github.com/wal-g/wal-g/pull/1044

PR #596, opened 2020-03-19
Implementation of copying backups between clouds

Denchick:
Hello! I want to receive feedback by the implementation of the feature that @patsevanton proposed.

Interface details.

1. I added the `copy` command that accepts the flags `--backup-name`, `--from`, `--to` and `--without-flags`.
2. If no `backup-name` is specified, then copy everything
3. If the paths to the `from` or` to` configs are not specified, then an error.
4. By default, copying with history is assumed.
5. It is assumed that the use of the `without-history` flag without the` backup-name` will be prohibited.

Now to the implementation.

0. Everything i did works only for Postgres (at least for now).
1. I can copy all files from one storage to another with the whole history.
2. I can copy a specific backup(it seems) with history (but not quite right, i think).
3. So far, only synchronously

My problems.

I could not understand how to determine which exact history i should copy. How to understand which files belong to the backup and which not). I tried to explore how other commands work (mark, delete, push, fetch, etc) but it didn't help me.


g0djan:
1. Do you mean without-history? Not without-flags.
2. Let's copy everything if there are flag `--everything`, and throw an error if not backup-name specified and `--everything` was not used.
3. OK
4. OK
5. OK

internal/backup.go
```diff
@@ -322,3 +323,19 @@ func shouldUnwrapTar(tarName string, sentinelDto BackupSentinelDto, filesToUnwra
...
 	return false
 }
+
+func getLastWalFilename(backup *Backup) (string, error) {
+	meta, err := backup.fetchMeta()
+	if err != nil {
+		tracelog.DebugLogger.FatalError(err)
+		return "", err
+	}
+	timelineID64, err := strconv.ParseUint(backup.Name[len(utility.BackupNamePrefix):len(utility.BackupNamePrefix)+8], 0x10, sizeofInt32bits)
```
g0djan, line 333: Extract to variable value that you use more than one time. len(utility.BackupNamePrefix)

g0djan, line 333: 0x10 - don't use magic constants. Put it in const variable with a nice name

internal/config.go
```diff
@@ -174,22 +175,32 @@ func isAllowedSetting(setting string, AllowedSettings map[string]bool) (exists b
...
+// GetSettingFrom extract setting by key from config if key is set, return empty string otherwise
+func GetSettingFrom(config *viper.Viper, key string) (value string, ok bool) {
+	if config.IsSet(key) {
+		return config.GetString(key), true
 	}
 	return "", false
 }
 
 func getWaleCompatibleSetting(key string) (value string, exists bool) {
+	return getWaleCompatibleSettingFrom(key, viper.GetViper())
```
g0djan, line 192: GetSettingFrom has the same arguments but in another order. Use the same order in both places.

internal/copy_handler.go
```diff
@@ -0,0 +1,141 @@
...
+// HandleCopy copy specific or all backups from one storage to another
+func HandleCopy(fromConfigFile string, toConfigFile string, backupName string, withoutHistory bool) {
+	var from, fromError = ConfigureFolderFromConfig(fromConfigFile)
+	var to, toError = ConfigureFolderFromConfig(toConfigFile)
+	if fromError != nil || toError != nil {
+		return
+	}
+	var infos, err = getObjectsToCopy(backupName, from, to, withoutHistory)
+	if err != nil {
+		tracelog.ErrorLogger.FatalError(err)
```
g0djan, line 27: Go way: handle error on the top level.
It's top-level and if it was Fatal you should exit. Use FatalOnError here

internal/copy_handler.go
```diff
@@ -0,0 +1,141 @@
...
+	for i := 0; i < len(infos); i += maxParallelJobsCount {
+		var jobsToRun = utility.Min(maxParallelJobsCount, len(infos)-(i+1))
+		for j := 0; j < jobsToRun; j++ {
+			var info = infos[i+j]
+			go copyObject(info, jobs)
+		}
+		for j := 0; j < jobsToRun; j++ {
+			var isSuccess = <-jobs
+			if !isSuccess {
+				tracelog.DebugLogger.Fatal("Something went wrong.")
```
g0djan, line 135: Go way: handle error on the top level. It's not. Update message also)

internal/copy_handler.go
```diff
@@ -0,0 +1,141 @@
...
+	var jobs = make(chan bool, maxParallelJobsCount)
+
+	for i := 0; i < len(infos); i += maxParallelJobsCount {
+		var jobsToRun = utility.Min(maxParallelJobsCount, len(infos)-(i+1))
+		for j := 0; j < jobsToRun; j++ {
+			var info = infos[i+j]
+			go copyObject(info, jobs)
+		}
+		for j := 0; j < jobsToRun; j++ {
+			var isSuccess = <-jobs
```
g0djan, line 133: Instead of using channels, you can use waitGroup https://gobyexample.com/waitgroups

internal/config.go
Denchick: unexpectedly GetSettingFrom method was not needed

Denchick:
1. Yes
2. Are you sure about this naming? it seems to me that `everything` more general word than `history`. I'm just afraid of misunderstanding by users.

internal/copy_handler.go
```diff
@@ -55,87 +101,49 @@ func getObjectsToCopy(backupName string, from storage.Folder, to storage.Folder,
...
 	if err != nil {
 		return nil, err
 	}
 	var older = func(object storage.Object) bool { return lastWalFilename <= object.GetName() }
-	return buildCopyingInfos(fromWalFolder, to, objects, older), nil
+	return BuildCopyingInfos(fromWalFolder, to, objects, older), nil
 }
 
-func copyAll(from storage.Folder, to storage.Folder) ([]copyingInfo, error) {
+func GetAllObjects(from storage.Folder, to storage.Folder) ([]CopyingInfo, error) {
```
g0djan, line 133: -> GetAll**Objects**
-> return type  []**CopyingInfo**
GetAllCopyingInfo?

internal/copy_handler.go
```diff
@@ -55,87 +101,49 @@ func getObjectsToCopy(backupName string, from storage.Folder, to storage.Folder,
...
 	if err != nil {
 		return nil, err
 	}
 	var hasBackupPrefix = func(object storage.Object) bool { return strings.HasPrefix(object.GetName(), backupPrefix) }
-	return buildCopyingInfos(from, to, objects, hasBackupPrefix), nil
+	return BuildCopyingInfos(from, to, objects, hasBackupPrefix), nil
 }
 
-func copyHistory(backup *Backup, from storage.Folder, to storage.Folder) ([]copyingInfo, error) {
+func GetHistoryObjects(backup *Backup, from storage.Folder, to storage.Folder) ([]CopyingInfo, error) {
```
g0djan, line 115: -> GetHistoryObjects
-> return type []CopyingInfo
GetHistoryCopyingInfo?

internal/copy_handler.go
```diff
@@ -55,87 +101,49 @@ func getObjectsToCopy(backupName string, from storage.Folder, to storage.Folder,
...
 	tracelog.InfoLogger.Print("Collecting history files... ")
 	var fromWalFolder = from.GetSubFolder(utility.WalPath)
+	tracelog.InfoLogger.Print("getSubFolder succeess!")
 	var lastWalFilename, err = getLastWalFilename(backup)
 	if err != nil {
-		return nil, err
+		return make([]CopyingInfo, 0), nil
```
g0djan, line 121: WTF? Diff shows me that previously you return an error here. But now you just skip it

internal/copy_handler.go
```diff
@@ -55,87 +101,49 @@ func getObjectsToCopy(backupName string, from storage.Folder, to storage.Folder,
 	return infos, nil
 }
 
-func copyBackup(backup *Backup, from storage.Folder, to storage.Folder) ([]copyingInfo, error) {
+func GetBackupObjects(backup *Backup, from storage.Folder, to storage.Folder) ([]CopyingInfo, error) {
```
g0djan, line 104: -> GetBackupObjects
-> return type []CopyingInfo
GetBackupCopyingInfo?

internal/copy_handler.go
```diff
@@ -22,31 +23,76 @@ func HandleCopy(fromConfigFile string, toConfigFile string, backupName string, w
...
+	tracelog.ErrorLogger.FatalOnError(err)
+	isSuccess, err := StartCopy(infos)
+	tracelog.ErrorLogger.FatalOnError(err)
+	if isSuccess {
+		tracelog.InfoLogger.Println("Success.")
+	}
+}
+
+func StartCopy(infos []CopyingInfo) (bool, error) {
+	var maxParallelJobsCount = 8 // TODO place for improvement
```
g0djan, line 36: I' m almost sure that Golang has something like std::hardware_concurrency() in C++

internal/config.go
```diff
@@ -307,6 +315,46 @@ func CheckAllowedSettings(config *viper.Viper) {
...
 }
 
+func bindToEnv(k string, val string) error {
+	if err := os.Setenv(strings.ToUpper(k), val); err != nil {
+		return errors.Wrap(err, "Failed to bind config to env variable")
+	}
+	return nil
+}
+
+func AssertRequiredSettingsSet() error {
```
g0djan: Does it mean that cobra and viper doesn't check it?

docker/pg_tests/scripts/tests/copy_all_test.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "Copying all backups (when there are only 1 backup) failed"
+    exit 2
+fi
+
+# push a few more backups
+for i in 2 3
+do
+    pgbench -i -s 1 postgres &
+    sleep 1
+    wal-g --config=${TMP_CONFIG} backup-push ${PGDATA}
```
g0djan, line 54: There are you push only base backups. Increase WALG_MAX_DELTA_STEPS and you'll test when folder also contain delta backups.

docker/pg_tests/scripts/tests/copy_all_test.sh
```diff
@@ -0,0 +1,72 @@
...
+done
+wal-g --config=${TMP_CONFIG} backup-list
+
+# copy all again
+wal-g copy --from=${TMP_CONFIG} --to=${TO_TMP_CONFIG}
+last_backup_name=`wal-g --config=${TMP_CONFIG} backup-list | tail -n 1 | cut -f 1 -d " "`
```
g0djan, line 60: Instead of comparison only last files redirect output to file and check with diff that sorted lists are the same.

docker/pg_tests/scripts/tests/copy_backup_test.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# push a few more backups
+for i in 2 3
+do
+    pgbench -i -s 1 postgres &
+    sleep 1
+    wal-g --config=${TMP_CONFIG} backup-push ${PGDATA}
+done
+
+backup_name=`wal-g --config=${TMP_CONFIG} backup-list | tail -n 1 | cut -f 1 -d " "`
```
g0djan, line 57: You don't use it in next lines

docker/pg_tests/scripts/tests/copy_backup_test.sh
```diff
@@ -0,0 +1,79 @@
...
+    pgbench -i -s 1 postgres &
+    sleep 1
+    wal-g --config=${TMP_CONFIG} backup-push ${PGDATA}
+done
+
+# copy that backup
+wal-g copy --backup-name=${last_backup_name} --from=${TMP_CONFIG} --to=${TO_TMP_CONFIG}
+
+# check if backup copied
+if [ $last_backup_name != $copied_last_backup_name ];
```
g0djan, line 71: You don't declare last_backup_name

docker/pg_tests/scripts/tests/copy_all_test.sh
Denchick: Good idea, thank you

docker/pg_tests/scripts/tests/copy_backup_test.sh
Denchick: Fixed test's scenario

internal/copy_handler.go
Denchick: Yeap.

internal/copy_handler.go
Denchick: Yes. That's ok when there are no wal files (maybe yet). That just means that i have nothing to copy.

internal/copy_handler.go
Denchick: The problem is too many opened files, so std::hardware_concurrency() will not help

internal/config.go
Denchick: I think yes. I just decomposed the old code into several methods, this is one of them

internal/copy_handler.go
Denchick: I don't know which number i should choose and i didn't find the answer in google :)

internal/config.go
g0djan: Hmm

patsevanton:
Thanks! Where instruction? readme?

Denchick:
I forgot :) At least, you can use `wal-g copy --help`. I will add instructions but i don't know when.

NikolayS:
@Denchick thanks for the feature. What about documentation (except `--help`)? ;-) Without it, people don't know about this feature.

internal/copy_handler.go
NikolayS: @Denchick 

>  i'd found out that 8 is optimal on my local machine.

WAL-G has good settings to control parallelism -- I wonder, why those aren't used here? 
Hard-coded `8` may be too big or too small depending on concrete case.

Denchick:
@NikolayS, that's definitely a good idea 😃 I added docs to [PostgreSQL.md](https://github.com/wal-g/wal-g/blob/master/PostgreSQL.md).

PR #598, opened 2020-03-22
Configuration using command line flags

Leoltron:


internal/config.go
```diff
@@ -267,16 +297,44 @@ func InitConfig() {
...
-	for k, v := range viper.AllSettings() {
-		val, ok := v.(string)
-		if ok {
-			if err := os.Setenv(strings.ToUpper(k), val); err != nil {
-				tracelog.ErrorLogger.Println("failed to bind config to env variable", err.Error())
-				os.Exit(1)
-			}
+func bindToEnv(k string, val string) {
+	if err := os.Setenv(strings.ToUpper(k), val); err != nil {
+		tracelog.ErrorLogger.Println("failed to bind config to env variable", err.Error())
```
g0djan, line 304: Go way: handle error on the top level.
Return error here.

internal/config.go
```diff
@@ -267,16 +297,44 @@ func InitConfig() {
...
+		os.Exit(1)
```
g0djan, line 305: And you can do it with one line.
tracelog.ErrorLogger.FatalOnError()
It err == nil than nothing happened, but if there is an error it will print logline and exit.

internal/config.go
```diff
@@ -267,16 +297,44 @@ func InitConfig() {
...
+	}
+}
+
+func AssertRequiredSettingsSet() {
+	if !isAnyStorageSet() {
+		tracelog.ErrorLogger.Println("Failed to find any configured storage")
+		os.Exit(1)
```
g0djan, line 312: Same thing about FatalOnError and returning error

internal/config.go
```diff
@@ -267,16 +297,44 @@ func InitConfig() {
...
+	}
+
+	for setting, required := range RequiredSettings {
+		isSet := viper.IsSet(setting)
+
+		if !isSet && required {
+			message := "Required variable " + setting + " is not set. You can set is using --" + toFlagName(setting) +
+				" flag or variable " + setting
+			tracelog.ErrorLogger.Println(message)
+			os.Exit(1)
```
g0djan, line 322: Same thing about FatalOnError and returning error

cmd/mysql/binlog_fetch.go
```diff
@@ -25,6 +25,11 @@ var binlogFetchCmd = &cobra.Command{
 		tracelog.ErrorLogger.FatalOnError(err)
 		mysql.HandleBinlogFetch(folder, fetchBackupName, fetchUntilDt)
 	},
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
Leoltron: * init() is called for every command of the module before user command is processed. The only settings that could be set in init() function are settings that are common for all commands.
* PersistentPreRun is called when command is already known to us and about to be executed. This is the correct place to set all required settings for this particular command.

PR #616, opened 2020-03-30
Add uuid filtration to mongodb restore procedure

perekalov:


ArnobKumarSaha:
Hi @perekalov @VGoshev , 
Can you please explain why we don't preserve the UUID value in oplog-replay ? 
Is there any specific reason ? 

[I know, its a long time ago. But your answer will be very helpful to me.]

PR #629, opened 2020-04-13
Use newer pip for fixing build

defntvdm:
I don't know why, but newer pip uses whl, so we don't need to install python3-wheel, python3-dev, gcc, etc. to build wheel from source


defntvdm:
I hope soon be fixed in over PRs

perekalov:
Thanks!

PR #644, opened 2020-05-08
Reimplement BgUploader and fix tests

bunsenmcdubbs:
Individual commits are relatively self-contained and reviewable.

This PR
1. Fixes and extends test coverage for BgUploader. Previously the tests were completely broken and didn't run at all.
2. Re-implements a large portion of BgUploader, specifically targeting concurrency controls. My goal was to simplify the logic and it more easily understood.

### DIFF
--- a/internal/archive_status_manager_fake.go
+++ b/internal/archive_status_manager_fake.go
@@ -1,0 +1,42 @@
+package internal
+
+import "sync"
+
+type FakeASM struct {
+	uploaded map[string]bool
+	mutex    sync.Mutex
+}
+
+var _ ArchiveStatusManager = &FakeASM{}
+
+func NewFakeASM() *FakeASM {
+	return &FakeASM{
+		uploaded: make(map[string]bool),
+	}
+}
+
+func (asm *FakeASM) isWalAlreadyUploaded(walFilePath string) bool {
+	asm.mutex.Lock()
+	defer asm.mutex.Unlock()
+	isUploaded, ok := asm.uploaded[walFilePath]
+	return ok && isUploaded
+}
+
+// IsWalAlreadyUploaded is used for testing
+func (asm *FakeASM) IsWalAlreadyUploaded(walFilePath string) bool {
+	return asm.isWalAlreadyUploaded(walFilePath)
+}
+
+func (asm *FakeASM) markWalUploaded(walFilePath string) error {
+	asm.mutex.Lock()
+	defer asm.mutex.Unlock()
+	asm.uploaded[walFilePath] = true
+	return nil
+}
+
+func (asm *FakeASM) unmarkWalFile(walFilePath string) error {
+	asm.mutex.Lock()
+	defer asm.mutex.Unlock()
+	asm.uploaded[walFilePath] = false
+	return nil
+}
--- a/internal/bguploader.go
+++ b/internal/bguploader.go
@@ -1,15 +1,16 @@
 package internal
 
 import (
+	"context"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/spf13/viper"
 	"github.com/wal-g/tracelog"
+	"golang.org/x/sync/semaphore"
 )
 
 // BgUploader represents the state of concurrent WAL upload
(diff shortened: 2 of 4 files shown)


x4m:
Looks like a really big chunk of work, thanks! Review will take some time, but I hope it won't be long.
BTW CI was fixed few minutes ago, I hope red X's on commit numbers will be gone :)

internal/bguploader.go
```diff
@@ -1,146 +1,171 @@
...
-	atomic.StoreInt32(&bgUploader.maxParallelWorkers, 0) // stop new jobs
-	bgUploader.mutex.Unlock()
+	go b.scanFiles()
+}
 
-	bgUploader.running.Wait() // wait again for those who jumped to the closing door
+// Stop pipeline. Stop can be safely called concurrently and repeatedly.
+func (b *BgUploader) Stop() {
+	// Send signal to stop scanning for and uploading new files
+	b.cancelFunc()
```
jschaf, line 77: Maybe document why it's safe to call cancelFunc multiple times (https://golang.org/pkg/context/#CancelFunc):

>  A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing.

internal/bguploader.go
```diff
@@ -1,146 +1,171 @@
...
 
-	mutex sync.Mutex
+	// ctx signals internals to keep/stop enqueueing more uploads
+	ctx context.Context
+	// cancelFunc signals internals to stop enqueuing more uploads
+	cancelFunc context.CancelFunc
 
-	preventWalOverwrite bool
+	// workerCount tracks number of concurrent uploaders. Limited to maxParallelWorkers
+	workerCount *semaphore.Weighted
```
jschaf, line 32: Count suffix usually implies an `int`.  Maybe `workerCountSem`?

internal/bguploader.go
```diff
@@ -1,146 +1,171 @@
...
+
+		for _, f := range files {
+			select {
+			case <-b.ctx.Done():
+				return
+			case fileChan <- f:
+			}
+		}
+
+		// Sleep 5 seconds before scanning filesystem again. Exit if
```
jschaf, line 105: Can you expand why sleep for 5 seconds? Is it to avoid busy-loop scanning the file-system? It's got a bit of a magic constant feel.

internal/bguploader.go
```diff
@@ -1,146 +1,171 @@
...
+	// ctx signals internals to keep/stop enqueueing more uploads
+	ctx context.Context
+	// cancelFunc signals internals to stop enqueuing more uploads
+	cancelFunc context.CancelFunc
 
-	preventWalOverwrite bool
+	// workerCount tracks number of concurrent uploaders. Limited to maxParallelWorkers
+	workerCount *semaphore.Weighted
+	// maxParallelWorkers usually defined by WALG_DOWNLOAD_CONCURRENCY
+	maxParallelWorkers int32
```
jschaf, line 34: Explain what maxParallelWorkers does?  Looks like it's the number of concurrent goroutines to upload WAL files.

internal/bguploader.go
```diff
@@ -1,146 +1,171 @@
...
+
+	// numUploaded is tracked to control amount of work done in one cycle of archive_command
+	numUploaded int32
+	// maxNumUploaded controls the amount of work done in one cycle of archive_command
+	maxNumUploaded int32
```
jschaf, line 39: Can you document the interaction between `maxNumUploaded` and `maxParallelWorkers`?  On brief inspection it looks like `maxNumUpload` is set by `TotalBgUploadedLimit` which defaults to 32.  What happens when `maxParallelWorkers` exceeds that value? Do workers race for work and some lose?

internal/archive_status_manager_fake.go
```diff
@@ -0,0 +1,42 @@
+package internal
```
bunsenmcdubbs, line 1: Oh wait, I just noticed that I can do `NewDataFolderASM(&MockDataFolder)` instead of this `FakeASM`. I'll look into implementing that instead (possibly just by converting fake to use that as an implementation) - "MockDataFolder-backed DataFolderASM"

bunsenmcdubbs: Upon second thought, this seems lightweight enough that its harmless to leave as-is. Happy to implement if someone feels strongly.

bunsenmcdubbs:
@x4m I am interested in testing this in our staging environment. Does this project work with cross compilation (especially with (native?) brotli compilation)

x4m:
Currently, by default brotli build id disabled. You can enable it manually or we can just get it merged and create tesing pre-release.

internal/bguploader.go
x4m: I agree. Also waiting for 5 seconds seems way too long. Tiplically WALs are uplodad 4-16 files per second.

bunsenmcdubbs: I'll reduce this wait to... 1 second?

In production, we've noticed that when wal-g falls behind and we accumulate 100GB's-1TB's of WAL, `ls pg_wal > /dev/null` takes up to one second. I wanted to add a conservative wait time in order to avoid scanning the filesystem more than strictly necessary. 

Since all of BgUploader's functionality happens while `HandleWALPush` uploads the initial WAL file, I expect 1 scan & process iteration to be sufficient. Example:
- 10 "ready" WAL segments are on disk
- `wal-g wal-push` the first file
- HandleWALPush starts uploading the first file and invokes BgUploader.start()
- BgUploader scans and finds the other 9 files
- BgUploader starts uploading the other 9 files
- HandleWALPush completes uploading the first file and calls BgUploader.Stop()
- wait until BgUploader.Stop() returns, indicating that all 9 files are complete
- exit

Assuming that all files are roughly the same size and therefore take the same amount of time to upload, it shouldn't be necessary or possible for scanAndProcessFiles need to do more than 1 loop (scan the filesystem more than once).

Alternatively, we could try adopting something like https://github.com/fsnotify/fsnotify to avoid scanning the entire directory (probably unnecessary and out of scope for this PR/project).

x4m: We could reuse open TCP connections to upload WALs if we upload not 9 but 32 files. And avoid wal-g startup costs. And if ls is so long - I think we should memorize results from previous run if it was not 100% used.

bunsenmcdubbs: I've reduced this to 100 milliseconds.

bunsenmcdubbs:
@x4m PR is ready to merge! If possible, I'd like to cut a pre-release to (load) test in a production-like environment.

x4m:
OK, LGTM. Let's merge it for QA.
Can you please also send a PR with cache for ReadDir call?
https://github.com/wal-g/wal-g/pull/644/files#diff-44bea216d3dc443a08f48b2ad2d54c49R113
You mentioned 1 second ```ls```.  It's kind of bottleneck which was not visible to me before, but I understand how this can be a problem.

bunsenmcdubbs:
> Can you please also send a PR with cache for ReadDir call?
https://github.com/wal-g/wal-g/pull/644/files#diff-44bea216d3dc443a08f48b2ad2d54c49R113
You mentioned 1 second ls. It's kind of bottleneck which was not visible to me before, but I understand how this can be a problem.

@x4m I'm going to do a bit more of a general investigation making further changes. Slow `ls` may not ultimately be my/the bottleneck

Issue #632, opened 2020-04-17
hang with `WALG_LIBSODIUM_KEY`


nwns:
When `WALG_LIBSODIUM_KEY` is set, wal-push hangs.  The below is the result of a kill -ABRT on the wal-g process.

wal-g was complied in docker in an alpine-linux 3.10 base image because we run postgres on alpine-linux 3.10 in GKE production.

I have encoutered this hang with both the FILE backup and S3 (using minio as documented in wal-g documentation).  I have not tried it with GCS yet as the test cluster does not use GCS for its object store (it uses minio).

Thanks,

```git clone https://github.com/wal-g/wal-g.git wal-g -b v0.2.15 \
    && cd wal-g \
    && export GOBIN=/usr/local/bin \
    && export USE_LIBSODIUM=true \
    && make install deps pg_build pg_install \
    && cd / \
    && rm -r wal-g
```

```
SIGABRT: abort
PC=0x556a6e945e71 m=0 sigcode=0

goroutine 0 [idle]:
runtime.futex(0x556a708d7d28, 0x80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7ffc784a92d8, 0x556a6e8f4e45, ...)
	/usr/lib/go/src/runtime/sys_linux_amd64.s:535 +0x21
runtime.futexsleep(0x556a708d7d28, 0xc000000000, 0xffffffffffffffff)
	/usr/lib/go/src/runtime/os_linux.go:46 +0x4b
runtime.notesleep(0x556a708d7d28)
	/usr/lib/go/src/runtime/lock_futex.go:151 +0xa5
runtime.stopm()
	/usr/lib/go/src/runtime/proc.go:1936 +0xc5
runtime.findrunnable(0xc000044a00, 0x0)
	/usr/lib/go/src/runtime/proc.go:2399 +0x550
runtime.schedule()
	/usr/lib/go/src/runtime/proc.go:2525 +0x220
runtime.park_m(0xc00008bb00)
	/usr/lib/go/src/runtime/proc.go:2605 +0xa7
runtime.mcall(0x7ffc784a94d8)
	/usr/lib/go/src/runtime/asm_amd64.s:299 +0x53

goroutine 1 [select, 7 minutes]:
io.(*pipe).Write(0xc00038a0f0, 0xc000546880, 0x18, 0x18, 0x0, 0x0, 0x0)
	/usr/lib/go/src/io/pipe.go:87 +0x1d9
[...]

PR #646, opened 2020-05-10
Fix libsodium hanging

defntvdm:
Should solve #632

### DIFF
--- a/internal/crypto/libsodium/reader.go
+++ b/internal/crypto/libsodium/reader.go
@@ -25,9 +25,9 @@ type Reader struct {
 	outLen int
 
 	// In case of using io.Pipe we can't read header until writer doesn't write, therefor we use these sync
-	o         sync.Once
-	key       []byte
-	headerErr error
+	onceHeader sync.Once
+	key        []byte
+	headerErr  error
 }
 
 // NewReader creates Reader from ordinary reader and key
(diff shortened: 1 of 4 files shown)


internal/crypto/libsodium/reader.go
```diff
@@ -22,40 +23,56 @@ type Reader struct {
 
 	outIdx int
 	outLen int
+
+	// In case of using io.Pipe we can't read header until writer doesn't write, therefor we use these sync
+	o         sync.Once
```
g0djan, line 28: "o" rename it

internal/crypto/libsodium/writer.go
```diff
@@ -21,36 +22,49 @@ type Writer struct {
 	out []byte
 
 	inIdx int
+
+	// In case of using io.Pipe we can't write header until reader doesn't read, therefor we use these sync
+	o         sync.Once
```
g0djan, line 27: And this

g0djan:
Hi, thanks a lot for this fix. Can you explain the root causes of hanging wal-push?
I understand that it could happen without crypter.mutex.RUnlock(). What did you fix in other files?

defntvdm:
Hi! The root of problem is io.Pipe. I wrote comment above onceHeader in libsodium Reader and Writer. We started to write header in Writer creation, so it locked until Reader start read, but Reader will never start read, because it created after Writer. And the same problem in Reader.
RUnlock don't solve problem, it's just good practice to unlock everything we locked.

internal/crypto/libsodium/reader.go
defntvdm: renamed

internal/crypto/libsodium/writer.go
defntvdm: renamed

PR #651, opened 2020-05-16
Sustain high-parallelism in wal-push during archive lag

bunsenmcdubbs:
This PR implements a lightweight performance optimization which should help sustain high wal-push concurrency even when WAL archiving falls behind.

When WAL archiving lags significantly and `pg_wal` grows very large (in my experience: >500GB), this can reduce the actual parallelism of wal-push. Before, BgUploader solely used `ioutil.ReadDir` to find other WAL files to upload while the "main" WAL file is being uploaded. When the main upload is complete, BgUploader is stopped - all running uploads are allowed to complete but no new files are enqueued. If `ioutil.ReadDir` is very slow (worst case: slower than main WAL upload), this can cause BgUploader to exit before it even uploads any WALs. This effectively reduces wal-push parallelism.

This PR adds "speculative WAL filename generation" which effectively short-circuits the filesystem scan (which finds actual WAL files on disk) and guesses for subsequent WAL files which are expected to exist given the main WAL filename. This process races against `ioutil.ReadDir` and is stopped as soon as the filesystem scan returns successfully (or a maximum number of guesses have been made, which ever comes first).

Test coverage in BgUploader has been extended to simulate a slow filesystem scan. The individual commits should be reviewable.

### DIFF
--- a/internal/bguploader.go
+++ b/internal/bguploader.go
@@ -166,7 +166,8 @@ func (b *BgUploader) speculativelyGenerateNextWalFilenames(ctx context.Context, 
 	for i := 0; i < 100*int(b.maxNumUploaded); i++ {
 		nextWalSegment, err := GetNextWalFilename(walSegment)
 		if err != nil {
-			tracelog.InfoLogger.Printf("failed to parse %s as wal segment name: %v", walSegment, err)
+			// couldn't generate next WAL segment name, stop try to
+			// guess next WAL
 			return
 		}
 
--- a/internal/bguploader.go
+++ b/internal/bguploader.go
@@ -39,7 +39,8 @@ type BgUploader struct {
 
 	// ctx signals internals to keep/stop enqueueing more uploads
 	ctx context.Context
-	// cancelFunc signals internals to stop enqueuing more uploads
+	// cancelFunc signals internals to stop enqueuing more uploads. It is
+	// safe to call this function repeatedly and concurrently.
 	cancelFunc context.CancelFunc
 
 	// workerCountSem tracks number of concurrent uploaders. Limited to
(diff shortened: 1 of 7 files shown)


internal/bguploader.go
```diff
@@ -135,23 +158,47 @@ func (b *BgUploader) scanAndProcessFiles() {
...
 
+// speculativelyGenerateNextWalFilenames continually generates and sends the
+// subsequent WAL segment names after b.initialWalSegment to fileChan until ctx
+// is canceled or 100 * b.maxNumUploaded filenames have been generated.
+func (b *BgUploader) speculativelyGenerateNextWalFilenames(ctx context.Context, fileChan chan<- string) {
+	walSegment := b.initialWalSegment
+	for i := 0; i < 100*int(b.maxNumUploaded); i++ {
+		nextWalSegment, err := GetNextWalFilename(walSegment)
+		if err != nil {
+			tracelog.InfoLogger.Printf("failed to parse %s as wal segment name: %v", walSegment, err)
```
bunsenmcdubbs, line 169: TODO: this is probably an irrelevant log message for most/all wal-g users. I'll remove this.

internal/bguploader_test.go
```diff
@@ -185,3 +230,40 @@ func cleanup(t *testing.T, dir string) {
...
+	dirname      string
+	ctx          context.Context
+	waitDuration time.Duration
+}
+
+func (s *stallableDataFolder) ListFilenames() ([]string, error) {
+	select {
+	case <-s.ctx.Done():
+		return nil, fmt.Errorf(
+			"fake dir reader was cancelled before wait completed: %v",
```
bunsenmcdubbs, line 248: TODO: for this functionality to be correct, it shouldn't return an error (that triggers a separate code path in BgUploader - either return empty list with no error or actual ReadDir results (like happy path).

internal/bguploader.go
```diff
@@ -135,23 +158,48 @@ func (b *BgUploader) scanAndProcessFiles() {
 
 }
 
+// speculativelyGenerateNextWalFilenames continually generates and sends the
+// subsequent WAL segment names after b.initialWalSegment to fileChan until ctx
+// is canceled or 100 * b.maxNumUploaded filenames have been generated.
+func (b *BgUploader) speculativelyGenerateNextWalFilenames(ctx context.Context, fileChan chan<- string) {
+	walSegment := b.initialWalSegment
+	for i := 0; i < 100*int(b.maxNumUploaded); i++ {
```
x4m, line 166: Why 100?

bunsenmcdubbs: This was arbitrary - I thought we should have _some_ limit here. That said, it is a bit paranoid. I'm fine with removing it - let me know what you think

x4m: 1. I cannot find limits of one upload anywhere else
2. It seems like here we should have infinite loop?
3. You still synchronously read directory in a please where result of chan passed to speculativelyGenerateNextWalFilenames() is used

I simply do not understand how it's intended to work

bunsenmcdubbs:
Closing for now because I don't intend on working on this for the foreseeable future

PR #660, opened 2020-05-27
Add gomaxprocs to config

munakoiso:


x4m:
Isn't it just too late to set GOMAXPROCs? Go process is already running with it's limits...

PR #661, opened 2020-05-28
SQLServer support (pre-alpha)

mialinx:


internal/databases/mongo/archive/layout.go
```diff
@@ -67,7 +67,7 @@ func SequenceBetweenTS(archives []models.Archive, since, until models.Timestamp)
 	return nil, fmt.Errorf("cycles in archive sequence detected")
 }
 
-// ArchivingResumeTS returns archiving Start timestamp
+// ArchivingResumeTS returns archiving Offset timestamp
```
x4m, line 70: I think this is extraneous change

internal/databases/mongo/models/archive.go
```diff
@@ -30,7 +30,7 @@ type Archive struct {
 // NewArchive builds Archive struct with given arguments
 func NewArchive(start, end Timestamp, ext string, atype string) (Archive, error) {
 	if LessTS(end, start) {
-		return Archive{}, fmt.Errorf("malformed archive, Start timestamp < End timestamp: %s < %s", start, end)
+		return Archive{}, fmt.Errorf("malformed archive, Offset timestamp < End timestamp: %s < %s", start, end)
```
x4m, line 33: I think this is extraneous change

mialinx: fixed

PR #662, opened 2020-05-29
add libsodium to travis build

g0djan:


lachesis:
Does this being unmerged mean that the release version available here does not support libsodium?
https://github.com/wal-g/wal-g/releases/tag/v0.2.19

serprex:
Closeing as `USE_LIBSODIUM=1` is used in .github already & .travis.yml is gone

PR #663, opened 2020-05-31
Mysql incremental backups

a-shchupakov:
Added possibility to make incremental backups of MySQL-based servers

### DIFF
--- a/cmd/mysql/backup_push.go
+++ b/cmd/mysql/backup_push.go
@@ -12,10 +12,6 @@ const (
 
 	FullBackupFlag      = "full"
 	FullBackupShorthand = "f"
-
-	XtrabackupExtraLsnDir    = "--extra-lsndir"
-	XtrabackupIncrementalLSN = "--incremental-lsn"
-	XtrabackupCheckpoints    = "xtrabackup_checkpoints"
 )
 
 var (
--- a/cmd/mysql/backup_push.go
+++ b/cmd/mysql/backup_push.go
@@ -7,21 +7,35 @@ import (
 	"github.com/wal-g/wal-g/internal/databases/mysql"
 )
 
-const backupPushShortDescription = "Creates new backup and pushes it to storage"
+const (
+	backupPushShortDescription = "Creates new backup and pushes it to storage"
 
-// backupPushCmd represents the streamPush command
-var backupPushCmd = &cobra.Command{
-	Use:   "backup-push",
-	Short: backupPushShortDescription,
-	Run: func(cmd *cobra.Command, args []string) {
-		uploader, err := internal.ConfigureUploader()
-		tracelog.ErrorLogger.FatalOnError(err)
-		backupCmd, err := internal.GetCommandSetting(internal.NameStreamCreateCmd)
-		tracelog.ErrorLogger.FatalOnError(err)
-		mysql.HandleBackupPush(uploader, backupCmd)
-	},
-}
+	FullBackupFlag      = "full"
+	FullBackupShorthand = "f"
+
+	XtrabackupExtraLsnDir    = "--extra-lsndir"
+	XtrabackupIncrementalLSN = "--incremental-lsn"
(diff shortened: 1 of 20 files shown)


a-shchupakov:
By the way, how do you grant permissions to run new integration test?

g0djan:
chmod +x ?

internal/databases/mysql/backup_push_handler.go
```diff
@@ -19,21 +94,80 @@ func HandleBackupPush(uploader *internal.Uploader, backupCmd *exec.Cmd) {
...
+		if err != nil {
+			tracelog.WarningLogger.Printf("Failed to read to_lsn. Future deltas cannot be done on top of the current backup: %v", err)
+		}
 	}
 
 	binlogEnd := getMySQLCurrentBinlogFile(db)
 	tracelog.DebugLogger.Println("Binlog end file", binlogEnd)
-	sentinel := StreamSentinelDto{BinLogStart: binlogStart, BinLogEnd: binlogEnd, StartLocalTime: timeStart}
 
+	sentinel := BackupSentinelDto{
```
mialinx, line 120: breaks backward compatibility

ostinru:
Wow! It was an exciting patch :)

It seems that we should also fallback to full backup when `serer-uuid` have changed (I don't expect to have the same LSNs on all nodes of the cluster)

x4m:
@ostinru +1
is there any chance to revivew this?

cmd/mysql/backup_push.go
```diff
@@ -7,21 +7,31 @@ import (
...
+			mysql.HandleBackupPush(uploader, backupCmd, fullBackup)
+		},
+	}
+	fullBackup = false
+)
 
 func init() {
 	Cmd.AddCommand(backupPushCmd)
+
+	backupPushCmd.Flags().BoolVarP(&fullBackup, FullBackupFlag, FullBackupShorthand, false, "Make full backup-push")
```
mialinx, line 36: This should be a default for backaward compatibility.
Incremental should be an option

internal/databases/mysql/backup_push_handler.go
```diff
@@ -1,14 +1,89 @@
...
+}
+
+func createAndPushBackup(
+	uploader *internal.Uploader, backupCmd *exec.Cmd,
+	previousBackupSentinelDto internal.BackupSentinelDto,
+	incrementCount int,
+) {
+	extraLsnDir := path.Join("/tmp", "extra_lsn")
+	isXtrabackup := isXtrabackupCmd(backupCmd)
+	fromLSN, previousBackupName := previousBackupSentinelDto.BackupFinishLSN, previousBackupSentinelDto.IncrementFullName
```
mialinx, line 75: LSN in MySQL (innodb) has meaning only inside the single instance.
Backups may be taken from different instances (primary or one of secondary replicas).
So if current backup and previous are from different instances - we can rely on the LSN.

But we still can rely on LSN as long as we are taking backup from the same server

internal/databases/mysql/backup_fetch_handler.go
```diff
@@ -1,16 +1,125 @@
...
+
+const (
+	XtrabackupApplyLogOnly   = "--apply-log-only"
+	XtrabackupIncrementalDir = "--incremental-dir"
 )
 
 func HandleBackupFetch(folder storage.Folder, backupName string, restoreCmd *exec.Cmd, prepareCmd *exec.Cmd) {
+	var err error
+	if isXtrabackupCmd(restoreCmd) {
+		tempDeltaDir := path.Join("/tmp", "delta")
```
mialinx, line 24: `/tmp`  is typically very limited in free space, database diff may take far more space
we need an option to choose directory for backup diffs

a-shchupakov:
Feel free to catch up on this PR if you want. I'm afraid I don't remember a thing about it

ostinru:
Closed as we merged another version of MySQL incremental backups (with some picks from this PR).
Thank you everyone for initial PR and Suggestions on how to improve it.

PR #666, opened 2020-06-03
yet another fix require variables mysql

munakoiso:


g0djan:
internal.RequiredSettings[internal.MysqlDatasourceNameSetting] = true
Is it outside of init now because backup-fetch doesn't require it?

munakoiso:
@g0djan, yes
this variable only required here https://github.com/wal-g/wal-g/blob/f6abd0cc33ae03c5133f5e38148ec0488379953c/internal/databases/mysql/mysql.go#L63

And this function is used only in backup_push and binlog_push
https://github.com/wal-g/wal-g/search?q=getMySQLConnection&unscoped_q=getMySQLConnection

PR #670, opened 2020-06-08
Refactoring for reverse delta unpack

usernamedt:
Since the implementation of reverse delta unpacking, the unpack logic for default backups (backup-push) and for catchup backups (catchup-push) began to differ even more.

To improve the code readability and maintainability I propose dividing the catchup backup and default backup unwrap logic. I created an interface and added two implementations for backup file unwrap: one for default backup, one for catchup backup.

### DIFF
--- a/internal/catchup_file_unwrapper.go
+++ b/internal/catchup_file_unwrapper.go
@@ -14,7 +14,11 @@ type CatchupFileUnwrapper struct {
 
 func (u *CatchupFileUnwrapper) UnwrapNewFile(reader io.Reader, header *tar.Header, file *os.File) error {
 	if u.options.isIncremented {
-		err := CreateFileFromIncrement(reader, file)
+		fileInfo, err := file.Stat()
+		if err != nil {
+			return err
+		}
+		err = CreateFileFromIncrement(reader, NewReadWriterAtFrom(file, fileInfo))
 		return errors.Wrapf(err, "Interpret: failed to create file from increment '%s'", file.Name())
 	}
 
@@ -23,7 +27,11 @@ func (u *CatchupFileUnwrapper) UnwrapNewFile(reader io.Reader, header *tar.Heade
 
 func (u *CatchupFileUnwrapper) UnwrapExistingFile(reader io.Reader, header *tar.Header, file *os.File) error {
 	if u.options.isIncremented {
-		err := WritePagesFromIncrement(reader, file, true)
+		fileInfo, err := file.Stat()
+		if err != nil {
+			return err
+		}
+		err = WritePagesFromIncrement(reader, NewReadWriterAtFrom(file, fileInfo), true)
 		return errors.Wrapf(err, "Interpret: failed to write increment to file '%s'", file.Name())
 	}
 
--- a/internal/catchup_file_unwrapper.go
+++ b/internal/catchup_file_unwrapper.go
@@ -1,0 +1,37 @@
+package internal
+
+import (
+	"archive/tar"
+	"github.com/pkg/errors"
+	"io"
+	"os"
+)
+
+// CatchupFileUnwrapper is used for catchup (catchup-push) backups
+type CatchupFileUnwrapper struct {
+	BackupFileUnwrapper
+}
+
+func (u *CatchupFileUnwrapper) UnwrapNewFile(reader io.Reader, header *tar.Header, file *os.File) error {
+	if u.options.isIncremented {
+		err := CreateFileFromIncrement(reader, file)
+		return errors.Wrapf(err, "Interpret: failed to create file from increment '%s'", file.Name())
+	}
+
+	return u.writeLocalFile(reader, header, file)
+}
+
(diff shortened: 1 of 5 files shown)


internal/file_unwrapper.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type BackupFileOptions struct {
+	isIncremented bool
+	isPageFile    bool
+}
+
+type IBackupFileUnwrapper interface {
+	UnwrapNewFile(reader io.Reader, header *tar.Header, file *os.File) error
+	UnwrapExistingFile(reader io.Reader, header *tar.Header, file *os.File) error
```
g0djan, line 36: It's better to pass interfaces like Reader, Writer, ReaderWriter instead of a pointer to a file. It makes writing unit tests easier.

internal/tar_interpreter_new.go
```diff
@@ -11,69 +11,62 @@ import (
...
+		return NewFileUnwrapper(CatchupBackupFileUnwrapper, options)
 	}
-	return unwrapToNewFile(fileReader, fileInfo, targetPath, isIncremented)
+	return NewFileUnwrapper(DefaultBackupFileUnwrapper, options)
 }
 
 // unwrap the file from tar to existing local file
-func unwrapToExistFile(fileReader io.Reader, fileInfo *tar.Header, targetPath string,
-	isPageFile, isIncremented, isCatchup bool) error {
+func unwrapToExistFile(fileReader io.Reader, header *tar.Header, targetPath string, unwrapper IBackupFileUnwrapper) error {
```
g0djan, line 48: Cool!

internal/tar_interpreter_new.go
```diff
@@ -11,69 +11,62 @@ import (
...
-	}
-	// skip the non-page file because newer version is already on the disk
-	return nil
+	return unwrapper.UnwrapExistingFile(fileReader, header, localFile)
 }
 
 // unwrap file from tar to new local file
-func unwrapToNewFile(fileReader io.Reader, fileInfo *tar.Header, targetPath string, isIncremented bool) error {
-	localFile, err := createLocalFile(targetPath, fileInfo)
+func unwrapToNewFile(fileReader io.Reader, header *tar.Header, targetPath string, unwrapper IBackupFileUnwrapper) error {
```
g0djan, line 58: It's used only in one place. And as I got it doesn't contain specific logic except for open file. Now it looks like a method of IBackupFileUnwrapper. But actually it isn't.

usernamedt: So, I assume there is two ways to solve this:
1. Move this file creating/opening logic to IBackupFileUnwrapper 
2. Move this back to the place it is called and just get rid of the separate function

Which will be better?

internal/file_unwrapper.go
usernamedt: Since we need file-specific methods later, like file.Sync(), file.WriteAt(), you propose to wrap all these methods in interface?

g0djan: Hm, I just mean that if you can extract some logic from using file, it's better to do it. Because it will be possible to test it. But if there is no logic without using files it will be useless.

internal/tar_interpreter_new.go
g0djan: I vote for second

usernamedt: fixed in 4a91a430a19ebdcae18eebccf9a061de16a85156

internal/file_unwrapper.go
usernamedt: fixed in 4a91a430a19ebdcae18eebccf9a061de16a85156

internal/pagefile_new.go
```diff
@@ -15,11 +15,17 @@ import (
...
+	io.ReaderAt
+	io.WriterAt
+	Stat() (os.FileInfo, error)
+	Name() string
+}
+
 // RestoreMissingPages restores missing pages (zero blocks)
 // of local file with their base backup version
-func RestoreMissingPages(base io.Reader, file *os.File) error {
+func RestoreMissingPages(base io.Reader, file ReadWriterAt) error {
```
g0djan, line 27: "file" is not a file anymore. Rename it in all functions with such inconvenient naming.

usernamedt: Renamed in 2e4979b381e5d441acaa7269833595e39f651f7f

internal/catchup_file_unwrapper.go
```diff
@@ -14,7 +14,11 @@ type CatchupFileUnwrapper struct {
 
 func (u *CatchupFileUnwrapper) UnwrapNewFile(reader io.Reader, header *tar.Header, file *os.File) error {
 	if u.options.isIncremented {
-		err := CreateFileFromIncrement(reader, file)
+		fileInfo, err := file.Stat()
```
g0djan, line 17: Instead of repeat this logic in each place you can inline it to NewReadWriterAtFrom

PR #672, opened 2020-06-08
Replace FastCopy sync.Pool by simple channel-based pool

perekalov:
Unfortunately the previous approach (#668) doesn't work in real-life applications.
GC periodically collects unreferenced []byte from sync.Pool, so for long-lived objects we need to maintain our own pool.

```
$ go test -bench=.  -benchmem -benchtime=5x
goos: darwin
goarch: amd64
pkg: github.com/wal-g/wal-g/utility
BenchmarkTestFastCopy/sync_pool_block/2-4         	       5	2540738674 ns/op	 7267673 B/op	   13001 allocs/op
BenchmarkTestFastCopy/ch_pool_block/2-4           	       5	2398465157 ns/op	 3041155 B/op	   12000 allocs/op
BenchmarkTestFastCopy/sync_pool_block*2-4         	       5	12900628822 ns/op	26142048 B/op	   13002 allocs/op
BenchmarkTestFastCopy/ch_pool_block*2-4           	       5	12157811518 ns/op	17721219 B/op	   12000 allocs/op
PASS
ok  	github.com/wal-g/wal-g/utility	183.743s
```

### DIFF
--- a/utility/utility.go
+++ b/utility/utility.go
@@ -125,23 +125,47 @@ func GetSubdirectoryRelativePath(subdirectoryPath string, directoryPath string) 
 	return NormalizePath(SanitizePath(strings.TrimPrefix(subdirectoryPath, directoryPath)))
 }
 
-var copyBufPool = sync.Pool{New: func() interface{} {
-	return make([]byte, CopiedBlockMaxSize)
-}}
-
-func getCopyBuf() []byte {
-	return copyBufPool.Get().([]byte)
+// BytesPool holds []byte.
+type BytesPool struct {
+	pool chan []byte
 }
 
-func putCopyBuf(buf []byte) {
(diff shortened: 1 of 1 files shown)


utility/utility.go
```diff
@@ -125,23 +125,47 @@ func GetSubdirectoryRelativePath(subdirectoryPath string, directoryPath string)
...
+// NewBytesPool creates a new pool of Clients.
+func NewBytesPool(max int) *BytesPool {
+	return &BytesPool{
+		pool: make(chan []byte, max),
+	}
 }
 
-func putCopyBuf(buf []byte) {
-	copyBufPool.Put(buf)
+const CopyBytesPoolSize = 2
```
perekalov, line 140: We don't have convenient way to set size of the pool, so just provide default for now.

PR #687, opened 2020-06-24
Backup-fetch redundant archives skipping

usernamedt:
This is a draft.
The end goal of this feature is to skip the download of unnecessary files during backup-fetch.

What is implemented:
1. Collecting statistics about page files updates count from pg_stat_all_tables 
Seems like it is working

2. Recording this statistics into backup sentinel json
Seems like it is working

3. Composing files into tarballs based on their updates count change
It is kinda working, but need adjustments:
-currently, it does not support parallelism, the way I see to implement it is to divide the sorted file list from TarBundleComposer into N equal parts and launch a goroutine for each part
-todo: assigning the _vm and _fsm files the same updates count as page file or implement some other way to ensure that they will go into the same archive as the corresponding page files
-maybe add some other heuristics to achieve better results
-also, the bundle.go seems like a huge file now (~800 rows), maybe it deserves some decomposition
-TarBundleComposer was initially designed to create tarballs inside himself, but idk the correct way to pass the corresponding methods from bundle.go, it seems like including Bundle in TarBundleComposer is not a great idea...

4. Collecting unwrap results in tar_interpreter to calculate which files are completed and no longer need any new information
Seems like it is working 

5. Excluding completed files from the list of the files to unwrap during backup-fetch.
Seems like it is working 

6. Skipping the download of tarballs without interesting files
Seems like it is working

PR #702, opened 2020-07-08
Fixes #701

### LINKED ISSUE #701: Wrong metadata, uncompressed_size reflects size of single part of the backup.
In case if tar file is bigger than WALG_TAR_SIZE_THRESHOLD, uncompressed_size stores only size of the last tar file and doesn't include previous parts of tar file.

yanchenko-igor:
@x4m Could you have a look if this is a good approach? It fixes the issue with wrong metadata.

### DIFF
--- a/internal/backup_push_handler.go
+++ b/internal/backup_push_handler.go
@@ -128,13 +128,13 @@ func createAndPushBackup(
 	tracelog.ErrorLogger.FatalOnError(err)
 	err = bundle.FinishQueue()
 	tracelog.ErrorLogger.FatalOnError(err)
-	uncompressedSize := atomic.LoadInt64(bundle.AllTarballsSize)
-	compressedSize := atomic.LoadInt64(uploader.tarSize)
 	err = bundle.UploadPgControl(uploader.Compressor.FileExtension())
 	tracelog.ErrorLogger.FatalOnError(err)
 	// Stops backup and write/upload postgres `backup_label` and `tablespace_map` Files
 	finishLsn, err := bundle.uploadLabelFiles(conn)
 	tracelog.ErrorLogger.FatalOnError(err)
+	uncompressedSize := atomic.LoadInt64(bundle.AllTarballsSize)
+	compressedSize := atomic.LoadInt64(uploader.tarSize)
 
 	timelineChanged := bundle.checkTimelineChanged(conn)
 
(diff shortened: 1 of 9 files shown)


x4m:
@usernamedt this, probably, interferes much with your work. Can you please review this PR?

internal/bundle.go
```diff
@@ -184,6 +187,7 @@ func (bundle *Bundle) CheckSizeAndEnqueueBack(tarBall TarBall) error {
 		bundle.mutex.Lock()
 		defer bundle.mutex.Unlock()
 
+		atomic.AddInt64(bundle.AllTarballsSize, tarBall.Size())
```
usernamedt, line 190: Seems like this row should be added before each .CloseTar() call, not only in CheckSizeAndEnqueueBack(), but in uploadLabelFiles, UploadPGControl, FinishQueue too

usernamedt:
I've noticed this issue too, fix seems legit except one issue that I commented.

As for my PR https://github.com/wal-g/wal-g/pull/687 there is definitely some interference but since I don't use or change the uncompressedSize logic it seems like OK. Maybe some more merging will be required :)

internal/bundle.go
yanchenko-igor: It wasn't necessary because I used current tarball size together with bundle.AllTarballssize 
```	uncompressedSize := bundle.TarBall.Size() + atomic.LoadInt64(bundle.AllTarballsSize)``` but it will work as well if we do it each time before closing tarball.

Comment (usernamedt) [reply internal/bundle.go:191]: I think it is a good idea to extract this increment to a method to avoid duplication

Comment (yanchenko-igor) [reply internal/bundle.go:191]: yeah, I already pushed the change.

Comment (yanchenko-igor) [reply internal/bundle.go]: fixed that.

File: internal/backup_push_handler.go [hunk 2]
```diff
@@ -128,7 +128,7 @@ func createAndPushBackup(
 	tracelog.ErrorLogger.FatalOnError(err)
 	err = bundle.FinishQueue()
 	tracelog.ErrorLogger.FatalOnError(err)
-	uncompressedSize := bundle.TarBall.Size()
+	uncompressedSize := atomic.LoadInt64(bundle.AllTarballsSize)
 	compressedSize := atomic.LoadInt64(uploader.tarSize)
```
usernamedt, line 132: These two size assignments are executed before uploadPGControl and uploadlabelfiles tars, I think they should be assigned after.

yanchenko-igor: fixed.

usernamedt:
Seems legit for me

Need a review from someone with write access @x4m

x4m:
@usernamedt you have write access :) Just accept the invite.

PR #705, opened 2020-07-13
Bump storages to update GCS SDK

x4m:


kustodian:
Do you plan to make a pre-release?

x4m:
@kustodian eventually, [pre]release is unavoidable, but right now CI seems to be reddish...

kustodian:
As far as I understand those tests failing are S3 tests, but the changes you made shouldn't affect S3. So it's probably some other change that created the issue. Right?

x4m:
Yes. And we do not know which one. CI was green, than flappy brown, then red.

x4m:
probably something changed in dependencies, e.g. minio

PR #707, opened 2020-07-16
Reduce pgbench sizes to fix red CI

x4m:


defntvdm:
Hi, you need to reduce row_count [here](https://github.com/wal-g/wal-g/blob/master/docker/pg_tests/scripts/tests/catchup_test.sh#L57) too

x4m:
@defntvdm thanks! Is test still viable to test functionality?

defntvdm:
Yes, it's correct)

PR #709, opened 2020-07-18
Create .walg.json

satyamuralidhar:


satyamuralidhar:
.walg.json reference for minio cloud parameters

PR #718, opened 2020-08-04
Add mongodb backup-delete

perekalov:


g0djan:
Hey, what is the reason for creating a separate handler for deleting concrete backup for a concrete database?
We already have functions that require passing selector as an argument and it allows delete only files that you want.

PR #720, opened 2020-08-07
Add retries to upload logic

jschaf:
When uploading large amounts of data to S3, we occasionally see failures where the AWS sdk tries to use a closed network connection. The upstream bug appears to be https://github.com/aws/aws-sdk-go/issues/3406.  I'm not sure why the error manifests but it's causing us significant pain. Rather than retry the entire base backup, we'll retry the WAL segment upload.

```
ERROR: 2020/08/06 08:39:52.198782 failed to upload 'basebackups_005/base_0000000100006F04000000E4/tar_partitions/part_19148.tar.br' to bucket 'S3_BUCKET':
    MultipartUpload: upload multipart failed
caused by: RequestError: send request failed
caused by: Put https://S3_BUCKET/basebackups_005/base_0000000100006F04000000E4/tar_partitions/part_19148.tar.br?partNumber=2:
    write tcp 10.64.18.161:42118->52.216.134.19:443: use of closed network connection
ERROR: 2020/08/06 08:39:52.198805 upload: could not upload 'base_0000000100006F04000000E4/tar_partitions/part_19148.tar.br'
ERROR: 2020/08/06 08:39:52.198818 failed to upload 'basebackups_005/base_0000000100006F04000000E4/tar_partitions/part_19148.tar.br' to bucket 'S3_BUCKET':
    MultipartUpload: upload multipart failed
caused by: RequestError: send request failed
caused by: Put https://S3_BUCKET/basebackups_005/base_0000000100006F04000000E4/tar_partitions/part_19148.tar.br?partNumber=2
    write tcp 10.64.18.161:42118->52.216.134.19:443: use of closed network connection
ERROR: 2020/08/06 08:39:52.198833 Unable to complete uploads
```

### DIFF
--- a/internal/uploader.go
+++ b/internal/uploader.go
@@ -100,12 +100,18 @@ func (uploader *Uploader) Upload(path string, content io.Reader) error {
 	if uploader.tarSize != nil {
 		content = &WithSizeReader{content, uploader.tarSize}
 	}
-	err := uploader.UploadingFolder.PutObject(path, content)
-	if err == nil {
-		return nil
+	// Add retries to work around https://github.com/aws/aws-sdk-go/issues/3406
+	const retries = 3
+	var err error
+	for i := 0; i < retries; i++ {
+		err = uploader.UploadingFolder.PutObject(path, content)
+		if err == nil {
+			return nil
+		}
+		tracelog.ErrorLogger.Printf(tracelog.GetErrorFormatter()+"Retrying upload error:\n", err)
 	}
+	tracelog.ErrorLogger.Printf(tracelog.GetErrorFormatter()+"Exhausted upload retries:\n", err)
 	uploader.Failed.Store(true)
-	tracelog.ErrorLogger.Printf(tracelog.GetErrorFormatter()+"\n", err)
 	return err
 }
 


bunsenmcdubbs:
I'll take a closer look in the next couple days/early next week. @jschaf this fix may be better applied in wal-g/storages in order to isolate the impact to just s3 and preserve the abstraction.

jschaf:
> @jschaf this fix may be better applied in wal-g/storages in order to isolate the impact to just s3 and preserve the abstraction.

I think moving it to storages would increase the impact since anyone that uses the library now has retries whether they want it or not.

x4m:
The problem is that you just cannot retry part of an already consumed stream. Other end of the stream is reading from disk and you just cannot reset it...

jschaf:
> The problem is that you just cannot retry part of an already consumed stream.

Oh, good catch. Are the files ever buffered to disk or is everything streamed?

x4m:
> > The problem is that you just cannot retry part of an already consumed stream.
> 
> Oh, good catch. Are the files ever buffered to disk or is everything streamed?

WAL-G never stores temp files to disk if they are not intended to survive restart.
I.e. we do not want to occupy extra space, but we have caches of what WAL is prefetched, cache of public crypto keys and some others

PR #737, opened 2020-08-22
Receivewal

sebasmannem:


sebasmannem:
I fced some smaller issues I noticed with pg9.6. Also byte tested against pg_receivewal, and files are same, so 9.6 now works fine too.
I also noticed that Jackc merged my first commit, so I reverted some of the workarounds I added on that front.
Will now work on some tests added for my second commit as requested by Jackc.

internal/queryRunner.go
```diff
@@ -170,3 +185,47 @@ func (queryRunner *PgQueryRunner) stopBackup() (label string, offsetMap string,
...
+	if err != nil {
+		return 0, err
+	}
+	segBlocks, err = strconv.ParseUint(strValue, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	if queryRunner.Version <= 100000 {
+		// For PG 10 and below, wal_segment_size is in 8k blocks
+		segBlocks *= 8192
```
x4m, line 211: FWIW we currently only support 16Mb segments.... But I think this can be configured in future

internal/wal_segment.go
```diff
@@ -0,0 +1,222 @@
...
+// Name returns the filename of this wal segment.
+// This is also used by the WalUploader to set the name of the destination file during upload of the wal segment.
+func (seg *WalSegment) Name() string {
+	// Example LSN -> Name:
+	// '0/2A33FE00' -> '00000001000000000000002A'
+	segID := uint64(seg.StartLSN)/uint64(seg.walSegmentBytes)
+	if seg.IsComplete() {
+		return fmt.Sprintf("%08X%016X", seg.TimeLine, segID)
+	}
+	return fmt.Sprintf("%08X%016X.partial", seg.TimeLine, segID)
```
x4m, line 98: Recieveing timeline switch record ia a bit tricky.
Keep in mind that we need to archive partial segment on old timeline and full segmet on new timeline.

internal/queryRunner.go
sebasmannem: Correct, I think it will work with custom segment sizes, but I have not tested it yet.

internal/wal_segment.go
sebasmannem: Good thinking. I am now checking, and I don't see the partial segments being written out. Will fix.

x4m:
Finally with updated pgconn we will have SCRAM auth :)

internal/wal_segment.go
sebasmannem: Fixed.

x4m (changes requested):
LGTM, but one small thing.
Let's add IP or hostname to default slot name? We will need to escape it though to match slot name requirements.
How things can go:
1. DBA1 setup wal-recieve just to toy around and test new functionality, not setting up any monitoring for livelines of VM.
2. VM goes offline, it's a toy VM after all
3. Prod DB goes out of space due to disconnected slot
4. DBA2 tries to find out "what was that?", finds slot and blames WAL-G developers for fussup :)
OTOH if slot name is "walg_on_DBA1_VM", DBA2 will know who actually messed things up.

sebasmannem:
I see your point, and we could add that option.
What I fear for, is for unexpected slotname change. Like when DBA migrates wal-g from VM1 to VM2.
In that case wal-g would create a new slot, and leave old slot hanging around, creating two issues:
1: The new slot is ahead of the olds slot and with that wal-g has a gap. Easily fixable by moving to old slot, but note that VM2 would be streaming from slot walg_on_VM1, which is a bit awkward too.
2: The old slot is hanging around in the end causing 'ProdDB out of diskspace' issue.
There are other situations that could cause that behaviour too, like a container environment where containers have a different hostname on every reschedule.

Of course everyone could make it static by always setting WALG_SLOTNAME, but to me that still sounds like the static should be the default option.
If you want me to, I will make it dynamic, but I don't think we should...
Let me know either way...

x4m:
OK, makes sense. Static slot is dangerous, dynamic slot is dangerous. Great feature comes with great responsibility.

x4m (approved):
LGTM

x4m:
Pushed, thanks!

PR #746, opened 2020-08-26
Add wal-verify command

usernamedt:
In this PR I propose wal-verify command.

After launching wal-verify, WAL-G connects to the Postgres cluster and fetches the current cluster segment number and timeline.

Then, it traverses the WAL segments history in reversed chronological order and checks the WAL segments history integrity (finds missing segments).

Currently, there are 3 missing segments types:

- MISSING_DELAYED segments are not present in WAL storage, but probably Postgres did not try to archive them via archive_command yet
- MISSING_UPLOADING segments are the segments which are not present in WAL storage, but looks like that they are in the process of uploading to storage 
- MISSING_LOST segments are the segments which are not present in WAL storage and not MISSING_UPLOADING nor MISSING_DELAYED

Result of wal-verify command is report which consists of 2 parts:

1. WAL storage status, calculated with the following logic: 
-If there are no missing segments, then OK
-If there are some missing segments, but they are not MISSING_LOST, then WARNING
-If there are some MISSING_LOST segments, then it is a FAILURE

2. Report that shows WAL segments in chronological order grouped by Timeline and Status (FOUND/MISSING_DELAYED/MISSING_UPLOADING/MISSING_LOST).

Some examples of executing the wal-verify command

WAL storage status: FAILURE
image

WAL storage status: WARNING
image

WAL storage status: OK
image

Also, the output may be written as JSON with `--json` flag:
```json
{
   "storage_status":"FAILURE",
   "integrity_scan_result":[
      {
         "timeline_id":1,
         "start_segment":"000000010000000000000001",
         "end_segment":"000000010000000000000077",
         "segments_count":119,
         "status":"MISSING_LOST"
      },
      ...
      {
         "timeline_id":3,
         "start_segment":"000000030000000400000026",
         "end_segment":"000000030000000400000026",
         "segments_count":1,
         "status":"MISSING_LOST"
      },
      {
         "timeline_id":3,
         "start_segment":"000000030000000400000027",
         "end_segment":"00000003000000040000002F",
         "segments_count":9,
         "status":"FOUND"
      }
   ]
}
```


internal/wal_segment_runner.go
```diff
@@ -47,15 +47,21 @@ type WalSegmentRunner struct {
...
 func NewWalSegmentRunner(
 	startWalSegment WalSegmentDescription,
 	segments map[WalSegmentDescription]bool,
 	stopSegmentNo WalSegmentNo,
+	timelineSwitchMap map[WalSegmentNo]*TimelineHistoryRecord,
 ) *WalSegmentRunner {
-	return &WalSegmentRunner{startWalSegment,
-		segments, stopSegmentNo}
+	return &WalSegmentRunner{
+		currentWalSegment: startWalSegment,
```
g0djan, line 60: ->currentWalSegment: startWalSegment
rename one

internal/wal_segment_runner_test.go
```diff
@@ -27,40 +29,145 @@ func TestWalSegmentRunner_MissingSegments(t *testing.T) {
...
+	for timeline, segmentNumbers := range storageSegmentsByTimeline {
+		for _, segmentNo := range segmentNumbers {
+			segment := internal.WalSegmentDescription{Number: segmentNo, Timeline: timeline}
+			storageSegments[segment] = true
+		}
+	}
+	walSegmentRunner := internal.NewWalSegmentRunner(startSegment, storageSegments, stopSegmentNo, timelineSwitchMap)
+
+	// convert expectedSegmentsByTimeline
+	expectedSegments := make(map[internal.WalSegmentDescription]bool, 0)
```
g0djan, line 137: // convert storageSegmentsByTimeline
and
// convert expectedSegmentsByTimeline
have the same code, DRY

internal/wal_verify_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	integrityScanSegmentSequences, err := collapseSegmentsByStatusAndTimeline(segmentScanner.ScannedSegments)
+	tracelog.ErrorLogger.FatalOnError(err)
+
+	err = outputWriter.Write(newWalVerifyResult(integrityScanSegmentSequences))
+	tracelog.ErrorLogger.FatalOnError(err)
+}
+
+// runWalIntegrityScan invokes the following storage scan series:
+// 1. At first, it runs scan until it finds some segment in WAL storage
+// and marks all encountered missing segments as "missing, probably delayed"
```
g0djan, line 126: For example, we have 8 WAL segments locally and latest WAL segment in storage is 4'th.
So we mark 5,6,7,8 as Delayed

internal/wal_verify_handler.go
```diff
@@ -0,0 +1,219 @@
...
+// 2. Then it scans exactly uploadingSegmentRangeSize count of segments,
+// if found any missing segments it marks them as "missing, probably still uploading"
```
g0djan, line 128: Here we use maxUploadCOncurrency as uploadingSegmentRangeSize, for example, it will be 2.

So we will mark 7 and 8 as Uploading, and 5,6 will have the status delayed. But 5,6 appears earlier and probable these segments should have Uploading status instead of 7,8.

g0djan (changes requested):

internal/wal_verify_handler.go
usernamedt: In this case will mark all of them as delayed I guess. Uploading segment range size may contain or may not contain some segments, it is started exactly after some segment has been found on storage
I draw some illustration, forgot to pin to the PR
image

usernamedt: So red (missing) segments in the first group on the picture are MISSING_DELAYED, missing in the second group are  MISSING_UPLOADING, missing in third are MISSING_LOST

internal/wal_segment_runner.go
usernamedt: currentWalSegment is tend to change over iterations, startWalSegment is just the initial value

internal/queryRunner.go
```diff
@@ -53,6 +53,11 @@ func (queryRunner *PgQueryRunner) buildGetVersion() string {
 	return "select (current_setting('server_version_num'))::int"
 }
 
+// BuildGetCurrentLSN formats a query to get cluster LSN
+func (queryRunner *PgQueryRunner) buildGetLsn() string {
```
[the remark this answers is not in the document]
usernamedt: Fixed in https://github.com/wal-g/wal-g/pull/746/commits/2ac93e53ef134fc4d91b29e7c8b0117d2d222fac

internal/queryRunner.go
```diff
@@ -111,6 +116,13 @@ func (queryRunner *PgQueryRunner) getVersion() (err error) {
 	return errors.Wrap(err, "GetVersion: getting Postgres version failed")
 }
 
+// Get current LSN of cluster
+func (queryRunner *PgQueryRunner) getCurrentLsn() (lsn string, err error) {
+	conn := queryRunner.connection
+	err = conn.QueryRow(queryRunner.buildGetLsn()).Scan(&lsn)
```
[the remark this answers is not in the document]
usernamedt: Fixed in https://github.com/wal-g/wal-g/pull/746/commits/2ac93e53ef134fc4d91b29e7c8b0117d2d222fac

internal/wal_segment_runner_test.go
usernamedt: Fixed in https://github.com/wal-g/wal-g/pull/746/commits/b00b8d131ed0462e6aa017ea4771c094bedd41e2

internal/wal_verify_handler.go
usernamedt: I've added more docs in https://github.com/wal-g/wal-g/pull/746/commits/d4de3fe15bce6ac9281221cee14facc172c85a20

g0djan (approved):
Lgtm

g0djan:
Restart checks if there MySQL test failed again.

kustodian:
I'm curious about what is a use case for this, except for debugging when setting up WAL-G the first time? Could you provide an example?

usernamedt:
> I'm curious about what is a use case for this, except for debugging when setting up WAL-G the first time? Could you provide an example?

This command helps to verify that there is no "holes" in WAL segments sequence so you can do a PITR for the backup. You can find more details in this post https://usernamedt.me/google-summer-of-code-2020-at-postgresql/#walhistorycheck

kustodian:
Why would there be holes if WAL-G is working fine, archive command is working fine? I guess this could be used as an additional check if everything is fine, basically a double-check, right?

usernamedt:
> Why would there be holes if WAL-G is working fine, archive command is working fine? I guess this could be used as an additional check if everything is fine, basically a double-check, right?
> 
> Don't get me wrong, I like the feature, but this feature got me a little worried that WAL-G could report all is good, but it actually didn't upload a WAL or something. In general curious how would you use this feature in a production environment.

Actually, cases when holes occur are very rare. Yes, this is essentially a double-check in case of some unexpected Postgres  or storage behavior.

kustodian:
Aha, so we do some PG fix up, or master switchover or something like that
and this will allow us to double check if all is still good. Makes sense.
Thanks 👍

PR #764, opened 2020-09-15
Update PostgreSQL.md

Zyian:
Adds in instructions for MacOS (or unix in general as well) to mention the need to link the dependencies so that we can `go build` the wal-g binary


x4m (approved):
LGTM, thanks!

PR #766, opened 2020-09-21
sqlserver backup database to existing backup

mialinx:
Usecase:

* there is a huge nightly backup with several databases
* we add new empty database and need to backup it immediately, but don't want to backup all existing database

Solution:
* backup new database to existing nightly backup


x4m (approved):
LGTM

Issue #770, opened 2020-09-25
Why custom FastCopy exists?


defntvdm:
BenchmarkCopy uses io.Copy

Benchmarks with race
``` sh
PeCa utility (master!) go test -bench=. -benchmem -race
goos: linux
goarch: amd64
pkg: github.com/wal-g/wal-g/utility
BenchmarkFastCopy-8         	       1	1101841782 ns/op	167775224 B/op	      22 allocs/op
BenchmarkFastCopyBuffer-8   	   25048	     41019 ns/op	   38561 B/op	       1 allocs/op
BenchmarkCopy-8             	 2097900	       501 ns/op	     127 B/op	       1 allocs/op
PASS
ok  	github.com/wal-g/wal-g/utility	71.996s
```

Usual benchmarks
``` sh
PeCa utility (master!) go test -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/wal-g/wal-g/utility
BenchmarkFastCopy-8         	      40	  26411827 ns/op	 7340172 B/op	       1 allocs/op
BenchmarkFastCopyBuffer-8   	 1565972	       713 ns/op	    3637 B/op	       1 allocs/op
BenchmarkCopy-8             	44488784	        27.1 ns/op	      51 B/op	       1 allocs/op
PASS
ok  	github.com/wal-g/wal-g/utility	6.866s
```

``` go
func FastCopyBuffer(dst io.Writer, src io.Reader) (int64, error) {
	buf := copyBytesPool.Get()
	defer copyBytesPool.Put(buf)

	return io.CopyBuffer(dst, src, buf)
}
```

Benchmark function example
``` go
func BenchmarkFastCopyBuffer(b *testing.B) {
	buf := bytes.Repeat([]byte("16 bytes string "), 10<<20) // 160Mib
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			utility.FastCopyNew(ioutil.Discard, bytes.NewReader(buf))
		}
	})
}

```

x4m:
Are these functions equal in specification?

defntvdm:
It looks so, but `io.Copy` and `io.CopyBuffer` can return error when read and write number of bytes is not equal. `io.Copy` creates 32KiB buffer for each call, so I suggest use `io.CopyBuffer` to prevent frequent write syscalls.

defntvdm:
Tested again with command that Pavel Perekalov used and code from test
``` sh
[...]

PR #771, opened 2020-09-26
Use io.CopyBuffer instead of custom realization in FastCopy

defntvdm:
so far only for tests #770


x4m:
Maybe let's merge this?

defntvdm:
Yes, we can. What editorialization you would like me to do?

x4m:
What if we remove FastCopy() entirely? will it be faster?

defntvdm:
I'm not sure. Every io.Copy allocate 32kb buffer. Maybe we want to prevent this behavior and use our Pool of buffers.

x4m (approved):
Yeah, it’s buffered, not fast. But fast is ok too, probably.

PR #783, opened 2020-10-14
wal-g copy for mysql

reshke:


x4m (approved):
I'd be much better to implement history copy too. But as a first step this is fine too. Thnaks!

Issue #779, opened 2020-10-02
Prefetch error: wrong size of prefetched file


vstax:
During restore, wal-g prints errors (or warnings?). I'm not sure if this actually breakfs prefetch or not (normal fetch works fine).

```
ERROR: 2020/10/02 13:21:18.404589 WAL-G: Prefetch error: wrong file size of prefetched file  67108864
2020-10-02 13:21:18.580 MSK::@:[ 213932]: LOG:  restored log file "000000010000003A00000010" from archive
ERROR: 2020/10/02 13:21:18.692997 WAL-G: Prefetch error: wrong file size of prefetched file  67108864
2020-10-02 13:21:18.869 MSK::@:[ 213932]: LOG:  restored log file "000000010000003A00000011" from archive
```

Database (PG12) was created with "--wal-segsize=64" and uses 64MB WAL files. Each one is exactly 67108864 bytes and this is correct size.

Maybe related to #765, but in this case breaks prefetch? Can this size check be disabled somehow? Or maybe allow wal-g getting correct WAL size from postgresql configuration, or at least environment variable?

x4m:
I don't think this breakes prefetch, but it will bread delta backups for sure
I'll appreciate PR which makes our aseertions about WAL size configurable

vstax:
Thanks for looking.

>>> I don't think this breakes prefetch, but it will bread delta backups for sure

I'm not familiar with go, but doesn't https://github.com/wal-g/wal-g/blob/62eb316360c2dfab8b464edcb293210225da488d/internal/wal_fetch_handler.go#L64 mean prefetched file won't be moved to the correct target? So it's all wasted.

There seem to be no issues with making or restoring delta backup.
Making:
```
INFO: 2020/09/30 02:02:01.544791 Delta backup from base_000000010000003900000007_D_00000001000000380000002E with LSN 391c000028. 
INFO: 2020/09/30 02:02:01.563705 Calling pg_start_backup()
WARNING: 2020/09/30 02:02:01.677677 Couldn't get current timeline because of error: 'bytes_per_wal_segment of the server does not match expecte
d value'
INFO: 2020/09/30 02:02:01.677724 Walking ...
INFO: 2020/09/30 02:02:01.678083 Starting part 1 ...
INFO: 2020/09/30 02:02:01.678197 Starting part 2 ...
INFO: 2020/09/30 02:02:01.678367 Starting part 3 ...
[...]

Issue #765, opened 2020-09-17
"bytes_per_wal_segment of the server does not match expected value" warning


vstax:
I'm getting non-fatal warning during full backups:
```
WARNING: 2020/09/17 02:02:01.845203 Couldn't get current timeline because of error: 'bytes_per_wal_segment of the server does not match expected value'
```
Not sure if it affects anything or not?

Database (PG12) was created with "--wal-segsize=64" for larger (64MB) WAL files (https://blog.dbi-services.com/the-wal-segement-size-becomes-changeable-in-postgresql-11/), which allows to have less WAL files under high load.

x4m:
This warning will lead to unability to abort backup made during wal timelime switch.
Basically speaking this disables one safety check.
I think you can ignore it, but if your are going to exploit WAL-G with WAL seg size != 16MB, plz send a PR with configuring this value. I think it's 10-15 loc. + tests (just copy some basic test for initdb with other wal seg size)

PR #807, opened 2020-11-16
Allow the wal segment size to be set as a variable 

Wrhector:
Postgres 11+ allows the wal segment size to be adjusted in initdb, `wal-segsize`. This PR allows you to set the value that Wal-g uses.

Also reported in #765 and #779

I'm not sure `SetWalSize` should be in `pg.go`, let me know if it should be moved.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -52,6 +52,7 @@ const (
 	PgDatabaseSetting            = "PGDATABASE"
 	PgSslModeSetting             = "PGSSLMODE"
 	PgSlotName                   = "WALG_SLOTNAME"
+	PgWalSize                    = "WALG_PG_WAL_SIZE"
 	TotalBgUploadedLimit         = "TOTAL_BG_UPLOADED_LIMIT"
 	NameStreamCreateCmd          = "WALG_STREAM_CREATE_COMMAND"
 	NameStreamRestoreCmd         = "WALG_STREAM_RESTORE_COMMAND"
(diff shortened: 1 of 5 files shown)


x4m:
Hi! Looks super cool, thanks!
There's failed test https://travis-ci.org/github/wal-g/wal-g/jobs/743858326
I suspect that this test failed because one of your tests did not restore default WAL segment size.

internal/config.go
```diff
@@ -52,6 +52,7 @@ const (
 	PgDatabaseSetting            = "PGDATABASE"
 	PgSslModeSetting             = "PGSSLMODE"
 	PgSlotName                   = "WALG_SLOTNAME"
+	PgWalSize                    = "WALG_PG_WAL_SIZE"
```
CyberDem0n, line 55: The actual value MUST be taken from pg_control.

x4m: We cannot read pg_control during wal-push and wal-fetch.
But we can check config agains pg_control during backup-push.

Wrhector: 99% of users won't need to change this value. One change could be changing the error message `bytes_per_wal_segment of the server does not match expected value` to indicate that a user might need to set this value, which will happen on backup-push if the segment size doesn't match.

x4m: Oh, I did not realize that this check is already in place..

x4m: Yes, rewording this would be good. Before this patch user could not do anything, that's why I written it this way...

Wrhector: Committed this change + docs

x4m:
Test 25 seems flaky due to insufficient file limit
I think we can merge this right away

x4m (approved):

x4m:
Pushed, thanks!

Wrhector:
Thanks!

PR #809, opened 2020-11-27
Refactor parsing timelineId from backup name

usernamedt:
Remove code duplication and add a check on empty backup name


reshke (approved):
:ok: :sunglasses:

PR #812, opened 2020-11-30
Do not use TLI in delete history comparisons

x4m:


ants:
I am not sure if this the correct solution, consider also the case when a cluster undergoes a point-in-time-recovery. Then backup from the just recovered database is going to be considered older than the pre-PITR backup.

In theory a similar issue can happen if someone promotes a standby that is way behind. There the correct behavior is even less clear.

HeikoOnnebrink:
@ants 
for the PITR case I agree.. but many people (we included) recover always to a different new server next to the original DB machine with its own new backup bucket ... in these cases it does not matter.
But for an in-place PITR it would cause problem too if the DB after the PITR continues to write to the same backup bucket. 

for the "slave case" I did not get the point. A slave is not sharing the backup bucket with this master (which makes sense as a slave in many cases is a HA component and should have no shared components with his master.. he should ship his own wal to different bucket)
Having this in mind it does not matter how far he is behind his master when he gets promoted.
He makes a timeline switch and continues to ship from here.. I see no conflict ..

In every case this fix already eliminates the most typical use case. But further improvement is also good :-)

x4m:
@ants 's comments are correct. But we really need to prevent problems with upgraded installation losing new backups.

HeikoOnnebrink (approved):
after discussion we have agreed that the fix is not 100% perfect and there are some edge cases not covered. nonetheless it makes sense to apply the change as it fixes a severe bug on pg_upgrade

usernamedt:
I propose to use the backup creation time instead https://github.com/wal-g/wal-g/pull/816

serprex:
@x4m can this PR be closed given #816?

x4m:
Thanks everyone. The root cause issue was settled, let’s close this.

PR #813, opened 2020-11-30
Create go.yml

reshke:


serprex:
From what I see it seems it was created when travis was still in use, since then CI has moved to GitHub Actions & essentially does this. Closing

PR #816, opened 2020-12-01
Fix backup list order in delete command

usernamedt:
In this PR I propose ordering the backups by creation time in the delete command handler.

Backup creation time is taken from backup metadata. If there is no metadata (for example, if we are dealing with an old backup without metadata), the comparator will use the old behavior (timeline and LSN combination) to compare this backup with others.

### DIFF
--- a/cmd/fdb/delete.go
+++ b/cmd/fdb/delete.go
@@ -6,6 +6,7 @@ import (
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
 	"github.com/wal-g/wal-g/utility"
+	"time"
 )
 
 var confirmed = false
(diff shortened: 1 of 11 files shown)


x4m (approved):

usernamedt:
I've redesigned this PR a little bit.

Now user can specify if he wants to use the new logic for backup time fetch or not via --use-sentinel-time flag.
The new logic will apply if --use-sentinel-time flag is specified and all backups in storage contain metadata (i.e. there are no ancient backups without metadata included). If there is some backup without metadata, --use-sentinel-flag would be ignored.

Backup time fetch logic will affect both delete retain and delete before commands (for Postgres). For other databases, nothing has changed.

x4m (approved):

PR #829, opened 2020-12-10
little refactor: move extractor & retries to seperate directory

reshke:


usernamedt:
I propose to deal with only fs-related functions in this PR

usernamedt:
Looks like block_location_* stuff doesn't really belong to fsutil, maybe move it somewhere else?

usernamedt (approved):

Issue #837, opened 2020-12-17
~postgres/env ~postgres/wal-g backup-push "$DATADIR" --rating-composer fails with panic: startReadingFile: failed to open file


banlex73:
Unfortunately, it fails
INFO: 2020/12/17 19:59:32.918513 Starting part 80 ...
panic: startReadingFile: failed to open file '/pgsql/cluster/data/base/2516071/t267_11123082'
: open /pgsql/cluster/data/base/2516071/t267_11123082: no such file or directory

goroutine 439965 [running]:
github.com/wal-g/wal-g/internal.(*RatingTarBallComposer).PackTarballs.func1(0xc0163a9220, 0xc0002e6270, 0x1fe8820, 0xc014f18190)
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/rating_tar_ball_composer.go:186 +0x101
created by github.com/wal-g/wal-g/internal.(*RatingTarBallComposer).PackTarballs
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/rating_tar_ball_composer.go:182 +0x40f
Error running handler: exit status 2

usernamedt:
Hi!

Thanks for reporting the issue. Did you manage to reproduce the error or it occurred only once?

banlex73:
Hi, 
issue is constantly reproducible (last 4 days in a row)
`INFO: 2020/12/18 04:00:32.409950 Starting part 1 ...
panic: startReadingFile: failed to open file '/pgsql/cluster/data/base/1016799/t698_11145705'
: open /pgsql/cluster/data/base/1016799/t698_11145705: no such file or directory

goroutine 163 [running]:
github.com/wal-g/wal-g/internal.(*RatingTarBallComposer).PackTarballs.func1(0xc01394d640, 0xc0022be750, 0x1fe8820, 0xc0002a25a0)
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/rating_tar_ball_composer.go:186 +0x101
created by github.com/wal-g/wal-g/internal.(*RatingTarBallComposer).PackTarballs
        /home/travis/gopath/src/github.com/wal-g/wal-g/internal/rating_tar_ball_composer.go:182 +0x40f
Error running handler: exit status 2```
INFO: 2020/12/19 04:00:24.103370 Querying pg_stat_all_tables
INFO: 2020/12/19 04:00:24.122519 Walking ...
[...]

PR #839, opened 2020-12-19
Fix possible not exist errors during backup-push.

usernamedt:
Correctly handle file not exist error on startReadingFile() call.
Also, replace os.IsNotExist(err) with errors.Is(err, os.ErrNotExist) due to [this issue](https://github.com/golang/go/issues/38198).

This should fix errors like #837


internal/tar_ball_file_packer.go
```diff
@@ -141,10 +133,6 @@ func (p *TarBallFilePacker) createFileReadCloser(cfi *ComposeFileInfo) (io.ReadC
 			return nil, errors.Wrapf(err, "PackFileIntoTar: failed to find corresponding bitmap '%s'\n", cfi.path)
 		}
 		fileReadCloser, cfi.header.Size, err = ReadIncrementalFile(cfi.path, cfi.fileInfo.Size(), *p.incrementFromLsn, bitmap)
-		if os.IsNotExist(err) { // File was deleted before opening
```
x4m, line 144: I believe the file can be deleted between delta construction and opening for a read. Both places are important.

x4m (approved):

PR #847, opened 2021-01-08
Support for Yandex Cloud KMS

luc-lynx:
This PR introduces support for [Yandex Cloud KMS](https://cloud.yandex.com/docs/kms/quickstart/).

Some notes:
- Yandex Cloud SDK doesn't have credentials discovery like AWS SDK, so I implemented an approach when a user either passes a file name with credentials or relies on wal-g attempt to get an access token from virtual machine metadata service. AWS SDK uses similar approach.
- AES-256 is used to encrypt Data Encryption Key.
- In contrast to AWS KMS, Data Encryption Key is generated by Yandex Cloud KMS.
- Tests use a mocked version of key and don't make api calls to Yandex Cloud KMS. AWS KMS support uses the same approach. However in my opinion it's better to come up with tests for AWS KMS and YC KMS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,14 @@ To configure AWS KMS key region for client-side encryption and decryption (i.e.,
 To configure the compression method used for backups. Possible options are: `lz4`, 'lzma', 'brotli'. The default method is `lz4`. LZ4 is the fastest method, but the compression ratio is bad.
 LZMA is way much slower. However, it compresses backups about 6 times better than LZ4. Brotli is a good trade-off between speed and compression ratio, which is about 3 times better than LZ4.
 
+* `YC_CSE_KMS_KEY_ID`
+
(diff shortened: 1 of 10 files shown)


x4m (approved):
Looks great, thanks!

x4m:
Is the operation of obtaining key time-consuming?
I mean we are doing wal-push in PG for each 16Mb file. Will not reading key be a bottleneck?

luc-lynx:
> Is the operation of obtaining key time-consuming?

Just an API call for YC KMS. In case of AWS KMS it also requires one API call. On the KMS side it's just one encryption operation, so it shouldn't be slow.

> I mean we are doing wal-push in PG for each 16Mb file.

Does it mean that wal-g binary is executed on every 16Mb chunk?

x4m:
> Does it mean that wal-g binary is executed on every 16Mb chunk?
Yes

x4m:
Thanks, pushed!

PR #848, opened 2021-01-10
Allow dependabot to keep github actions up-to-date

jlosito:
This change allows dependabot to check any GitHub action which this project uses and submit PRs with version bumps in order to keep packages up-to-date on a weekly basis.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    allow:
+      - dependency-name: "github.com/wal-g/tracelog"
+      - dependency-name: "github.com/wal-g/storages"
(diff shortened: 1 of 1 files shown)


x4m:
Looks cool, thanks!
@g0djan @reshke @usernamedt what do you think?

g0djan (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
usernamedt, line 6:
```suggestion
      interval: "weekly"
  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "weekly"
```

usernamedt: I suggest adding the go modules dependencies check too

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
usernamedt, line 10:
```suggestion
      interval: "weekly"
    allow:
      - dependency-name: "github.com/wal-g/tracelog"
      - dependency-name: "github.com/wal-g/storages"
```

usernamedt: Limit go modules dependency alerts only for wal-g repositories.

usernamedt (approved):

PR #869, opened 2021-02-09
Fix bug with mistyping in comparator

usernamedt:
Looks like I found some mistyping in the comparator, this PR should fix it.


mialinx (approved):

mialinx:
nice catch.

Issue #856, opened 2021-01-26
Backups in Azure Storage, storage account endpoint for environtments different than AzurePublicCloud 


theok-nice:
When setting up wal-g with Azure storage accounts, I noticed that storage account endpoints have been hard-coded to Azure global endpoints.

https://github.com/wal-g/storages/blob/7fdfaf977425d091d0a9b622ef378dbe9f63c268/azure/folder.go#L98

`serviceURL, err = url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net/%s%s", accountName, containerName, accountToken))
`

https://github.com/wal-g/storages/blob/7fdfaf977425d091d0a9b622ef378dbe9f63c268/azure/folder.go#L103
`serviceURL, err = url.Parse(fmt.Sprintf("https://%s.blob.core.windows.net/%s", accountName, containerName))`

Microsoft is also providing services in [Azure Government](https://docs.microsoft.com/en-us/azure/azure-government/compare-azure-government-global-azure), [Azure China](https://docs.microsoft.com/en-us/azure/china/resources-developer-guide) and [Azure Germany](https://docs.microsoft.com/en-us/azure/germany/germany-developer-guide) which use different endpoint for various resources. 

In wal-g case, we care about storage account endpoint

Azure Global: `*.blob.core.windows.net/` 
Azure USGovernment: `*.blob.core.usgovcloudapi.net/`
Azure China: `*.blob.core.chinacloudapi.cn/`
Azure Germany: `*.blob.core.cloudapi.de/`

I would like to suggest that using https://pkg.go.dev/github.com/Azure/go-autorest/autorest/azure, wal-g could work for those environments also.

**One solution could be to change Wal-g config file:** For example could have an additional parameter what would allow user to indicate what environment should be used (PublicCloud,USGovernmentCloud,ChinaCloud,GermanCloud) and wal-g
select the correct endpoint available in https://pkg.go.dev/github.com/Azure/go-autorest/autorest/azure#pkg-variables  

**Another solution could be to have wal-g determine the correct endpoint without changing the config file:**  For example say you have configured wal-g on a VM in Azure China. Using [metadata available on the VM](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service?tabs=linux ) 
`curl -H Metadata:true -s "http://169.254.169.254/metadata/instance?api-version=2020-10-01" | jq -r .compute.azEnvironment` 

[...]

PR #873, opened 2021-02-11
Azure: Add SA endpoints for non AzurePublicCloud

theok-nice:
Addresses issue : [#856](https://github.com/wal-g/wal-g/issues/856)

Changes allow wal-g to work in regions other than AzurePublicCloud.
It adds optional config setting where user can select AzureUSGovernmentCloud, AzureChinaCloud, AzureGermanCloud.
When one of these three options is selected, module will select correct storage account endpoint. For the correct endpoint I am using https://pkg.go.dev/github.com/Azure/go-autorest/autorest/azure module.

If setting is missing from the wal-g config file,or set to any other value (including "AzurePublicCloud") will default to AzurePublicCloud's endpoint.


theok-nice:
I don't understand why these tests fail on code that I haven't change in my commit. Not sure how to resolve this.

x4m (approved):
Thanks!
WAL-G is now able to work with the government cloud. Sounds cool 😎

x4m:
> I don't understand why these tests fail on code that I haven't change in my commit. Not sure how to resolve this.

We added few new checkers recently and still are working to fix all those problems found by these checkers.

kim0:
Thanks @theok-nice .. One typo exists I think .. It says to use `AZURE_STORAGE_SAS_TOKEN`, while I guess the correct one should be `AZURE_ENVIRONMENT_NAME` right ?

PR #875, opened 2021-02-12
sqlserver: backup/restore large databases into multiple blobs

mialinx:


internal/databases/sqlserver/blob/blob.go
```diff
@@ -4,15 +4,23 @@ import (
...
+	"github.com/wal-g/tracelog"
+
 	"github.com/wal-g/storages/storage"
 )
 
 const IndexFileName = "__blob_index.json"
 
+const CacheItemsPerBlob = 3
+
+const MaxCacheBlockSize = 16 * 1024 * 1024 // 10M
```
ostinru, line 22: 10M or 16M ?

mialinx: 16M, fixed

reshke (approved):

PR #898, opened 2021-03-01
Check completition of wal-prefetch more often

x4m:
Currently we check that WAL files grows every 50ms which can cause
unreasonable stalls. Thic commit separates logic of prefetch completition
from timing out.


internal/wal_fetch_handler.go
```diff
@@ -84,19 +86,24 @@ func HandleWALFetch(folder storage.Folder, walFileName string, location string,
 		if runStat, err := os.Stat(running); err == nil {
 			observedSize := runStat.Size() // If there is no progress in 50 ms - start downloading myself
```
usernamedt, line 87: Below 50 ms is replaced with 2 ms, same goes here?

usernamedt (approved):

Issue #865, opened 2021-02-04
[Feature Request] Add additional metadata file to pg WAL 


rjayakri:
##  Issue description

- Currently wal-list provides information about missing wal logs in specific timeline. But does not provide clarity of till what time PITR is possible, since the time of wal generation is missing.

- Often backups are moved to external media or alternate storage locations as time passes (NFS -> tape, or S3 to cheaper longterm storage class). In such scenarios it is important to have access to this metadata so that one can identify specific backups/wals required for a given PITR restore request to specify in the recovery_target_time. 

## Proposed solution

To provide a complete list of files for PITR, wal-push will be enhanced to add  metadata details along with the WAL backup. An additional json file will be created for every WAL and uploaded to the default metadata (WALG_FILE_PREFIX or WALG_S3_PREFIX) store with the following information.

The information is similar to "StartTime" and "FinishTime" of the backup metadata.json (https://github.com/wal-g/wal-g/blob/master/internal/backup_sentinel_dto.go#L32-L33)

File Name : 0000000200000000000000A8.json
Contents: 
{
  "CreatedTime": "2020-10-30T03:24:20.465889208Z",
  "LastWrittenTime": "2020-10-30T03:24:20.465889208Z",
  "DatetimeFormat": "%Y-%m-%dT%H:%M:%S.%fZ"
}

The CreatedTime => access time and LastWrittenTime => modify time of the wal-file generated by Postgres. 
Additional fields like timeline, name, size can be added. But most of it can be derived from the wal-file name.

Happy to help make updates and add tests as requested.

x4m:
Hi! Thanks for your proposal.
Currently, we push WAL-delta files which describe many WAL files at once. Maybe we could use the same approach for WAL metadata?

rjayakri:
@x4m 
Could you please elaborate on the WAL-delta files. Is this the current logic where multiple WAL files are uploaded at once by the archive_command. 
Thanks

rjayakri:
For example , 
following will the files under the wal_005 directory
```
-rw-------  1 user  staff      167 Feb  3 21:37 0000000100000000000000B8.json
-rw-------  1 user  staff  1495286 Feb  3 21:37 0000000100000000000000B8.lz4
-rw-------  1 user  staff      167 Feb  3 21:37 0000000100000000000000B7.json
-rw-------  1 user  staff  1495435 Feb  3 21:37 0000000100000000000000B7.lz4
[...]

PR #899, opened 2021-03-01
[Feature Request] Adding additional metadata file to pg WAL

rjayakri:
Creating PR for the issue reported in https://github.com/wal-g/wal-g/issues/865.


PostgreSQL.md
```diff
@@ -127,6 +127,20 @@ To configure the size of one backup bundle (in bytes). Smaller size causes granu
...
+Sample metadata file (000000020000000300000071.json)
+```
+{
+    "000000020000000300000071": {
+    "created_time": "2021-02-23T00:51:14.195209969Z",
+    "date_fmt": "%Y-%m-%dT%H:%M:%S.%fZ"
+    }
+}
+```
+If not specified it will fallback to default setting(no wal metadata generation)
```
x4m, line 142: Please add an explicit value for  "no metadata". This may be important if environment var overrides config var.

internal/bguploader_test.go
```diff
@@ -176,8 +176,8 @@ func addTestDataFile(t *testing.T, dir string, i int) {
 	}
 }
 
-func testFilename(i int) string {
-	return fmt.Sprintf("B%010d", i)
+func testFilename(i string) string {
+	return fmt.Sprintf("%08d%08d%08s", 1, 1, i)
```
x4m, line 180: Will the new functionality work OK when wal-push is invoked for not WAL-formatted name?
PostgreSQL archived 00000007.history for example.
Or from time to time users archive their stuff :)

internal/wal_push_handler.go
```diff
@@ -2,21 +2,51 @@ package internal
...
+	"strings"
+	"time"
 )
 
 type CantOverwriteWalFileError struct {
 	error
 }
 
+const (
+	WalBulkMetadataLevel       = "BULK"
```
x4m, line 26: Are other our settings capitalized? I'm not sure.

PostgreSQL.md
rjayakri: Added the explicit value "NOMETADATA" and updated the PostgreSQL.md

internal/bguploader_test.go
rjayakri: It picks up any file generated in the directory and creates the metadata file. 
For example
0000000200000007000000C7.partial.json
0000000200000007000000C7.partial.lz4
00000003.history.json
00000003.history.lz4
0000000200000007000000C4.00000028.backup.json
0000000200000007000000C4.00000028.backup.lz4

internal/wal_push_handler.go
rjayakri: I took the example from "WALG_LOG_LEVEL": "NORMAL/DEVEL". Open for any suggestion in-terms of naming covention.

rjayakri:
@x4m Please advice if any more changes required. Thanks.

x4m:
@usernamedt can you please look at this PR too?
I don't have any objections, just don't feel confident enough. You were revisiting metadata stuff recently and, probably, you have a more consistent opinion than mine.

internal/wal_push_handler.go
```diff
@@ -50,6 +80,12 @@ func HandleWALPush(uploader *WalUploader, walFilePath string) {
 	bgUploader.Start()
 	err = uploadWALFile(uploader, walFilePath, bgUploader.preventWalOverwrite)
 	tracelog.ErrorLogger.FatalOnError(err)
+	if err == nil && (viper.IsSet(UploadWalMetadata) || viper.GetString(UploadWalMetadata) == WalNoMetadataLevel) {
```
usernamedt, line 83: I guess here you check instead for `viper.GetString(UploadWalMetadata) != WalNoMetadataLevel `?

Actually, let's just set the default value for UploadWalMetadata to "NOMETADATA" to avoid viper.IsSet check.
This can be done in config.go:108

internal/wal_push_handler_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestWalMetadataIndividualUploader(t *testing.T) {
+	viper.Set(internal.UploadWalMetadata, "INDIVIDUAL")
+	uploader, _, dir, testFileName := generateAndUploadWalFile(t, "1")
+	defer cleanup(t, dir)
+	_, err := uploader.UploadingFolder.ReadObject(testFileName + ".json")
+	assert.NoError(t, err)
+}
+
+func TestWalMetadataBulkUploader(t *testing.T) {
```
usernamedt, line 41: Add test for NOMETADATA setting too?

internal/wal_push_handler.go
```diff
@@ -2,21 +2,51 @@ package internal
...
+const (
+	WalBulkMetadataLevel       = "BULK"
+	WalIndividualMetadataLevel = "INDIVIDUAL"
+	WalNoMetadataLevel         = "NOMETADATA"
+)
+
+var WalMetadataLevels = []string{WalBulkMetadataLevel, WalIndividualMetadataLevel}
+
+type WalMetadataDescription struct {
+	CreateTime     time.Time `json:"created_time"`
```
usernamedt, line 33: createTime or created_time?

internal/wal_push_handler.go
```diff
@@ -2,21 +2,51 @@ package internal
...
 	error
 }
 
+const (
+	WalBulkMetadataLevel       = "BULK"
+	WalIndividualMetadataLevel = "INDIVIDUAL"
+	WalNoMetadataLevel         = "NOMETADATA"
+)
+
+var WalMetadataLevels = []string{WalBulkMetadataLevel, WalIndividualMetadataLevel}
```
usernamedt, line 30: Add WalNoMetadataLevel here?

internal/wal_push_handler.go
```diff
@@ -50,6 +80,12 @@ func HandleWALPush(uploader *WalUploader, walFilePath string) {
 	bgUploader.Start()
 	err = uploadWALFile(uploader, walFilePath, bgUploader.preventWalOverwrite)
 	tracelog.ErrorLogger.FatalOnError(err)
+	if err == nil && (viper.IsSet(UploadWalMetadata) || viper.GetString(UploadWalMetadata) == WalNoMetadataLevel) {
+		// Creating consolidated wal metadata only for bulk option
+		if viper.GetString(UploadWalMetadata) == WalBulkMetadataLevel && walFilePath[len(walFilePath)-1:] == "F" {
```
usernamedt, line 85: Add a comment regarding "F" logic?

internal/wal_push_handler.go
```diff
@@ -59,6 +95,36 @@ func HandleWALPush(uploader *WalUploader, walFilePath string) {
 	}
 } //
 
+func walMetadataBulkUploader(uploader *WalUploader, walFilePath string) {
```
usernamedt, line 98: rename to uploadBulkMetadata?

internal/wal_push_handler.go
```diff
@@ -75,9 +141,46 @@ func uploadWALFile(uploader *WalUploader, walFilePath string, preventWalOverwrit
...
+	err := checkWalMetadataLevel(viper.GetString(UploadWalMetadata))
+	if err != nil {
+		return errors.Wrapf(err, "Incorrect wal metadta level")
+	}
+	fileStat, err := os.Stat(walFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "upload: could not stat wal file'%s'\n", walFilePath)
+	}
+	var walMetadata WalMetadataDescription
+	walMetadataS := make(map[string]WalMetadataDescription)
```
usernamedt, line 164: Rename walMetadataS to something more meaningful?

internal/wal_push_handler.go
```diff
@@ -75,9 +141,46 @@ func uploadWALFile(uploader *WalUploader, walFilePath string, preventWalOverwrit
 		return errors.Wrapf(err, "upload: could not open '%s'\n", walFilePath)
 	}
 	err = uploader.UploadWalFile(walFile)
+	if err == nil && (viper.IsSet(UploadWalMetadata) || viper.GetString(UploadWalMetadata) == WalNoMetadataLevel) {
+		err = uploadWALMetadataFile(uploader, walFilePath)
+		if err != nil {
+			return errors.Wrapf(err, "Failed to upload metadata file")
```
usernamedt, line 147: Should we abort the entire WAL file upload if we failed to upload metadata?

usernamedt (changes requested):
Needs some fixes

internal/wal_push_handler_test.go
rjayakri: Added test for NOMETADATA. Since the s3 mock uploader returns any file present, could not build the proper test cases.

internal/wal_push_handler.go
rjayakri: Changed the parameter name.

internal/wal_push_handler.go
rjayakri: Added. Thanks

internal/wal_push_handler.go
rjayakri: Added comments explaining the process involved.

internal/wal_push_handler.go
rjayakri: Renamed it.

internal/wal_push_handler.go
rjayakri: Renamed it.

internal/wal_push_handler.go
rjayakri: WAL metadata may be critical for PITR and that was the reason aborting with error. If agreed, I can change it to log the error and continue the process.

internal/wal_push_handler_test.go
rjayakri: Updated the unit tests. Now it provides proper test case for NOMETADATA.

usernamedt:
Closed PR accidentaly, sorry :)

Have you tested it with `WALG_UPLOAD_CONCURRENCY` > 1? It seems like the current logic with bundle metadata uploading might not work correctly in this case.

rjayakri:
> Closed PR accidentaly, sorry :)
> 
> Have you tested it with `WALG_UPLOAD_CONCURRENCY` > 1? It seems like the current logic with metadata uploading might not work correctly in this case because of bgUploader.

Thanks. I tested it for WALG_UPLOAD_DISK_CONCURRENCY but missed the WALG_UPLOAD_CONCURRENCY.
Fixed the code for WALG_UPLOAD_CONCURRENCY > 1 and added test case.

internal/wal_push_handler.go
```diff
@@ -58,17 +58,19 @@ func (err CantOverwriteWalFileError) Error() string {
...
 // HandleWALPush is invoked to perform wal-g wal-push
 func HandleWALPush(uploader *WalUploader, walFilePath string) {
+	uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(utility.WalPath)
 	if uploader.ArchiveStatusManager.IsWalAlreadyUploaded(walFilePath) {
 		err := uploader.ArchiveStatusManager.UnmarkWalFile(walFilePath)
 
 		if err != nil {
 			tracelog.ErrorLogger.Printf("unmark wal-g status for %s file failed due following error %+v", walFilePath, err)
 		}
+		if viper.GetString(UploadWalMetadata) == WalBulkMetadataLevel && walFilePath[len(walFilePath)-1:] == "F" {
```
usernamedt, line 68: This looks identical to the line 85.

I think that it might be better to move the whole metadata upload logic to the `WalUploader` and upload the metadata files similarly to WAL delta files.

usernamedt (changes requested):

internal/wal_push_handler.go
usernamedt: Also, what if the WAL segment ending with "F" will be uploaded faster than the previous segments?

rjayakri: > This looks identical to the line 85.
> 
> Maybe extract the common logic to avoid duplication? Or even move the whole metadata upload logic to the `WalUploader` and upload the metadata files similarly to WAL delta files.

Thanks. Just for my understanding, to move the metadata logic to https://github.com/wal-g/wal-g/blob/master/internal/wal_uploader.go#L45 ?
wal_receiver_handler.go also uses the UploadWalFile@wal_uploader.go, should it have any clash if the metadata logic is moved there.

rjayakri: > Also, what if the WAL segment ending with "F" will be uploaded faster than the previous segments? Will the current bulk uploading logic work?

The logic was added here https://github.com/wal-g/wal-g/pull/899/files#diff-f66d6177f62076d018bb01e9b1e9056c991e8206cdb9d16db9e817faf4356abfR68 for that purpose only.
The archive_command(postgresql.conf) will be called for every wal-log, including the one uploaded by the bguploder. Since the archive_command is called in serial, the "F" file will be called at last (after completing the check for all the files created before). So if any of the previous files were not uploaded, it will be handled by the event called from archive_command. Hope my explanation and understanding is clear. Thanks

rjayakri:
Fixed the duplicate code. Happy to make any required changes to move the code to different block.

internal/wal_push_handler.go
usernamedt: > wal_receiver_handler.go also uses the [UploadWalFile@wal_uploader.go](mailto:UploadWalFile@wal_uploader.go), should it have any clash if the metadata logic is moved there.

Nice catch, I think that it might be a good idea to upload the WAL metadata during wal-receive too.

> The archive_command(postgresql.conf) will be called for every wal-log, including the one uploaded by the bguploder. Since the archive_command is called in serial, the "F" file will be called at last (after completing the check for all the files created before). So if any of the previous files were not uploaded, it will be handled by the event called from archive_command. Hope my explanation and understanding is clear. Thanks

When `WALG_UPLOAD_CONCURRENCY` is higher than 1, WAL-G invokes the BgUploader, which looks for WAL files adjacent to the current WAL file and uploads them concurrently to the storage. So, on a single archive_command call there could be multiple WALs being uploaded.

rjayakri: > > wal_receiver_handler.go also uses the [UploadWalFile@wal_uploader.go](mailto:UploadWalFile@wal_uploader.go), should it have any clash if the metadata logic is moved there.
> 
> Nice catch, I think that it might be a good idea to upload the WAL metadata during wal-receive too. So I guess that becomes one more reason to move the metadata uploading to the wal_uploader.go
> 
> > The archive_command(postgresql.conf) will be called for every wal-log, including the one uploaded by the bguploder. Since the archive_command is called in serial, the "F" file will be called at last (after completing the check for all the files created before). So if any of the previous files were not uploaded, it will be handled by the event called from archive_command. Hope my explanation and understanding is clear. Thanks
> 
> When `WALG_UPLOAD_CONCURRENCY` is higher than 1, WAL-G invokes the BgUploader, which looks for WAL files adjacent to the current WAL file and uploads them concurrently to the storage. So, on a single archive_command call there could be multiple WALs being uploaded. So there is no guarantee that the "F" file will be uploaded exactly after the previous WAL files.

Thanks. I will move the logic to wal_uploader.go so that wal-push and wal-receive can use the metadata logic.

usernamedt:
```
INFO: 2021/03/19 13:25:18.372627 FILE PATH: 00000001000000030000004D.lz4
ERROR: 2021/03/19 13:25:18.372858 upload: could not Upload 'pg_wal/00000001000000030000004D'
: upload: could not stat wal file'/tmp/00000001000000030000004D'
: stat /tmp/00000001000000030000004D: no such file or directory
2021-03-19 13:25:18.379 +05 [9202] LOG:  archive command failed with exit code 1
2021-03-19 13:25:18.379 +05 [9202] DETAIL:  The failed archive command was: /usr/local/bin/wal-g wal-push pg_wal/00000001000000030000004D --config /Users/usernamedt/wal-g.yaml
INFO: 2021/03/19 13:25:20.761373 FILE PATH: 00000001000000030000004F.lz4
ERROR: 2021/03/19 13:25:20.761720 upload: could not Upload 'pg_wal/00000001000000030000004F'
: upload: could not stat wal file'/tmp/00000001000000030000004F'
: stat /tmp/00000001000000030000004F: no such file or directory
2021-03-19 13:25:20.765 +05 [9202] LOG:  archive command failed with exit code 1
2021-03-19 13:25:20.765 +05 [9202] DETAIL:  The failed archive command was: /usr/local/bin/wal-g wal-push pg_wal/00000001000000030000004F --config /Users/usernamedt/wal-g.yaml
```
Currently the archive command is failing, needs fix

internal/wal_receive_handler.go
```diff
@@ -105,6 +107,7 @@ func HandleWALReceive(uploader *WalUploader) {
 			tracelog.ErrorLogger.FatalOnError(err)
 			err = uploader.UploadWalFile(ioextensions.NewNamedReaderImpl(tlh, tlh.Name()))
 			tracelog.ErrorLogger.FatalOnError(err)
+			uploadBulkMetadata(uploader, filepath.Join(getWalFolderPath(), segment.Name()))
```
usernamedt, line 110: Why not move this to the wal_uploader.go too?

rjayakri:
> ```
> INFO: 2021/03/19 13:25:18.372627 FILE PATH: 00000001000000030000004D.lz4
> ERROR: 2021/03/19 13:25:18.372858 upload: could not Upload 'pg_wal/00000001000000030000004D'
> : upload: could not stat wal file'/tmp/00000001000000030000004D'
> : stat /tmp/00000001000000030000004D: no such file or directory
> 2021-03-19 13:25:18.379 +05 [9202] LOG:  archive command failed with exit code 1
> 2021-03-19 13:25:18.379 +05 [9202] DETAIL:  The failed archive command was: /usr/local/bin/wal-g wal-push pg_wal/00000001000000030000004D --config /Users/usernamedt/wal-g.yaml
> INFO: 2021/03/19 13:25:20.761373 FILE PATH: 00000001000000030000004F.lz4
> ERROR: 2021/03/19 13:25:20.761720 upload: could not Upload 'pg_wal/00000001000000030000004F'
> : upload: could not stat wal file'/tmp/00000001000000030000004F'
> : stat /tmp/00000001000000030000004F: no such file or directory
> 2021-03-19 13:25:20.765 +05 [9202] LOG:  archive command failed with exit code 1
> 2021-03-19 13:25:20.765 +05 [9202] DETAIL:  The failed archive command was: /usr/local/bin/wal-g wal-push pg_wal/00000001000000030000004F --config /Users/usernamedt/wal-g.yaml
> ```
> 
> Currently, the archive command is failing, needs a fix

My apologies. I had PGDATA set in the environment and that was the reason it was working. Will fix it soon

rjayakri:
Need some clarification/suggestion on the wal_receiver. Currently implemented logic for wal_receive to create metadata based on the file timestamp on pg_wal/<wal_file_name>. But  wal_receive can run from any other system and may not have access to the pg_wal directory to identify the file stats (to create metadatafile). Can I use the system timestamp (at the time of wal-receive ) instead of checking the wa-file timestamp on pg_wal directory.

usernamedt:
> Need some clarification/suggestion on the wal_receiver. Currently implemented logic for wal_receive to create metadata based on the file timestamp on pg_wal/<wal_file_name>. But wal_receive can run from any other system and may not have access to the pg_wal directory to identify the file stats (to create metadatafile). Can I use the system timestamp (at the time of wal-receive ) instead of checking the wa-file timestamp on pg_wal directory.

Since replication is synchronous, the network lag won't affect the monotonic increase of CreatedTime but will add some delay. From my point of view, it is OK. @x4m what do you think?

x4m:
> > Need some clarification/suggestion on the wal_receiver. Currently implemented logic for wal_receive to create metadata based on the file timestamp on pg_wal/<wal_file_name>. But wal_receive can run from any other system and may not have access to the pg_wal directory to identify the file stats (to create metadatafile). Can I use the system timestamp (at the time of wal-receive ) instead of checking the wa-file timestamp on pg_wal directory.
> 
> Since replication is synchronous, the network lag won't affect the monotonic increase of CreatedTime but will add some delay. From my point of view, using the system timestamp is OK. @x4m what do you think?

Sure, anyway this CreatedTime does not affect how WALs will be applied by PostgreSQL. As long as we delete these files basing judgment on LSN - CeatedTime safely can be a random number.

rjayakri:
> > > Need some clarification/suggestion on the wal_receiver. Currently implemented logic for wal_receive to create metadata based on the file timestamp on pg_wal/<wal_file_name>. But wal_receive can run from any other system and may not have access to the pg_wal directory to identify the file stats (to create metadatafile). Can I use the system timestamp (at the time of wal-receive ) instead of checking the wa-file timestamp on pg_wal directory.
> > 
> > 
> > Since replication is synchronous, the network lag won't affect the monotonic increase of CreatedTime but will add some delay. From my point of view, using the system timestamp is OK. @x4m what do you think?
> 
> Sure, anyway this CreatedTime does not affect how WALs will be applied by PostgreSQL. As long as we delete these files basing judgment on LSN - CeatedTime safely can be a random number.

Thanks for the feedback. Will implement the logic for wal-receive

internal/wal_receive_handler.go
rjayakri: Moved that logic to wal_uploader.go for the wal-receive command. Thanks

internal/configure.go
```diff
@@ -156,6 +156,21 @@ func getWalFolderPath() string {
...
+	if _, err := os.Stat(dataFolderPath); err == nil {
+		return dataFolderPath
+	}
+
+	dataFolderPath = filepath.Join("pg_xlog")
+	if _, err := os.Stat(dataFolderPath); err == nil {
+		return dataFolderPath
+	}
+
+	return DefaultDataFolderPath
```
usernamedt, line 171: let's use this function in the above getWalFolderPath() too

internal/wal_push_handler.go
```diff
@@ -50,7 +50,7 @@ func HandleWALPush(uploader *WalUploader, walFilePath string) {
 	bgUploader.Start()
 	err = uploadWALFile(uploader, walFilePath, bgUploader.preventWalOverwrite)
 	tracelog.ErrorLogger.FatalOnError(err)
-	uploadBulkMetadata(uploader, walFilePath)
+	_ = uploadBulkMetadata(uploader, filepath.Join(getRelativeArchiveDataFolderPath(), filepath.Base(walFilePath)))
```
usernamedt, line 53: Why ignore the error here?

internal/wal_uploader.go
```diff
@@ -168,12 +208,13 @@ func uploadBulkMetadata(uploader *WalUploader, walFilePath string) {
 		}
 	}
 	dtoBody, _ := json.Marshal(walMetadataArray)
```
usernamedt, line 210: Why ignore the error here and above?
```
file, _ := ioutil.ReadFile(walMetadataFile)
walMetadataFiles, _ := filepath.Glob(walMetadataBulkFolder.GetFilePath("")...

```

usernamedt (changes requested):

internal/configure.go
rjayakri: Thanks. Simplified the function call

internal/wal_uploader.go
rjayakri: There was still open discussion about  whether we need to fail if the wal-metadata upload fails for any reason (missing file/upload failure etc). I will change all the function calls based on the decision.

usernamedt:
I've slightly refactored the code:
- moved all metadata related logic to the wal_metadata_uploader.go
- removed WALG_RECEIVE_BULKMETADATA_PATH. wal-g already has the logic (in getRelativeArchiveDataFolderPath) to fall back to using /tmp folder when pg_wal/pg_xlog is not available so I think we can utilize it. 
- changed the TestWalPush_HandleWALPush to test if the WAL file has been successfully uploaded to the mock storage

usernamedt:
@rjayakri If I understood correctly, you plan to add the WAL metadata info to the `wal-show` command? If so, I think we can merge this PR now and add some integration tests later.

usernamedt (approved):

Issue #904, opened 2021-03-07
pg_wal/.wal-g breaks pg_rewind


thequailman:
Hello, I'm running wal-g with PostgreSQL 13.  When I run `pg_rewind -R -D /var/lib/postgresql/13/main --source-server=postgresql://myserver/postgres --restore-target-wal` with wal-g as the archive_command and restore_command, I get a fatal error from pg_rewind:

```
pg_rewind: servers diverged at WAL location 18C/35000000 on timeline 4
pg_rewind: rewinding from last common checkpoint at 18C/34000060 on timeline 4
pg_rewind: fatal: could not remove file "/var/lib/postgresql/13/main/pg_wal/.wal-g/prefetch/running/000000050000018C00000036": No such file or directory
```

This also has the fun side effect of breaking the underlying cluster (no pg_control) and creating a new timeline in wal-g.

This issue was reported on the PostgreSQL mailing list (https://www.postgresql-archive.org/Concurrency-issue-in-pg-rewind-td6154319.html) but I don't think it was ever resolved.  Basically, wal-g shouldn't be creating folders under pg_wal was what the maintainers recommended?  Could you make the .wal-g prefetch directory configurable?

x4m:
Thanks for reminding me. I've totally forgotten about it...

debebantur:
I believe, above prs solve this issue. Thanks for reporting this, if there are any concerns left, feel free to reopen issue

PR #905, opened 2021-03-07
Invent WALG_PREFETCH_DIR

x4m:
This will help to resolve issue #904

@CyberDem0n please take a look. Does it help with #904?


x4m:
Uhm....soooooo hacky...

CyberDem0n:
> Uhm....soooooo hacky...

Yes, it is, but also it reducing chances of user error down to absolute zero.

reshke (approved):

PR #909, opened 2021-03-16
Try to predict files in archive_status before listing directory

x4m:


internal/bguploader.go
```diff
@@ -109,6 +114,24 @@ func (b *BgUploader) scanAndProcessFiles() {
 	defer close(fileChan)
 	go b.processFiles(fileChan)
 
+	walName := b.firstWalName
+
+	for i := int32(0); i < b.maxNumUploaded ; i++ {
+		walName, err := GetNextWalFilename(walName)
```
usernamedt, line 120: walName always equals b.firstWalName, is it an expected behaviour?

usernamedt (approved):

PR #858, opened 2021-01-31
[Feature Request 857] Add Catchup-List command proposal draft

genslein:
Adds the `catchup-list` CLI command from #857

- Refactors existing `backup.go` internal code to allow path specification
- Backward compatible to allow the `pg` code to mature while leaving other shared paths in tact

Happy to help make updates  and  add tests as requested.

PR #912, opened 2021-03-28
[857 Request] Refactor  changes to unify methods with utility paths as params

genslein:
https://github.com/wal-g/wal-g/pull/858

This was the follow-up refactor to merge the code paths.  Will make additional changes as requested here.

### DIFF
--- a/cmd/fdb/backup_list.go
+++ b/cmd/fdb/backup_list.go
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/utility"
 )
 
 const backupListShortDescription = "Prints available backups"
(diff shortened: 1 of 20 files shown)


cmd/fdb/backup_list.go
```diff
@@ -16,7 +17,7 @@ var backupListCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		folder, err := internal.ConfigureFolder()
 		tracelog.ErrorLogger.FatalOnError(err)
-		internal.DefaultHandleBackupList(folder)
+		internal.DefaultHandleBackupList(folder, utility.BaseBackupPath)
```
usernamedt, line 20: Maybe I missed that point, but why we can't instead call this function like this?
```
internal.DefaultHandleBackupList(folder.GetSubFolder(utility.BaseBackupPath))
```

genslein: Ah fair point. I'll update

usernamedt:
Looks like the `bin/golangci-lint` is uploaded unintentionally

usernamedt (approved):
lgtm, remove the binary file

genslein:
> lgtm, remove the binary file

Ah, I thought I ignored it. Removed

PR #913, opened 2021-03-29
Delete target backup

usernamedt:
Add functionality to delete target backup for Postgres

User can specify the target for deletion via backup name or user data:
```
wal-g delete target base_0000000100000000000000C4 // delete base backup specified by name
wal-g delete target --target-user-data "{ \"x\": [3], \"y\": 4 }"	//delete backup specified by user data
```
When deleting full backup, all dependant increment backups will be deleted too.

Deleting the incremental backup is disabled by default, but can be enabled via `--find-full` flag:
```
wal-g delete target base_0000000100000000000000C9_D_0000000100000000000000C4 --find-full
```
In this case, wal-g will search for the base backup and will delete the base backup and all dependant increments.


g0djan (approved):
LGTM

PR #917, opened 2021-03-31
Delete target backup (improvements)

usernamedt:
In this PR I propose several improvements to the previously introduced (#913) delete target backup command.

1. Rename the `--find-full` flag to `FIND_FULL` to match the other delete commands.

2. Implement delta backups deletion. If delta backup is provided as the target, WAL-G will also delete all the dependant delta backups. Behaviour of the previously introduced `FIND_FULL` is preserved. If `FIND_FULL` is specified, WAL-G will delete all backups with the same base backup as the target.

Examples:
``target base_0000000100000000000000C9`` delete the base backup and all dependant delta backups

``target base_0000000100000000000000C9_D_0000000100000000000000C4``	delete delta backup and all dependant delta backups

``target FIND_FULL base_0000000100000000000000C9_D_0000000100000000000000C4`` delete delta backup and all delta backups with the same base backup

3. Add docs and tests


munakoiso (approved):
👍

PR #926, opened 2021-04-07
Enqueue the tarball back when PackFileIntoTar errors

JelteF:
If it's not put back then another AddFile or AddHeader will block
indefinitely on Deque. This can cause the program to deadlock and the
user to never see the error that PackFileIntoTar returned.

### DIFF
--- a/internal/regular_tar_ball_composer.go
+++ b/internal/regular_tar_ball_composer.go
@@ -56,6 +56,9 @@ func (c *RegularTarBallComposer) AddFile(info *ComposeFileInfo) {
 	c.errorGroup.Go(func() error {
 		err := c.tarFilePacker.PackFileIntoTar(info, tarBall)
 		if err != nil {
+			// put it back otherwise another Dequeu will block and the user
+			// might never see the error because of a deadlock there.
+			c.tarBallQueue.EnqueueBack(tarBall)
 			return err
 		}
 		return c.tarBallQueue.CheckSizeAndEnqueueBack(tarBall)


internal/regular_tar_ball_composer.go
```diff
@@ -56,6 +56,9 @@ func (c *RegularTarBallComposer) AddFile(info *ComposeFileInfo) {
 	c.errorGroup.Go(func() error {
 		err := c.tarFilePacker.PackFileIntoTar(info, tarBall)
 		if err != nil {
+			// put it back otherwise another Dequeu will block and the user
+			// might never see the error because of a deadlock there.
+			c.tarBallQueue.EnqueueBack(tarBall)
 			return err
 		}
 		return c.tarBallQueue.CheckSizeAndEnqueueBack(tarBall)
```
usernamedt, line 64: Hi, thanks for this catch!

I just realized that CheckSizeAndEnqueueBack may return the error too and not enqueue the tarBall back and which will cause a similar deadlock, so this case needs a similar fix.

But I think that instead of ensuring that we always put the tarball back, it might be simpler to use the context (which we ignore as of now), so I propose to replace this 
```
	errorGroup, _ := errgroup.WithContext(context.Background())
```
with
```
	errorGroup, ctx := errgroup.WithContext(context.Background())
```
 and then check for `ctx.Canceled` before `c.tarBallQueue.Deque()` calls. What do you think?

JelteF: I think using the context sounds like a good solution too, but I don't think checking `ctx.Canceled` before `Dequeu` is enough. The error might occur after the `ctx.Canceled`, so you would miss it like that and still get the deadlock. Instead you would need to use a `select` inside `Deque`. Something like this:
```
select {
case <-ctx.Done()
     return nil, ctx.Err()
case tarball := <-tarQueue.tarsToFillQueue
     return tarball, nil
}
```

usernamedt: Agreed

JelteF:
Closing this in favor of #928

PR #928, opened 2021-04-09
Use errorGroup ctx in Deque for RegularTarBallComposer

JelteF:
Otherwise errors inside the errorGroup could cause a deadlock, because
its errorGroup go routines only enqueue the TarBall back in their happy
path.

When making a similar change for RatingTargBallComposer unittests
started failing, so that's left out of this change.

### DIFF
--- a/internal/tar_ball_queue.go
+++ b/internal/tar_ball_queue.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"context"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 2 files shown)


internal/tar_ball_queue.go
```diff
@@ -57,11 +58,22 @@ func (tarQueue *TarBallQueue) StartQueue() error {
...
+// DequeCtx returns a TarBall from the queue. If the context finishes before it
+// can do so, then instead it returns nil.
+func (tarQueue *TarBallQueue) DequeCtx(ctx context.Context) TarBall {
 	if tarQueue.started.IsNotSet() {
 		panic("Trying to deque from not started Queue")
 	}
-	return <-tarQueue.tarsToFillQueue
+	select {
+	case <-ctx.Done():
+		return nil
```
usernamedt, line 69: Last thing:
In case of some weird situation when there is a nil tarball in the tarQueue.tarsToFillQueue, it might be hard to distinguish it from canceled context.

So I suggest returning an explicit error instead:
```
return nil, ctx.Err()
```

usernamedt (commented):
> When making a similar change for RatingTargBallComposer unittests
> started failing, so that's left out of this change.

I've inspected the RatingTarBallComposer code and it looks like it is not affected by this issue so we can leave it unchanged.

internal/tar_ball_queue.go
JelteF: Done

usernamedt (approved):
lgtm

PR #930, opened 2021-04-13
Extract pg related code to databases/postgres

usernamedt:
I moved all Postgres-related code from internal to internal/databases/postgres and left only common code in internal.

Also, I introduced the GenericMetadata so WAL-G can use the common logic for different databases. 

To support the common logic for chosen database engine, one should write its own GenericMetadataInteractor


munakoiso (approved):
👍

PR #931, opened 2021-04-15
Updating pgproto3 hoping to fix client_encoding issue

sebasmannem:


sebasmannem:
Only MySQL test failed, so we can merge this one...

docker/pg_tests/scripts/tests/test_functions/backup_and_restore_test.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/sh
```
usernamedt, line 1: This seems like the script from the remote backup patch :)

usernamedt (changes requested):

docker/pg_tests/scripts/tests/test_functions/backup_and_restore_test.sh
sebasmannem: Ouch. I removed it.

usernamedt (approved):

PR #944, opened 2021-04-27
Added few usefull comments in 'internal/*'

proggga:


internal/backup.go
```diff
@@ -1,5 +1,7 @@
 package internal
 
+// 'Sentinel' mean a 'file' with meta/external-info about backup (in some cases)
```
usernamedt, line 3: Move this comment to `backup_sentinel_dto.go`?

usernamedt, line 3: Move this to `backup_sentinel_dto.go`?

usernamedt: We can add something more detailed:

WAL-G stores information about single backup in the following files:
1. Sentinel file - contains useful metadata, such as backup start time, backup size, etc.
2. (only in Postgres) Metadata file  - Postgres sentinel files can be quite large (> 1GB), so the metadata file is useful for the quick fetch of backup-related information.

proggga: added this description to backup.go, fixed

usernamedt (approved):

PR #955, opened 2021-04-30
Fix wrong logging in mysql/backup_push, when err == nil

proggga:
Found in logs that we write error even we have 'nil' error, this PR fix this behavior

### DIFF
--- a/internal/databases/mysql/backup_push_handler.go
+++ b/internal/databases/mysql/backup_push_handler.go
@@ -1,7 +1,6 @@
 package mysql
 
 import (
-	"fmt"
 	"os"
 	"os/exec"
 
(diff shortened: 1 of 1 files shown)


internal/databases/mysql/backup_push_handler.go
```diff
@@ -42,10 +42,14 @@ func HandleBackupPush(uploader *internal.Uploader, backupCmd *exec.Cmd, isPerman
...
 	uploadedSize, err := uploader.UploadedDataSize()
-	tracelog.ErrorLogger.PrintOnError(fmt.Errorf("failed to calc uploaded data size: %v", err))
+	if err != nil {
+		tracelog.ErrorLogger.PrintError(fmt.Errorf("failed to calc uploaded data size: %v", err))
+	}
 
 	rawSize, err := uploader.RawDataSize()
-	tracelog.ErrorLogger.PrintOnError(fmt.Errorf("failed to calc raw data size: %v", err))
+	if err != nil {
+		tracelog.ErrorLogger.PrintError(fmt.Errorf("failed to calc raw data size: %v", err))
```
usernamedt, line 51: Usually, it is simplier to do this:
```
	tracelog.ErrorLogger.Printf("Failed to calc uploaded data size: %v", err)
```

proggga: fixed

usernamedt (approved):

PR #960, opened 2021-05-03
using live replay binlogs by default.

mialinx:
until timestamp may be used as deadline for live replay


internal/databases/mysql/mysql.go
```diff
@@ -193,7 +194,7 @@ outer:
 				break outer
 			}
 		}
-		if len(logsToFetch) == 0 || !liveReplay {
+		if len(logsToFetch) == 0 || now.After(endTS) {
```
ostinru, line 197: There are 2 condition to exit outer loop:
1) no new files = `len(logsToFetch) == 0`
OR
2) binlog timestamp is after endTS

Second check is on line 193, so I believe that we can remove this `now.After(endTS)` check

mialinx: fixed

ostinru:
👍

ostinru (approved):
👍

PR #961, opened 2021-05-04
Split settings: binary now ask db specific settings, not all settings

proggga:
Problem:
We have different binaries for every db, but that binary provider config options for all databases. That's not really good.

So I split the config for each binary. Now every certain binary has common options + db specific options.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -16,6 +16,13 @@ import (
 )
 
 const (
+	PG        = "PG"
+	SQLSERVER = "SQLSERVER"
+	MYSQL     = "MYSQL"
+	REDIS     = "REDIS"
+	FDB       = "FDB"
+	MONGO     = "MONGO"
+
 	DownloadConcurrencySetting   = "WALG_DOWNLOAD_CONCURRENCY"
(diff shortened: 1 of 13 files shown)


internal/config.go
```diff
@@ -250,6 +253,31 @@ var (
...
+		// GOLANG
+		GoMaxProcs: true,
+
+		// Web server
+		HTTPListen:       true,
+		HTTPExposePprof:  true,
+		HTTPExposeExpVar: true,
+	}
+
+	PGAllowedSettings = map[string]bool{
```
usernamedt, line 265: Can we extract these db specific settings to the corresponding packages?

proggga: I've tried to play with moving configuration to db specific files, and it looks ugly. It requires moving all constants from config, all variables to db files, add extra functions. All this increasing mess with that code. Right now this code  looks minimal, placed in one file, easy to maintain.

usernamedt: Ok, then I guess its fine as it is now

usernamedt (approved):

PR #962, opened 2021-05-04
Reduce race window in WAL prefetch

x4m:


ostinru (approved):
Looks good

internal/fetch_helper.go
```diff
@@ -172,10 +173,17 @@ func DownloadAndDecompressStorageFile(folder storage.Folder, fileName string) (i
...
 // DownloadFileTo downloads a file and writes it to local file
 func DownloadFileTo(folder storage.Folder, fileName string, dstPath string) error {
+	file, err := os.OpenFile(dstPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC|os.O_EXCL, 0666)
+	if err != nil {
+		return err
+	}
+
 	reader, err := DownloadAndDecompressStorageFile(folder, fileName)
 	if err != nil {
 		return err
```
proggga, line 183: if an error occurs, we will abandon an empty file - not really good, we should remove it

internal/fetch_helper.go
```diff
@@ -172,10 +173,17 @@ func DownloadAndDecompressStorageFile(folder storage.Folder, fileName string) (i
...
 	}
 	defer utility.LoggedClose(reader, "")
-	return ioextensions.CreateFileWith(dstPath, reader)
+
+	_, err = utility.FastCopy(file, reader)
```
proggga, line 187: and here

Issue #850, opened 2021-01-16
Need script for automatic downloading missing WAL's from cloud storage (our example in S3) on previous master after switching for pg_rewind


roveil:
Hello!
Thank you for great module.
After switching postgres master node and trying to syncronize previous master with current, we have missing WAL's during `pg_rewind` processing on previous master and manually download them from AWS S3 using `wal-g wal-fetch`
We can write our own script, that parse output of` pg_rewing` and will download missing WAL's on cycle, but i suppose, if that script will be the part of your module, it will be fine :)

x4m:
Hi!

Just use --restore-target-wal https://www.postgresql.org/docs/13/app-pgrewind.html

I'm proud to be a reviewer of this feature :) 
https://github.com/postgres/postgres/commit/a7e8ece41cf7a96d8a9c4c037cdfef304d950831

If you are on 12 I can send you a backpatch.

On earlier versions, probably we could have a script, but I do not have it.

x4m:
I think it's really a neat idea to have ```wal-g restore-target-wal``` command.
I'd be happy to review a PR with it if someone is inclined to implement it.

roveil:
This is the most relevant for old databases where `pg_rewind` does not support `--restore-target-wal` option.

x4m:
I'm not sure, but probably we have facilities to parse pg_control file in WAL-G code. So must be straightforward to implement the feature, by extracting LSNs.
But it's definitely huge chunk of work.

usernamedt:
Hi, WAL-G now has the `wal-restore` command that should solve this issue. As for now, it supports only the local source and target setup, but we plan to implement the local source and remote target setup support soon.

PR #977, opened 2021-05-15
WIP wal-restore

Xaspy:
That what I think:

1) I'm scary what I don't handle this case: we need wal segments that names already exists in local wal directory and need to rewrite them (Can this happen?). Because in current logic if we have wal segment with some name we can't download this wal file from external storage.

2) This right that I will download missed segments instantly in local wal directory? I think this dangerous :)

3) And I want to know how in one directory named 'internal' we have two packages: 'internal' and 'internal_test'.

4) And yes, my code should be covered by tests. But before this I want know what I have right architecture of the utility.

UPD: 

5) I think I should add checks that two clusters have one parent: check history wal files and then, if I find difference get all wal segments on external storage after this difference.

UPD2:

6) Should I do some checks which performs in pg_rewind?


x4m:
Hi! Many WAL-G developers, like me and you, speak the Russian language. But we do not want knowledge of Russian to be a prerequisite to participate in WAL-G discussions. In fact, every community member matters. Please, use English in Github discussions.

Xaspy:
Hello! I updated my first comment.

Xaspy:
And question:
1) Should I delete existing files with the same name with missed files? I think yes, because these files can be not for external cluster. But if user later will don't pg_rewind -> existing WAL files will be in inconsistent state.

usernamedt:
> That what I think:
> 
> 1. I'm scary what I don't handle this case: we need wal segments that names already exists in local wal directory and need to rewrite them (Can this happen?). Because in current logic if we have wal segment with some name we can't download this wal file from external storage.

I think that we just can download the WAL files that are missing on the current master (pg_rewind target) up to the segment number of divergence.
 
> 2. This right that I will download missed segments instantly in local wal directory? I think this dangerous :)

If you don't overwrite any existing files, why not?

> 3. And I want to know how in one directory named 'internal' we have two packages: 'internal' and 'internal_test'.

`*_test` packages contain tests, for example in `internal_test` we have tests for `internal` package.

> 1. I think I should add checks that two clusters have one parent: check history wal files and then, if I find difference get all wal segments on external storage after this difference.

I think comparing the `system identifier` of the clusters should be enough

internal/wal_restore_handler.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Because we need sorted slice in descending order
+func (p WalSegmentNumbers) Less(i, j int) bool { return p[i] > p[j] }
+func (p WalSegmentNumbers) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+// HandleWALRestore is invoked to perform wal-g wal-restore
+func HandleWALRestore(externalFolder storage.Folder, localFolder storage.Folder) {
+	externalHistoryRecs, err := getTimeLineHistoryRecords(1, externalFolder)
+	tracelog.ErrorLogger.FatalfOnError("Failed to get the external WAL history records %v\n", err)
+	localHistoryRecs, err := getTimeLineHistoryRecords(1, localFolder)
```
usernamedt, line 35: Actually I am not sure if we can trust the history record file to get the current timeline and LSN, Postgres stores this information in the `pg_control` file and I think it should be preffered to get the precise information.

cmd/pg/wal_restore.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// walRestoreCmd represents the walRestore command
+var walRestoreCmd = &cobra.Command{
+	Use:   WalRestoreUsage,
+	Short: WalRestoreShortDescription,
+	Long:  WalRestoreLongDescription,
+	Args:  cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		targetFolder, err := fs.ConfigureFolder(args[0], nil)
```
usernamedt, line 23: `storage.Folder` interface is used to interact with the backups/wal archive storage (e.g. S3, Azure, etc). As for the local filesystem operations, you need to use the `os` package.

usernamedt (changes requested):

docker-compose.yml
```diff
@@ -83,6 +83,7 @@ services:
       &&  mkdir -p /export/gpdeletebeforenamebucket
       &&  mkdir -p /export/gpdeletebeforetimebucket
       &&  mkdir -p /export/storagetoolsbucket
+      &&  mkdir -p /export/restorebucket
```
usernamedt, line 86: `restorebucket` name is too ambigious, maybe change it to the `walrestorebucket`? Same for the name of the tests and dockerfiles

docker/pg_tests/scripts/tests/restore_test.sh
```diff
@@ -0,0 +1,124 @@
...
+echo "host  replication  repl              127.0.0.1/32  md5" >> pg_hba.conf
+echo "wal_level = replica" >> postgresql.conf
+echo "wal_keep_segments = 100" >> postgresql.conf
+echo "max_wal_senders = 2" >> postgresql.conf
+echo "hot_standby = on" >> postgresql.conf
+echo "listen_addresses = 'localhost'" >> postgresql.conf
+
+echo "archive_mode = on" >> postgresql.conf
+echo "archive_command = '\
+AWS_ACCESS_KEY_ID=AKIAIOSFODNN7EXAMPLE \
```
usernamedt, line 33: I guess there is no need to duplicate the settings before the each wal-g call. Instead, add them to the `restore_test_config.json` and pass the wal-g config using the `--config` flag (see the other tests for examples).

docker/pg_tests/scripts/tests/restore_test.sh
```diff
@@ -0,0 +1,124 @@
...
+pgbench -i -s 10 -h 127.0.0.1 -p ${BETA_PORT} postgres
+
+/usr/lib/postgresql/10/bin/pg_ctl -D ${PGDATA_BETA} -m fast -W stop
+sleep 3
+
+# for more info to log
+ls "${PGDATA_BETA}/pg_wal"
+
+WAL_TO_DELETE_NAME="00000002000000000000000C"
+DELETED_WAL="${PGDATA_BETA}/pg_wal/${WAL_TO_DELETE_NAME}"
```
usernamedt, line 110: The main scenario for pg_rewind use case is the following:
1. Old master fail / become unavailable
2. Replica become promoted to the new master
3. pg_rewind is called to repair the old master history so it can became the replica of the new master.

So I propose to add the `pg_rewind` call at the end of the test to make sure it succeeds. Also, I propose to not manually remove the WALs from the new master, but:
1. set `max_wal_size/wal_keep_segments` to some small value
2. call a CHECKPOINT, write some transactions until (enough for old WAL to be removed)
3. call wal-restore
4. call pg_rewind to make sure it'll work okay

internal/databases/postgres/wal_restore_handler.go
```diff
@@ -0,0 +1,242 @@
...
+func NewTimelineWithSegmentNo(tl uint32, seg uint64) *TimelineWithSegmentNo {
+	return &TimelineWithSegmentNo{timeline: tl, segmentNo: seg}
+}
+
+func NewTimelineWithSegmentNoBy(record *TimelineHistoryRecord) *TimelineWithSegmentNo {
+	return NewTimelineWithSegmentNo(record.timeline, getSegmentNoFromLsn(record.lsn))
+}
+
+func getSegmentNoFromLsn(lsn uint64) uint64 {
+	return lsn / WalSegmentSize
```
usernamedt, line 33: Hm, I guess that this logic may exist somewhere in WAL-G already, are you sure it is not a duplicate?

usernamedt (changes requested):

docker/pg_tests/scripts/tests/restore_test.sh
Xaspy: Fixed.

internal/databases/postgres/wal_restore_handler.go
Xaspy: This logic used in WalSegmentNo. Moved this function.

internal/databases/postgres/wal_restore_handler.go
```diff
@@ -0,0 +1,237 @@
...
+	mapOfTargetTimelineWithSegNo := timelineWithSegmentNoSliceToMap(targetTimelineWithSegNo)
+
+	folderFilenames, err := getDirectoryFilenames(targetWalDir)
+	tracelog.ErrorLogger.FatalfOnError("Failed to get WAL filenames: %v\n", err)
+
+	walsByTimelines := groupSegmentsByTimelines(getSegmentsFromFiles(folderFilenames))
+
+	filenamesToRestore, err := GetMissingWals(
+		getSegmentNoFromLsn(lastCommonLsn), lastCommonTl,
+		targetPgData.GetCurrentTimeline(), mapOfTargetTimelineWithSegNo, walsByTimelines)
```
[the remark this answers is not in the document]
Xaspy: Mix up logic in my mind for the moment.

internal/databases/postgres/wal_restore_handler.go
```diff
@@ -0,0 +1,237 @@
...
+) ([]string, error) {
+	result := make([]string, 0)
+	currentSeg := uint64(walsByTimelines[currentTl].maxSegmentNo)
+
+LOOP:
+	for ; currentTl >= lastTl; currentTl-- {
+		// Get wal segment sequence for current timeline
+		walSegSeq, ok := walsByTimelines[currentTl]
+
+		// Iterate over wal segment sequence for current timeline
```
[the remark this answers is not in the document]
Xaspy: Loop was deleted cause of redundant.

usernamedt (approved):
LGTM

One more thing: Can you please add some information regarding this feature to the documentation?

internal/databases/postgres/pg_control_data.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return result, nil
+}
+
+func extractPgControlData(pgControlReader io.Reader) (*PgControlData, error) {
+	bytes := make([]byte, 8192)
+
+	n, err := pgControlReader.Read(bytes)
+	if err != nil || n < 8192 {
```
usernamedt, line 45: use `io.ReadAtLeast`?

Xaspy: Fixed. Added unittests.

Xaspy:
I think we can close #850.

PR #991, opened 2021-05-26
Add readme for redis

proggga:


x4m (approved):

docs/MongoDB.md
```diff
@@ -20,7 +20,7 @@ make install
 make deps
 make mongo_build
 ```
-Users can also install WAL-G by using `make install`. Specifying the GOBIN environment variable before installing allows the user to specify the installation location. On default, `make install` puts the compiled binary in `go/bin`.
+Users can also install WAL-G by using `make mongo_install`. Specifying the GOBIN environment variable before installing allows the user to specify the installation location. On default, `make install` puts the compiled binary in `go/bin`.
```
usernamedt, line 23: Second `make install`left unchanged :(

proggga: Fix in #993

PR #1005, opened 2021-06-07
mysql: until option for binlog push

mialinx:
### Database name
mysql

# Pull request description

### Describe what this PR fix
After primary node crash and recovery one or two last binary logs file may contain invalid data break PITR.
We need option to specify which binary logs are ok to upload.

### DIFF
--- a/cmd/mysql/binlog_push.go
+++ b/cmd/mysql/binlog_push.go
@@ -9,6 +9,8 @@ import (
 
 const binlogPushShortDescription = ""
 
+var untilBinlog string
+
 // binlogPushCmd represents the cron command
 var binlogPushCmd = &cobra.Command{
 	Use:   "binlog-push",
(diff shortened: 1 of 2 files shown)


cmd/mysql/binlog_push.go
```diff
@@ -28,4 +30,5 @@ var binlogPushCmd = &cobra.Command{
 
 func init() {
 	cmd.AddCommand(binlogPushCmd)
+	binlogPushCmd.Flags().StringVar(&untilBinlog, "until", "", "binlog file name to stop at. Current active by default")
```
ostinru, line 33: It is hard to read `Current active by default` for me. I can suggest following description: `When omitted WAL-G uploads binlogs up to current active binlog.`

ostinru (approved):
Looks good for me.

Issue #1006, opened 2021-06-07
`wal-g.linux-amd64` for v1 not released?


Gurpartap:
The README suggests downloading `wal-g.linux-amd64` binary.

```md
A precompiled binary for Linux AMD 64 of the latest version of WAL-G can be obtained under the Releases tab.

To decompress the binary, use:

tar -zxvf wal-g.linux-amd64.tar.gz
mv wal-g /usr/local/bin/
```

However, the binaries attached with v1.0 release seem to only be compiled with ubuntu as the target.

**This breaks our deployment/setup scripts.**

The ubuntu based release fails to run on debian 5.8.10-x86_64-linode138 with the following log:

```
/usr/local/bin/wal-g-pg-ubuntu-20.04-amd64: /lib/x86_64-linux-gnu/libm.so.6:
 version `GLIBC_2.29' not found (required by /usr/local/bin/wal-g-pg-ubuntu-20.04-amd64)
```

usernamedt:
Hi!

Try using the `wal-g-pg-ubuntu-18.04-amd64` image, it is compiled with the older glibc version and should work on debian.

Gurpartap:
I ended up using an older version release for convenience. 

Not sure about using an "ubuntu" release on an upstream debian system. What's Ubuntu about it anyway? Just different glibc?

jhg03a:
The 18.04 ubuntu binary works on Centos7.  I am also curious because if the database support is per binary, that means there is no longer a singular wal-g binary which is going to make downstream packaging harder.  So anyone making a deb, rpm, apk, etc. is going to have to switch from a single wal-g package to something like wal-g-pg, wal-g-sql, wal-g-mysql, etc.  The split for versions of glibc is fine because it makes the difference in unbuntu 18.04/20.04 and Centos7/8 for example which would be different end repositories.

usernamedt:
> The 18.04 ubuntu binary works on Centos7. I am also curious because if the database support is per binary, that means there is no longer a singular wal-g binary which is going to make downstream packaging harder. So anyone making a deb, rpm, apk, etc. is going to have to switch from a single wal-g package to something like wal-g-pg, wal-g-sql, wal-g-mysql, etc. The split for versions of glibc is fine because it makes the difference in unbuntu 18.04/20.04 and Centos7/8 for example which would be different end repositories.

Hi! 

[...]

PR #1007, opened 2021-06-07
Add docs about build for different databases

usernamedt:
Add clarification in docs about precompiled binaries for the different databases. 

https://github.com/wal-g/wal-g/issues/1006

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,13 +28,22 @@ Installation
 ----------
 A precompiled binary for Linux AMD 64 of the latest version of WAL-G can be obtained under the [Releases tab](https://github.com/wal-g/wal-g/releases).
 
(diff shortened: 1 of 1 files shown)


Gurpartap:
Is there a good reason why the releases are targeted for Ubuntu, and are no longer universal (linux amd64)?

It isn't always convenient to build from source, especially on servers that don't have Go or dev tooling installed – only the database software is installed.

usernamedt:
> Is there a good reason why the releases are targeted for Ubuntu, and are no longer universal (linux amd64)?
> 
> It isn't always convenient to build from source, especially on servers that don't have Go or dev tooling installed – only the database software is installed.

Hi, there was no change in the build process, ubuntu-** suffix is here just to indicate the OS that has been used to make a binary (see the corrected docs for details).

x4m (approved):

PR #1012, opened 2021-06-10
[WIP] Add CopyComposer for full backups

slesarev-hub:
### Database name
Postgres

# Pull request description
Add new type of composer for remote copying unchanged tar files during full backup creation process.


go.mod
```diff
@@ -66,3 +66,5 @@ require (
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
 )
+
+replace github.com/wal-g/storages => github.com/slesarev-hub/storages v0.0.0-20210616180710-f709c4ac97aa
```
bunsenmcdubbs, line 70: I understand this is a WIP PR but make sure you remove this before the final review. Changes/updates to the storages module should be merged there first.

internal/databases/postgres/copy_tar_ball_composer.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (c *CopyTarBallComposer) copyTar(tarName string) {
+	tracelog.InfoLogger.Printf("Copying %s ...\n", tarName)
+	c.prevBackup.Folder.CopyObject(
+		utility.BaseBackupPath,
+		c.prevBackup.Name+internal.TarPartitionFolderName+tarName,
+		c.newBackupName+internal.TarPartitionFolderName+tarName)
+	c.copiedTars[tarName] = true
+	c.tarBallQueue.TarBallMaker.AddCopiedTarName(tarName)
```
usernamedt, line 162: Why do we need to include the remote-copied tar to the `tarBallQueue`?

internal/databases/postgres/copy_tar_ball_composer.go
```diff
@@ -0,0 +1,249 @@
...
+			header.status = processed
+			c.tarFileSets[tarName] = append(c.tarFileSets[tarName], fileName)
+			c.files.AddFile(header.fileInfoHeader, header.info, false)
+		}
+	}
+}
+
+func (c *CopyTarBallComposer) getTarBall() internal.TarBall {
+	tarBall := c.tarBallQueue.Deque()
+	tarBall.SetUp(c.crypter)
```
usernamedt, line 178: To get rid of collisions between old tarballs and new ones, you can create a new ones with the custom names, [like it is being done here](https://github.com/wal-g/wal-g/blob/196c92016f2ffb11b261f2e152f35ab066cdd6fb/internal/databases/postgres/bundle.go#L319).

usernamedt (changes requested):

usernamedt:
As for unit tests, you can reuse the WalkTest, just add the `CopyComposer` to it.

internal/databases/postgres/copy_tar_ball_composer.go
```diff
@@ -0,0 +1,249 @@
...
+		currFile.status = fromNew
+	}
+	currFile.info = info
+	c.fileInfo[fileName] = &currFile
+}
+
+func (c *CopyTarBallComposer) AddHeader(fileInfoHeader *tar.Header, info os.FileInfo) error {
+	var fileName = fileInfoHeader.Name
+	var currHeader = headerInfo{}
+	if !c.prevBackup.SentinelDto.Files[fileName].MTime.Equal(info.ModTime()) {
```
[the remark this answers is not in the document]
slesarev-hub: If we will always pack headers to the separate tars, we cannot copy tars with other composers or we should duplicate headers in tars.

internal/databases/postgres/copy_tar_ball_composer.go
slesarev-hub: If we will move name resolver into copy composer, we cannot check tar name correctness [here](https://github.com/wal-g/wal-g/blob/5f14deb57b7b9679f4307839f205e73cc7167c88/internal/databases/postgres/backup_push_handler.go#L262)

usernamedt: It doesn't seems to be a problem if we assign a prefix like `copy_****.tar.gz` to a copied tars.

internal/databases/postgres/copy_tar_ball_composer.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (c *CopyTarBallComposer) SkipFile(tarHeader *tar.Header, fileInfo os.FileInfo) {
+	c.files.AddSkippedFile(tarHeader, fileInfo)
+}
+
+func (c *CopyTarBallComposer) copyTar(tarName string) {
+	tracelog.InfoLogger.Printf("Copying %s ...\n", tarName)
+	var newTarName string
+	if tarName[:5] == "copy_" {
```
usernamedt, line 164: Why not to generate the copy_001,copy_002,... names? It would be simpler

internal/databases/postgres/tar_ball_composer.go
```diff
@@ -57,6 +62,21 @@ func NewTarBallComposerMaker(composerType TarBallComposerType, conn *pgx.Conn,
...
 		return NewRatingTarBallComposerMaker(relFileStats, filePackOptions)
+	case CopyComposer:
+		previousBackupName, err := internal.GetLatestBackupName(folder)
+		if err != nil {
+			return NewRegularTarBallComposerMaker(filePackOptions), nil
+		}
+		tracelog.ErrorLogger.PanicOnError(err)
+		previousBackup := NewBackup(folder, previousBackupName)
+		prevBackupSentinelDto, err := previousBackup.GetSentinel()
+		tracelog.ErrorLogger.PanicOnError(err)
```
usernamedt, line 73: Again, all errors should be handled on the top level, just return them, do not panic :)

pkg/storages/storage/folder.go
```diff
@@ -27,6 +27,8 @@ type Folder interface {
 	ReadObject(objectRelativePath string) (io.ReadCloser, error)
 
 	PutObject(name string, content io.Reader) error
+	
+	CopyObject(baseBackupPath string, objectRelativePath string, dstObject string) error
```
usernamedt, line 31: Is it possible to leave only the `fromPath` and `toPath`? Like this: `CopyObject(fromPath, toPath string) error`. It would be more universal.

internal/config.go
```diff
@@ -194,6 +196,7 @@ var (
 		VerifyPageChecksumsSetting:   true,
 		StoreAllCorruptBlocksSetting: true,
 		UseRatingComposerSetting:     true,
+		UseCopyComposerSetting:       false,
```
usernamedt, line 199: `CommonAllowedSettings` map acts like a set, so use `true` instead of `false`.

usernamedt (changes requested):
Need changes, and:
1. a doc entry regarding new composer type
2. `make fmt`

internal/databases/postgres/tar_ball_composer.go
```diff
@@ -57,6 +61,25 @@ func NewTarBallComposerMaker(composerType TarBallComposerType, conn *pgx.Conn,
 			return nil, err
 		}
 		return NewRatingTarBallComposerMaker(relFileStats, filePackOptions)
+	case CopyComposer:
+		previousBackupName, err := internal.GetLatestBackupName(folder)
+		if err != nil {
+			return NewRegularTarBallComposerMaker(filePackOptions), nil
```
usernamedt, line 67: Add some  tracelog.InfoLogger output to indicate that regular composer is going to be used instead of the copy.

docs/PostgreSQL.md
```diff
@@ -232,6 +232,19 @@ To activate this feature, do one of the following:
 wal-g backup-push /path --rating-composer
 ```
 
+#### Copy composer mode
+
+In the rating composer mode, WAL-G does full backup and copies unchanged tar files from previous full backup. In case when there are no previous full backup, `regular` composer is used.
```
usernamedt, line 237: In the copy composer mode?

usernamedt (changes requested):
Needs some small fixes mentioned above. Also, the outdated `session.go` file is still in place.  And I guess that `ssh` and `fs` storages also need the `CopyObject` implementation.

usernamedt (approved):
Looks OK, but let's add the support for CopyObject to in-memory storage too. I think it should be pretty much straightforward to implement.

internal/databases/postgres/tar_ball_composer.go
usernamedt: I see no change here

PR #1022, opened 2021-06-23
Add 6.2 support, add redis_cli.sh workaround

proggga:
### Database name
Redis

# Pull request description

### Describe what this PR fix
Add support of tests to 6.2, add workaround for redis-cli bug

here is description of bug https://github.com/redis/redis/pull/9135


redis_cli.sh
```diff
@@ -0,0 +1,16 @@
...
+# This script - workaround with redis-cli >= 6.2 bug, when redis-cli tries to fsync /dev/stdout and exit with error
+# see https://github.com/redis/redis/pull/9135
+#
+
+FILENAME=$(mktemp --suffix=redis-cli-stderr)
+redis-cli $@ --rdb - 2>$FILENAME
+exit_code=$?
+cat $FILENAME >&2
+if [[ $exit_code -ne 0 ]] ; then
+	grep "Fail to fsync" $FILENAME | grep -q "Invalid argument" || exit $exit_code
```
perekalov, line 14: Let's print stderr file contents if error is unknown. Then we can always delete temporary files.

proggga: we print stderr always - in previous line "cat $FILENAME >&2"

perekalov (approved):

PR #1024, opened 2021-06-28
Rocksdb basic support

hariton27sy:
### Database name
rocksdb

# Pull request description
Add basic functionality for rocksdb

### Describe what this PR fix
This PR add basic command to create backups for Rocksdb


internal/databases/rocksdb/backup_push_handler.go
```diff
@@ -0,0 +1,70 @@
...
+func HandleBackupPush(uploader *internal.Uploader, dbOptions DatabaseOptions) error {
+	tempDir, err := os.MkdirTemp("", internal.ROCKSDB)
+	if err != nil {
+		return err
+	}
+	tracelog.DebugLogger.Println("Temporaly backup to ", tempDir)
+	defer os.RemoveAll(tempDir)
+
+	var backupInfo BackupInfo
+	if backupInfo, err = saveBackupToLocalDirectory(tempDir, dbOptions); err != nil {
```
mialinx, line 24: It's a bad idea to rely on free disk space for whole backup.
Other database plugins designed to not use local disk, so user can create backup even with 95% disk used.

Can we backup to a stream here or use other tricks ?

hariton27sy: I'm creating backup using rokcdsb library methods. So these methods require path to physical directory to create backup.

On the other hand we can just copty db directory to a stream, altough it's not safe, when db is used by another process

usernamedt: Instead of utilizing the backup engine API, we could use the snapshots API https://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html

It should create hard links instead of plain file copies and also applicable both for full and incremental backups.

usernamedt (changes requested):

usernamedt:
Also, we should add some docker test for backup-push and subsequent backup-fetch.

PR #1031, opened 2021-07-01
Normalize pgDataDirectory to be windows compatible

### LINKED ISSUE #1029: Data directory read from Postgres is different then as parsed (windows)
### Database name
Postgres

# Issue description

Using windows, run:

```
wal-g.exe backup-push "C:/Program Files/PostgreSQL/13/data"
```

And the following error pops up:

```
ERROR: 2021/07/01 15:04:45.755930 !Data directory read from Postgres (C:\Program Files\PostgreSQL\13\data) is different then as parsed (C:/Program Files/PostgreSQL/13/data).
```

Note that the difference is the forward/back slash.

Investigating the code, found that:

https://github.com/wal-g/wal-g/blob/51e92e93cc77403069ca8acf94d7b30c0b23872b/internal/databases/postgres/backup_push_handler.go#L314-L317

Note that it uses "ResolveSymlink" in just one side of comparison. Changing it to both sides solves the problem (since it will normalize inverted slashed in both paths.

iget-master:
Fixes #1029 Details on issue

### DIFF
--- a/internal/databases/postgres/backup_push_handler.go
+++ b/internal/databases/postgres/backup_push_handler.go
@@ -448,7 +448,7 @@ func getPgServerInfo() (pgInfo BackupPgInfo, err error) {
 	if err != nil {
 		return pgInfo, err
 	}
-	pgInfo.pgDataDirectory = utility.ResolveSymlink(pgInfo.pgDataDirectory);
+	pgInfo.pgDataDirectory = utility.ResolveSymlink(pgInfo.pgDataDirectory)
 	tracelog.DebugLogger.Printf("Datadir: %s", pgInfo.pgDataDirectory)
 
 	err = queryRunner.getVersion()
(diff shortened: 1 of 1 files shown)


usernamedt:
Looks good to me, can you run 'make goimports' to make linter happy?

internal/databases/postgres/backup_push_handler.go
```diff
@@ -448,6 +448,7 @@ func getPgServerInfo() (pgInfo BackupPgInfo, err error) {
 	if err != nil {
 		return pgInfo, err
 	}
+	pgInfo.pgDataDirectory = utility.ResolveSymlink(pgInfo.pgDataDirectory);
```
serprex, line 451:
```suggestion
	pgInfo.pgDataDirectory = utility.ResolveSymlink(pgInfo.pgDataDirectory)
```

usernamedt (approved):

PR #1033, opened 2021-07-06
add a 'flags' subcommand to list global flags

### LINKED ISSUE #1032: All help messages list the complete set of global flags
### Database name
documentation issue ; fwiw : seen with main/pg

# Issue description

### Describe your problem
The help message for any `wal-g` subcommand always includes the complete `Global flags` section, which dutifully prints 116 lines on my terminal :

```
$ ./wal-g help backup-push
Makes backup and uploads it to storage

Usage:
wal-g backup-push db_directory [flags]

Flags:
      --add-user-data string          Write the provided user data to the backup sentinel and metadata files.
      --delta-from-name string        Select the backup specified by name as the target for the delta backup
      --delta-from-user-data string   Select the backup specified by UserData as the target for the delta backup
  -f, --full                          Make full backup-push
  -h, --help                          help for backup-push
  -p, --permanent                     Pushes permanent backup
  -r, --rating-composer               Use rating tar composer (beta)
  -s, --store-all-corrupt             Store all corrupt blocks found during page checksum verification
  -v, --verify                        Verify page checksums

Global Flags:
  <... 116 lines, mentioning options which may or may not be relevant to that command ...>
```

IMHO it makes the inline documentation less helpful, as the user will always have to scroll upwards to see the section relevant to the current command.

Just run `wal-g help` or `wal-g help [anycommand]`

### Suggestion

Have a separate flag or command to list all these global flags, and hide them from the default doc messages.

LeGEC:
### Database name
I work with PostgreSQL, right now I only added this `flags` subcommand to `main/pg`

# Pull request description
in `main/pg` binary :

add a 'flags' subcommand to list global flags

additionally :
* remove this list from the default list of options,
* change default doc template to mention :
    "list available flags with 'wal-g flags'"

fixes #1032 (for main/pg)


usernamedt:
Hi!

This is a cool idea. The amount of auto-generated flags is definitely overwhelming.  

Currently, when I run the `wal-g flags` command with no configuration, I get the following error:
```
➜  pg git:(1032-pg-help-flags) ./wal-g flags                                                                                                                                                                   
ERROR: 2021/08/18 16:15:03.156199 Failed to find any configured storage
```

I guess that this command should not depend on any configuration parameters, but simply output the available flags.

This happens because of the AssertRequiredSettingsSet() call in the PersistentPreRun of the `pg` command. Maybe there is another way to add the `flags` subcommand?

usernamedt (changes requested):
needs changes

usernamedt (approved):

cmd/pg/flags.go
```diff
@@ -0,0 +1,54 @@
+package pg
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// flagsCmd represents the delete command
```
usernamedt, line 8:
```suggestion
// flagsCmd represents the flags command
```

usernamedt:
Command itself looks good, but I think that `flags` should be a global subcommand for any database. Can you move it to a separate package? You can use [storage tools global subcommand](https://github.com/wal-g/wal-g/blob/master/cmd/st/st.go) as a reference.

usernamedt:
I moved the flags subcommand and all common cmd init logic to the `common`. Now it should work fine for all databases.

mialinx (approved):
LGTM

perekalov (approved):

Issue #738, opened 2020-08-24
Out of memory error on deployment with millions of files in PGDATA


MannerMan:
**WAL-G version:** 0.2.15 release binary

**PostgreSQL version:** 9.5

**Operating system/version:** CentOS 7.8

Hi,

I'm evaluating WAL-G as a replacement for my companys current backup system WAL-E. We have a schema-level style of sharding our tenants, where each of our database servers hosts 4000 customer schemas, spread over 16 databases. This has worked great for scalability, but presents a challenge for many backup systems - since this layout results in a lot of
files in postgresql data directory.

Output of `ls -1RL /var/lib/pgsql/9.5/data/ | wc -l`;
`6602805`

So above 6 million files, total size 77gb. Around 5 years ago we deployed WAL-E since it was one of the few backup systems that could handle so many files without a problem. However, since WAL-E is no longer maintained we're looking for alternatives. When testing WAL-G, I'm running out of memory when performing a full-backup. It seems to be some kind of go internal memory error, since the memory of the server is not fully utilized. See graphs;

![Screenshot from 2020-08-24 12-28-14](https://user-images.githubusercontent.com/1773068/91034587-86dbd400-e605-11ea-82d1-6f126609ccc8.png)

Server specs:
* 12gb of ram
* shared_buffers set to 3gb
* 1800 hugepages (~ 3.6 gb to fit shared_buffers )
* vm.overcommit = 2 (not allowed to overcommit)
* 7.8 gb regular memory limit
* CentOS 7
* postgres 9.5
* wal-g 0.2.15

Chunk of the error log;
```
INFO: 2020/08/24 11:49:19.354150 Calling pg_stop_backup()
fatal error: runtime: out of memory

runtime stack:
runtime.throw(0x193bfd8, 0x16)
        /home/travis/.gimme/versions/go1.13.9.linux.amd64/src/runtime/panic.go:774 +0x72
runtime.sysMap(0xc110000000, 0x10000000, 0x271a7f8)
        /home/travis/.gimme/versions/go1.13.9.linux.amd64/src/runtime/mem_linux.go:169 +0xc5
runtime.(*mheap).sysAlloc(0x27017c0, 0xfbe0000, 0x150b986, 0x288007575f7)
        /home/travis/.gimme/versions/go1.13.9.linux.amd64/src/runtime/malloc.go:701 +0x1cd
runtime.(*mheap).grow(0x27017c0, 0x7df0, 0xffffffff)
[...]

PR #1037, opened 2021-07-14
PostgreSQL - Simple Composer for reduced memory usage

krnaveen14:
### Database name
PostgreSQL

# Pull request description
Simple Composer for reduced memory usage on instances with `> 100k` files

- Does not track metadata of files backed up (Files and TarFileSets)
- Does not support `WALG_DELTA_MAX_STEPS`, `WALG_DELTA_ORIGIN`, `WALG_USE_REVERSE_UNPACK`, `WALG_SKIP_REDUNDANT_TARS`, `catchup-push`

### Describe what this PR fix
Similar to Issue #738 , WAL-G consumes huge memory on backing up instance with millions of files. This is due to tracking the metadata of each file, maintaining the list of files in each TarFileSet and finally marshalling the data structure to text for creating `sentinel.json` file. Even though PR #740 tries to address the marshalling part by streaming / pipe it, significant memory usage on `BackupSentinelDto` data structure still persists.

Based on the `BackupSentinelDto` usage, it is needed only for configurations such as `DELTA`, `INCREMENTAL`, `RATING_COMPOSER`, `catchup-push` etc... Since we don't use any of the above mentioned configs, `Files` and `TarFileSets` data in `BackupSentinelDto` simply unused.

To reduce the memory usage of WAL-G on instances with millions of files and to make it available & configurable for the community as whole, `SimpleTarBallComposer` is implemented which does not track any of the file metadata, list of files in each TarFileSet and the list of TarFileSets itself.

Simple Composer is supported on both local and remote backups.

**Usage :** `wal-g backup-push [path] --simple-composer` or `WALG_USE_SIMPLE_COMPOSER=true` environment variable

Create 500+ Databases with 1000+ Tables in each Database (not necessary to insert any data). Execute WAL-G `backup-push` to see the memory usage of `wal-g` gradually increases over time and failing with OOM after exhausting available memory (10GB in our case).

### DIFF
--- a/internal/databases/greenplum/backup_push_handler.go
+++ b/internal/databases/greenplum/backup_push_handler.go
@@ -8,8 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/pkg/errors"
-
 	"github.com/google/uuid"
 
 	"github.com/blang/semver"
@@ -91,7 +89,6 @@ func (bh *BackupHandler) buildCommand(contentID int) string {
 	segUserData := NewSegmentUserData()
 	bh.currBackupInfo.segmentBackups[segUserData.ID] = segment
 	cmd := []string{
-		"WALG_LOG_LEVEL=DEVEL",
 		fmt.Sprintf("PGPORT=%d", segment.Port),
 		"wal-g pg",
 		fmt.Sprintf("backup-push %s", segment.DataDir),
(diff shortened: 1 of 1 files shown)


krnaveen14:
@usernamedt waiting on approval for test execution and expecting code review suggestions

docs/PostgreSQL.md
```diff
@@ -232,6 +232,23 @@ To activate this feature, do one of the following:
 wal-g backup-push /path --rating-composer
 ```
 
+#### Simple composer mode
+
+In the simple composer mode, WAL-G does not track metadata of the files backed up. This significantly reduces the memory usage on instances with `> 100k` files.
+
+Limitations or drawbacks in using Simple compose mode is,
```
usernamedt, line 239: You can add an explicit note that simple mode does not support delta and catchup backups.

Also, currently, nothing prevents the user from making a delta or catchup backup with simple mode enabled, without any noticeable error. WAL-G should fail in this case and forbid creating a delta/catchup backup when the simple mode is enabled.

internal/databases/postgres/backup_push_handler.go
```diff
@@ -101,6 +101,8 @@ type BackupHandler struct {
 	pgInfo         BackupPgInfo
 }
 
+var isSimpleComposerTarBallType bool
```
usernamedt, line 104: Using global variables should be avoided, pass this setting directly to the `NewTarballStreamer`

internal/databases/postgres/simple_tar_ball_composer.go
```diff
@@ -0,0 +1,100 @@
+package postgres
```
usernamedt, line 1: I don't see any need for the additional `TarBallComposer` here. `SimpleTarBall` composer does not introduce any new way to pack tarballs, just alters the `BundleFiles` behavior.

So I suggest removing the new composer and passing the SimpleBundleFiles to the regular composer. 

Just replace the `*RegularBundleFiles` with the `BundleFiles` interface [here](https://github.com/wal-g/wal-g/blob/3a5120f1cc208a69db6a834c979a4c97a4d2e7bc/internal/databases/postgres/regular_tar_ball_composer.go#L18) so the regular tarball composer could support both `SimpleBundleFiles` and `RegularBundleFiles`.

usernamedt: As for `TarFileSets`, you can do the similar:
1. Create the `TarFileSets` interface
2. Create `SimpleTarFileSets`/`NopTarFileSets` (new behavior) and `RegularTarFileSets` (old behavior)

internal/databases/postgres/tar_ball_composer.go
```diff
@@ -11,6 +11,7 @@ import (
 // TarBallComposer is used to compose files into tarballs.
 type TarBallComposer interface {
 	AddFile(info *ComposeFileInfo)
+	AddLabelFiles(tarFileSets *TarFileSets, tarBallName string, labelFiles []string)
```
usernamedt, line 14: With the `SimpleTarFileSets` implemented, this will be unnecessary.

usernamedt (changes requested):
Hi!

This is definitely a cool feature, but it needs some adjustments.

usernamedt (commented):
Hi!

Looks like this unit test is failing:

```
2021-08-17T11:25:48.0366780Z === RUN   TestGetFilesToUnwrap_UtilityFiles
2021-08-17T11:25:48.0367313Z     backup_test.go:49: 
2021-08-17T11:25:48.0367713Z         	Error Trace:	backup_test.go:49
2021-08-17T11:25:48.0378206Z         	Error:      	Not equal: 
2021-08-17T11:25:48.0378808Z         	            	expected: map[string]bool{"/global/pg_control":true, "backup_label":true, "tablespace_map":true}
2021-08-17T11:25:48.0379442Z         	            	actual  : map[string]bool(nil)
2021-08-17T11:25:48.0379793Z         	            	
2021-08-17T11:25:48.0380100Z         	            	Diff:
2021-08-17T11:25:48.0380664Z         	            	--- Expected
2021-08-17T11:25:48.0381016Z         	            	+++ Actual
2021-08-17T11:25:48.0381447Z         	            	@@ -1,6 +1,2 @@
2021-08-17T11:25:48.0381949Z         	            	-(map[string]bool) (len=3) {
2021-08-17T11:25:48.0382567Z         	            	- (string) (len=18) "/global/pg_control": (bool) true,
2021-08-17T11:25:48.0383204Z         	            	- (string) (len=12) "backup_label": (bool) true,
2021-08-17T11:25:48.0383852Z         	            	- (string) (len=14) "tablespace_map": (bool) true
2021-08-17T11:25:48.0384349Z         	            	-}
2021-08-17T11:25:48.0384701Z         	            	+(map[string]bool) <nil>
2021-08-17T11:25:48.0385035Z         	            	 
2021-08-17T11:25:48.0385506Z         	Test:       	TestGetFilesToUnwrap_UtilityFiles
2021-08-17T11:25:48.0386267Z --- FAIL: TestGetFilesToUnwrap_UtilityFiles (0.00s)
2021-08-17T11:25:48.0386909Z === RUN   TestGetFilesToUnwrap_NoMoreFiles
2021-08-17T11:25:48.0387661Z --- PASS: TestGetFilesToUnwrap_NoMoreFiles (0.00s)
```

internal/databases/postgres/bundle_files.go
```diff
@@ -67,27 +72,51 @@ func (files *StatBundleFiles) AddFileWithCorruptBlocks(tarHeader *tar.Header,
...
 
+func (files *StatBundleFiles) AddFileDescription(name string, backupFileDescription internal.BackupFileDescription) {
+	files.Store(name, backupFileDescription)
+}
+
 func (files *StatBundleFiles) GetUnderlyingMap() *sync.Map {
 	return &files.Map
 }
 
+type SimpleBundleFiles struct {
```
usernamedt, line 100: `NopBundleFiles`?

cmd/pg/backup_push.go
```diff
@@ -54,6 +55,10 @@ var (
 			if useRatingComposer {
 				tarBallComposerType = postgres.RatingComposer
 			}
+			useSimpleComposer = useSimpleComposer || viper.GetBool(internal.UseSimpleComposerSetting)
```
usernamedt, line 58: Well, now this is not actually a composer type, but a flag to enable the low memory usage. So I suggest to rename  this to `--reduce-memory-usage` or something like this since it is not a new composer type.

usernamedt (changes requested):
Seems OK, but need to change the flag

krnaveen14:
Closing this in favor new PR #1101

PR #1039, opened 2021-07-19
use RFC3339 time format for mysql backup push

reshke:
### Database name
MySQL

# Pull request description
Fix MYSQL metadata backup start/stop time format 

### Describe what this PR fix

let's make the format of the backup time the same with the tonsure


mialinx (changes requested):
* StartLocalTime  is time.Time, not  a string
* changing sentinel format is not backward compatible change

PR #1049, opened 2021-07-29
Enforce that UserData is a valid JSON

usernamedt:


reshke (approved):

internal/databases/mysql/backup_push_handler.go
```diff
@@ -59,7 +59,7 @@ func HandleBackupPush(uploader *internal.Uploader, backupCmd *exec.Cmd, isPerman
 		CompressedSize:   uploadedSize,
 		UncompressedSize: rawSize,
 		IsPermanent:      isPermanent,
-		UserData:         userData,
+		UserData:         internal.UnmarshalSentinelUserData(userData),
```
mialinx, line 62: Any error hanlding during unmarshalling ?

mialinx (changes requested):

internal/databases/mysql/backup_push_handler.go
usernamedt: it is already being handled in `internal.UnmarshalSentinelUserData`. If failed to parse as JSON, it will leave the string as it is.

mialinx (approved):

perekalov (approved):

PR #1052, opened 2021-08-05
Always prevent overwrite of .history files

### LINKED ISSUE #420: Prevent overwrite of .history file
By default, WAL-G does not turn on WAL overwrite prevention, since this check consumes resources and does not provide much stronger guarantees.

But this is incorrect for .history files: PostgreSQL algorithm to pick timeline relies on unoverwritable archive.
this can affect a PostgreSQL cluster with concurrent promote.

serprex:
### Database name
Postgres

### Describe what this PR fix
Fixes #420 (wal-g should never overwrite .history files as postgres relies on overwrite detection in this case)

With WALG_PREVENT_WAL_OVERWRITE=false, create a .history file, push it, change it, push it again. The second push should fail


x4m (approved):
Looks great, thank you!

PR #1056, opened 2021-08-09
AZURE_STORAGE_SAS_TOKEN: prepend a ? when not included in token

### LINKED ISSUE #1034: Azure SAS Token Must be Prefixed with ?
### Database name
PostgreSQL

# Issue description

### Describe your problem

Azure SAS tokens must be prefixed with `?`, or the token get's merged with the path resulting in a 404 error. With `?` prefixed, the token is correctly merged with the query string.

(the go library handling Azure normally REDACTS the signature from logs, but only when prefixed with `?`)

```
> az storage container generate-sas \
> --account-name <blah> \
> --name <blah> \
> --auth-mode key \
> --permissions acdlrw \
> --https-only \
> --expiry $EXPIRY
se=2031-07-06T09%3A54%3A00Z&sp=racwdl&spr=https&sv=2018-11-09&sr=c&sig=<blah>
```

Giving this string directly to wal-g e.g.
```
export AZURE_STORAGE_SAS_TOKEN=se=2031-07-06T09%3A54%3A00Z&sp=racwdl&spr=https&sv=2018-11-09&sr=c&sig=<blah>
```

Results in 404 errors (with the string concatenated with the requires URL path).

---

IMO, this shouldn't be required (or it should be documented?). It feels rather odd to require it.

This could be a bug in the go library, not 100% sure. The go library is definitely impacted by this bug https://github.com/Azure/azure-storage-azcopy/issues/122#issuecomment-823077673 - so updating that dependency might be a good idea anyway.

The issue https://github.com/wal-g/wal-g/issues/1008 appears unrelated.

serprex:
I originally understood the ? to be part of the token,
but it turns out most SAS token libraries don't include it

Fixes #1034

### DIFF
--- a/pkg/storages/azure/folder.go
+++ b/pkg/storages/azure/folder.go
@@ -68,6 +68,11 @@ func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder,
 		if accountToken, usingToken = settings[SasTokenSetting]; !usingToken {
 			return nil, NewCredentialError(AccessKeySetting)
 		}
+
+		// Tokens may or may not begin with ?, normalize these cases
+		if !strings.HasPrefix(accountToken, "?") {
+			accountToken = "?" + accountToken
+		}
 	}
 	if environmentName, ok = settings[EnvironmentName]; !ok {
 		environmentName = defaultEnvName


usernamedt (approved):
lgtm

Issue #1059, opened 2021-08-11
wal-g for sqlserver - log backups have inaccurate timestamps


trootnev:
#Database
SQL Server

# Issue description
Log backup names do not reflect the real timestamp of log backup (backup finish date).

### Describe your problem
This causes issues during point in time recovery when StopAt mark is expected to be reached in one file but it actually is in the previous one and restore operation fails.

### Please provide steps to reproduce
try restoring database to a point 1-2 seconds before/after log backup. The bigger the log backup the bigger the drift between backup finish date and file timestamp. 

### Please add config and wal-g stdout/stderr logs for debug purpose
```

C:\Users\Administrator > &'C:\\Program Files\\wal-g-sqlserver\\wal-g-sqlserver.exe' --config 'C:\\ProgramData\\wal-g\\wal-g.yaml' log-restore --from db1 --databases db2 --since 'base_20210809T100051Z' --until '2021-08-09T10:00:52Z'
2021/08/09 13:10:32 WARN: You specified both instance name and port in the connection string, port will be used and instance name will be ignored
INFO: 2021/08/09 13:10:32.663861 running proxy at backup.local:443
INFO: 2021/08/09 13:10:33.122971 starting restore database [db2] log from URL = 'https://backup.local/wal_005/wal_20210809T100052Z/db1/blob_000'
INFO: 2021/08/09 13:10:33.930710 database [db2] log restore succefully finished
INFO: 2021/08/09 13:10:34.026205 starting restore database [db2] log from URL = 'https://backup.local/wal_005/wal_20210809T101502Z/db1/blob_000' «««« this is the log file that shouldn't be here at all
ERROR: 2021/08/09 13:10:34.614533 database [db2] log restore failed: mssql: RESTORE LOG is terminating abnormally.
ERROR: 2021/08/09 13:10:34.615531 overall log restore failed: mssql: RESTORE LOG is terminating abnormally.

```

usernamedt:
Fixed in https://github.com/wal-g/wal-g/pull/1060

usernamedt:
Fixed in https://github.com/wal-g/wal-g/pull/1060

PR #1060, opened 2021-08-11
Fix for #1059

trootnev:
### Database name
SQL Server

# Pull request description
Added function to get the real log backup metadata
it goes to the URL and gets it with RESTORE HEADERONLY
it is accumulated for each log backup that is supposed to be restored
later this data is iterated over during log restore making restore process more reliable.

### Describe what this PR fix
Issue #1059

### DIFF
--- a/internal/databases/sqlserver/log_restore_handler.go
+++ b/internal/databases/sqlserver/log_restore_handler.go
@@ -68,9 +68,12 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
 					logBackupName, fromname)
 				continue
 			}
-			err = restoreSingleLog(ctx, db, folder, logBackupName, dbname, fromname, stopAt)
+			isLastOne, err := restoreSingleLog(ctx, db, folder, logBackupName, dbname, fromname, stopAt)
 			if err != nil {
 				return err
+			}
+			if isLastOne {
+				break
 			}
 		}
 		if !noRecovery {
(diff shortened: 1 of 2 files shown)


internal/databases/sqlserver/log_restore_handler.go
```diff
@@ -56,22 +56,43 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
...
 		dbname := dbnames[i]
 		fromname := fromnames[i]
+		var lastPiece bool
+		var LogsMetadata []BackupProperties
+
 		for _, logBackupName := range logs {
-			ok, err := doesLogBackupContainDB(folder, logBackupName, fromname)
+			baseURL := getLogBackupURL(logBackupName, fromname)
+			basePath := getLogBackupPath(logBackupName, fromname)
+			blobs, err := listBackupBlobs(folder.GetSubFolder(basePath))
```
mialinx, line 65: What if blobs is empty list ? eg database `fromname` is not present in particular logBackupName.
Before this patch there was a `doesLogBackupContainDB` check here

internal/databases/sqlserver/log_restore_handler.go
```diff
@@ -56,22 +56,43 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
...
+			if err != nil {
+				return err
+			}
+			urls := buildRestoreUrls(baseURL, blobs)
+			trn, err := ListBackupProperties(db, urls, logBackupName)
```
mialinx, line 70: What `trn` stands for ?

internal/databases/sqlserver/log_restore_handler.go
```diff
@@ -56,22 +56,43 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
...
-			if !ok {
-				// some log backup may not contain particular database in case
-				// it was created or dropped between base backups
+			LogsMetadata = append(LogsMetadata, trn[0])
+		}
+
+		logBeginTS := time.Date(1900, 1, 1, 1, 1, 1, 1, time.UTC)
+
+		for _, bak := range LogsMetadata {
+			if bak.DatabaseName != fromname {
```
mialinx, line 80: Useless check. We collect properties only for log backups for `fromname` (line 63, 64)
There should not be properties for other databases.

internal/databases/sqlserver/log_restore_handler.go
```diff
@@ -89,16 +110,24 @@ func restoreSingleLog(ctx context.Context,
 	logBackupName string,
 	dbname string,
 	fromname string,
-	stopAt time.Time) error {
+	stopAt time.Time,
+	lastPiece bool) error {
```
mialinx, line 114: why do we need this flag?

mialinx (changes requested):

internal/databases/sqlserver/log_restore_handler.go
trootnev: Not to add WITH STOPAT to every log restore operation except the last one.

mialinx (approved):

PR #1064, opened 2021-08-17
Fix ubuntu versions mismatch in CI

usernamedt:
Ubuntu version has been bumped from 16.04 to 20.04 in [this PR](https://github.com/wal-g/wal-g/pull/866) but looks like these places should be bumped too.


usernamedt:
unrelated to the issue

PR #1069, opened 2021-08-19
Add get/put/rm to storage tools

usernamedt:
### ``get``
Download the specified storage object. It has 3 different download modes:

1. `raw` (default) - download the remote object as it is (no decryption/decompression)
2. `decrypt` - download and decrypt the remote object
3. `decompress` - download, decrypt and decompress the remote object

To specify the download mode, add the `-m [mode]` flag.

``wal-g st get path/to/remote_file path/to/local_file`` download the file from storage.

``wal-g st get path/to/remote_file path/to/local_file -m decrypt`` download and decrypt the file from storage.

### ``rm``
Remove the specified storage object.

``wal-g st rm path/to/remote_file`` remove the file from storage.

### ``put``
Upload the specified file to the storage. 

``wal-g st put path/to/local_file path/to/remote_file`` upload the local file to storage.

### DIFF
--- a/cmd/st/get_object.go
+++ b/cmd/st/get_object.go
@@ -1,8 +1,6 @@
 package st
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
@@ -12,8 +10,8 @@ import (
 const (
 	getObjectShortDescription = "Download the specified storage object"
 
-	downloadModeFlag      = "mode"
-	downloadModeShorthand = "m"
+	noDecryptFlag    = "no-decrypt"
+	noDecompressFlag = "no-decompress"
 )
 
 // getObjectCmd represents the getObject command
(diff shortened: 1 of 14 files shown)


cmd/st/get_object.go
```diff
@@ -0,0 +1,57 @@
...
+		tracelog.ErrorLogger.FatalOnError(err)
+
+		storagetools.HandleGetObject(objectPath, dstPath, folder, mode)
+	},
+}
+
+var downloadModeRaw string
+
+func ParseDownloadMode(mode string) (storagetools.DownloadMode, error) {
+	switch mode {
```
mialinx, line 41: What about  decrypt AND decompress ? It's the main case , afaik.

internal/storagetools/delete_object_handler.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func HandleDeleteObject(objectPath string, folder storage.Folder) {
+	// some storages may not produce an error on deleting the non-existing object
+	exists, err := folder.Exists(objectPath)
+	tracelog.ErrorLogger.FatalfOnError("Failed to check object existence: %v", err)
+	if !exists {
+		tracelog.ErrorLogger.Fatalf("Object %s does not exist", objectPath)
+	}
+	err = folder.DeleteObjects([]string{objectPath})
```
mialinx, line 15: Does every storage engine support compex object paths like `a/b/c/d.zip` ? 
Don't we need to traverse folder down to with SubFolder().SubFolder() before accessing object ?

internal/storagetools/get_object_handler.go
```diff
@@ -0,0 +1,105 @@
...
+	DownloadDecompress DownloadMode = "decompress"
+	DownloadDecrypt    DownloadMode = "decrypt"
+)
+
+func HandleGetObject(objectPath, dstPath string, folder storage.Folder, mode DownloadMode) {
+	fileName := path.Base(objectPath)
+	targetPath, err := getTargetFilePath(dstPath, fileName)
+	tracelog.ErrorLogger.FatalfOnError("Failed to determine the destination path: %v", err)
+
+	dstFile, err := os.OpenFile(targetPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC|os.O_EXCL, 0666)
```
mialinx, line 29: 0640 is more safe mask.    backup files may contain sensetive data.

internal/storagetools/put_object_handler.go
```diff
@@ -0,0 +1,51 @@
...
+	"path/filepath"
+
+	"github.com/wal-g/wal-g/utility"
+
+	"github.com/wal-g/tracelog"
+	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/internal/ioextensions"
+)
+
+func HandlePutObject(localPath, dstPath string, uploader *internal.Uploader, overwrite bool) {
```
mialinx, line 15: do we need compression and encryption here ?

mialinx (changes requested):

cmd/st/get_object.go
usernamedt: We can't decompress without the decryption so I choose to name it simply "decompress".

internal/storagetools/delete_object_handler.go
usernamedt: S3 should support this. As for other storages, I think that this logic needs to be implemented in the corresponding storage implementation. I propose to leave it like this here.

internal/storagetools/put_object_handler.go
usernamedt: Initially, I thought that we can control it via the config settings. But I agree that it would be more flexible to add the `--no-compress` and `--no-encrypt `flags here.

cmd/st/get_object.go
usernamedt: Done in 947351b

internal/storagetools/get_object_handler.go
usernamedt: Fixed in 947351b

internal/storagetools/put_object_handler.go
usernamedt: I implemented the `--no-compres` and `--no-encrypt` flags in 02ad732

mialinx (approved):

PR #1073, opened 2021-08-23
Update Copy composer benchmark

slesarev-hub:
### Database name
Postgres

# Pull request description

This is pr with Copy composer benchmark in case when database is updated between pushes.

Related links:
#1012


usernamedt (approved):
👍

PR #1075, opened 2021-08-23
Getting backup finish date from backup file.

trootnev:
Validating if stopat position falls into current log restore operation interval

### Database name
SQLSERVER
# Pull request description
Additional check per every log restore operation to make sure that stopat position is not left behind in the previous log or database backup. 
### Describe what this PR fix
This happens if we are trying to restore database to a point right after the full backup. Chances are that stopat position is in the full backup and when we are trying to apply the very first log backup we get and error that database is already rolled forward further that the stopat point.
// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/internal/databases/sqlserver/log_restore_handler.go
+++ b/internal/databases/sqlserver/log_restore_handler.go
@@ -56,24 +56,20 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
 	err = runParallel(func(i int) error {
 		dbname := dbnames[i]
 		fromname := fromnames[i]
-		baseURL := getDatabaseBackupURL(backupName, fromname)
-		basePath := getDatabaseBackupPath(backupName, fromname)
-		blobs, err := listBackupBlobs(folder.GetSubFolder(basePath))
 		if err != nil {
 			return err
 		}
-		urls := buildRestoreUrls(baseURL, blobs)
-		backupMetadata, err := ListBackupProperties(db, urls, baseURL)
+		backupMetadata, err := GetBackupProperties(db, folder, false, backupName, fromname)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


internal/databases/sqlserver/log_restore_handler.go
```diff
@@ -56,6 +56,24 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
...
+		if err != nil {
+			return err
+		}
+		var dbBackupMetadata *BackupProperties
+		for _, dbBackupMetadata = range backupMetadata {
+			if dbBackupMetadata.DatabaseName == fromname {
+				break
+			}
+		}
+		prevBackupFinishdate := dbBackupMetadata.BackupFinishDate
```
mialinx, line 76: May it have sense to create helpers like `getDatabaseBackupProperties(backupName, dbName)` & `getLogBackupProperties(logBackupName, dbName)` ? 

BTW, term metadata is already used in wal-g.. it may be confusing. or not )

internal/databases/sqlserver/log_restore_handler.go
```diff
@@ -68,13 +86,21 @@ func HandleLogRestore(backupName string, untilTS string, dbnames []string, fromn
 					logBackupName, fromname)
 				continue
 			}
-			isLastOne, err := restoreSingleLog(ctx, db, folder, logBackupName, dbname, fromname, stopAt)
+			isLastOne, CurrentBackupFinishdate, err := restoreSingleLog(ctx,
```
mialinx, line 89: `isLastOne` return argument is no longer needed as it computed using `backupFinishDate`.
We can return just `backupFinishDate`

mialinx (changes requested):

mialinx (approved):

PR #1079, opened 2021-08-26
setting db to single_user mode to avoid blocking

trootnev:
### Database name
SQLSERVER

# Pull request description
Kill database connections during database restore in case of replace.
### Describe what this PR fix
If we restore a database replacing the existing one, then, if there are open connections to it, then restore operation will be blocked.

open a connection to a database and try to restore it from backup. Operation will timeout.


serprex:
This code fails if dbname contains an apostrophe

PR #1087, opened 2021-09-09
libsodium: fix OOB read, add transforms (fixes #1086)

### LINKED ISSUE #1086: libsodium encryption/decryption: improper key handling (out-of-boundary reads, key truncation)
# Issue description

When using libsodium encryption the user specifies the key to use using `WALG_LIBSODIUM_KEY` or `WALG_LIBSODIUM_KEY_PATH`.

This key will be passed (unmodified, except for whitespace trimming) to the following C functions.
```
crypto_secretstream_xchacha20poly1305_init_push
crypto_secretstream_xchacha20poly1305_init_pull
```
The libsodium functions expect a **fixed-size** key of exactly 32 bytes.
The byte array that wal-g passes to the C function can be of **arbitrary** length. There is no key size specified in the wal-g documentation, and no runtime checks are performed on the key length.

The issues that arive depend on the size of the key.

**if len(key) > 32**
If the key used is longer than 32 bytes, all bytes after the 32nd byte are effectively ignored. There should be no further issue, backup & restore should work without issues.

**if len(key) < 32**
If the key used is shorter than 32 bytes. libsodium will read some bytes from outside of the go array boundaries. The resulting behaviour depends on go internals.

In my tests, when using a very short keys (i.e. len(key) = 5) decrypt fails with a "corrupted chunk" error. It is however unlikely that a user would use such a small key length.

A medium-sized keys of 20 characters can backup & restore fine. It's probably because the go memory allocator is zeroing 
& rounding up allocated memory. The "missing" 12 bytes will be read as zero bytes.

## Further Issue
There is no key derivation. Therefore using a "human" password is not ideal as a libsodium key.

For maximum entropy, the user should be able to supply a random 32 byte key.

Ideally, the application would accept base64 or hex-encoded keys. Supplying a raw 32 byte key in the key file supplied to `WALG_LIBSODIUM_KEY_PATH` is error-prone, due to whitespace trimming.

joelbohnes:
# Pull request description
Changes to libsodium Crypto:
- keys are now always 32byte in length before passing to libsodium
- a warning is now issued if wal-g was used with a libsodium key that isn't 32-byte
- key transforms allows keys to be given in base64 or hex formats

### Describe what this PR fix
See https://github.com/wal-g/wal-g/issues/1086


docs/README.md
```diff
@@ -68,12 +68,16 @@ To configure the name of a file containing private key of Yandex Cloud Service A
 
 * `WALG_LIBSODIUM_KEY`
 
-To configure encryption and decryption with libsodium. WAL-G uses an [algorithm](https://download.libsodium.org/doc/secret-key_cryptography/secretstream#algorithm) that only requires a secret key.
+To configure encryption and decryption with libsodium. WAL-G uses an [algorithm](https://download.libsodium.org/doc/secret-key_cryptography/secretstream#algorithm) that only requires a secret key. libsodium keys are fixed-size keys of 32 bytes. To generate a key, you can something like `openssl rand -hex 32` (set `WALG_LIBSODIUM_KEY_TRANSFORM` to `hex`) or `openssl rand -base64 32` (set `WALG_LIBSODIUM_KEY_TRANSFORM` to `base64`). Because environemnt variables and cli arguments are prone to accidental leaks, you might want to consider using the `WALG_LOBSIDUM_KEY_PATH` option instead.
```
usernamedt, line 71: s/WALG_LOBSIDUM_KEY_PATH/WALG_LIBSODUM_KEY_PATH/

internal/crypto/libsodium/crypter.go
```diff
@@ -66,17 +78,58 @@ func (crypter *Crypter) setup() (err error) {
...
+		if err != nil {
+			return
+		}
+
+		keyString = strings.TrimSpace(string(keyFileContents))
 	}
 
-	key, err := ioutil.ReadFile(crypter.KeyPath)
+	var rawKey []byte
+	if crypter.KeyTransform == "base64" {
```
usernamedt, line 94: This definitely deserves decomposing to a separate file or at least function. Ideal way is to introduce a `LibsodiumKeyTransformer` interface, write different implementations (base64/hex/rpad-zero/no-transform) and separate constructor for this logic like `NewLibsodiumKeyTransformer(name string)`

usernamedt (changes requested):

usernamedt:
Thanks for your contribution! Needs a couple of mostly cosmetic fixes. Also, I guess it is a good idea to set an explicit default value for `WALG_LIBSODIUM_KEY_TRANSFORM` [here](https://github.com/wal-g/wal-g/blob/e49e3bfd0fb747ec7d8abab08cb1e72eeca84c0d/internal/config.go#L134) and set it to `none`

joelbohnes:
Thank you for your feedback! I agree with your comments and proposed changes.
Please allow me some time to implement the changes.

serprex (approved):

internal/crypto/libsodium/crypter.go
joelbohnes: I moved the transform code into a separate file (`keytransform.go`). The transformations are now individual functions (`keyTransformBase64`,`keyTransformHex` and `keyTransformNone`).

Haven't implemented the interface approach yet. I presume if we would implement it like that, we'd change the
Crypter constructor from
```
CrypterFromKey(key string, keyTransform string)
```
to
```
CrypterFromKey(key string, keyTransform LibsodiumKeyTransformer)
```

This now also places the responsibility of creating a `LibsodiumKeyTransformer` instance to the caller. Creating a transform (i.e. using your suggested `NewLibsodiumKeyTransformer` function) may fail (by passing a unknown transformation name).
The caller (`ConfigureCrypter` in `internal/configure.go`) does not have error handling (i.e. cannot return an error), so you'd either have to
1) Panic()
2) Pass 'nil' as transform and check for nil in the crypter setup()
3) Pass a transform which always returns an error (which will return the error later during setup())

All of those options don't seem very clean. That's the reason why I haven't implemented the interface approach.

usernamedt:
@joelbohnes Looks OK, but can you have a look at the failing unit tests?

joelbohnes:
> @joelbohnes Looks OK, but can you have a look at the failing unit tests?

I fixed the unit test. Don't know how that blatant error managed to slip through my testing.

usernamedt (approved):

PR #1096, opened 2021-09-21
sqlserver: reuse running wal-g proxy service

mialinx:
###  SQLServer

Option to reuse wal-g proxy running as system service

This will allow run several wal-g command (like backup-push and log-push) in parallel.

### DIFF
--- a/docs/SQLServer.md
+++ b/docs/SQLServer.md
@@ -122,3 +122,12 @@ wal-g delete retain 3
 wal-g delete before backup_name
 wal-g delete everything
 ```
+
+Proxy as Service
+-----------------
+By default any wal-g command, like backup-push, runs proxy in background for the duration of the command.
+Alternatively wal-g commands may use proxy started as system service.
+In order to do this you need to start `wal-g proxy` as system service (or daemon on linux) and add following config options:
+```bash
+SQLSERVER_REUSE_PROXY: True
+```
--- a/internal/databases/sqlserver/sqlserver.go
+++ b/internal/databases/sqlserver/sqlserver.go
@@ -1,10 +1,12 @@
 package sqlserver
 
 import (
+	"context"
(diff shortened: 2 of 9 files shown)


internal/databases/sqlserver/sqlserver.go
```diff
@@ -493,3 +497,38 @@ func ListBackupProperties(db *sql.DB, urls string, logBackupName string) ([]*Bac
...
+		tracelog.ErrorLogger.PrintOnError(lw.c.Close())
+	}
+}
+
+func RunOrReuseProxy(ctx context.Context, cancel context.CancelFunc, folder storage.Folder) (*LockWrapper, error) {
+	bs, err := blob.NewServer(folder)
+	if err != nil {
+		return nil, xerrors.Errorf("proxy create error: %v", err)
+	}
+	reuse, _, err := internal.GetBoolSetting(internal.SQLServerUseProxyService)
```
usernamedt, line 516: We usually retrieve the config settings at the top level.

docs/SQLServer.md
```diff
@@ -122,3 +122,12 @@ wal-g delete retain 3
...
 wal-g delete everything
 ```
+
+Proxy as Service
+-----------------
+By default any wal-g command, like backup-push, runs proxy in background for the duration of the command.
+Alternatively wal-g commands may use proxy started as system service.
+In order to do this you need to start `wal-g proxy` as system service (or daemon on linux) and add following config options:
+```bash
+SQLSERVER_USE_PROXY_SERVICE: True
```
usernamedt, line 132: I see that in code it's actually called 'reuse'

usernamedt (approved):
seems OK

docs/SQLServer.md
mialinx: Name fixed

PR #1100, opened 2021-09-24
[greenplum] check for existing backup during backup-push

usernamedt:
GP6X doesn't support non-exclusive backup so the backup would fail if wal-g backup-push is launched while there is at least one backup is running on some of the segments. I added a small check to check if there is running backup on at least one segment and fail backup immediately in this case.


x4m (approved):
There is a race condition: absence of running backup does not guarantee that other backup will not start. But I think it’s OK.

usernamedt:
> There is a race condition: absence of running backup does not guarantee that other backup will not start. But I think it’s OK.

In this case, the backup will fail too, but not immediately so I agree it is fine to leave it as it is for now.

PR #1101, opened 2021-09-26
PostgreSQL - Reduce memory usage by not tracking metadata of the files backed up

krnaveen14:
### Database name
PostgreSQL

# Pull request description

Similar to Issue #738 , WAL-G consumes huge memory on backing up instance with millions of files. This is due to tracking the metadata of each file, maintaining the list of files in each TarFileSet and finally marshalling the data structure to text for creating `sentinel.json` file. Even though PR #740 tries to address the marshalling part by streaming / pipe it, significant memory usage on `BackupSentinelDto` data structure still persists.

Based on the `BackupSentinelDto` usage, it is needed only for configurations such as `DELTA`, `INCREMENTAL`, `RATING_COMPOSER`, `catchup-push` etc... Since we don't use any of the above mentioned configs, `Files` and `TarFileSets` data in `BackupSentinelDto` simply unused.

To reduce the memory usage of WAL-G on instances with millions of files and to make it available & configurable for the community as whole, `NopTarBallComposer` & `NopBundleFiles` is implemented which does not track any of the file metadata, list of files in each TarFileSet and the list of TarFileSets itself.

`reduce-memory-usage` is supported on both local and remote backups.

**Usage :** `wal-g backup-push [path] --reduce-memory-usage` or `WALG_REDUCE_MEMORY_USAGE=true` environment variable

( Successor to PR #1037 )

Create 500+ Databases with 1000+ Tables in each Database (not necessary to insert any data). Execute WAL-G `backup-push` to see the memory usage of `wal-g` gradually increases over time and failing with OOM after exhausting available memory (10GB in our case).

### DIFF
--- a/cmd/pg/backup_push.go
+++ b/cmd/pg/backup_push.go
@@ -26,6 +26,7 @@ const (
 	deltaFromUserDataFlag     = "delta-from-user-data"
 	deltaFromNameFlag         = "delta-from-name"
 	addUserDataFlag           = "add-user-data"
+	reduceMemoryUsageFlag     = "reduce-memory-usage"
 
 	permanentShorthand             = "p"
 	fullBackupShorthand            = "f"
@@ -56,7 +57,8 @@ var (
 			if useRatingComposer {
 				tarBallComposerType = postgres.RatingComposer
 			}
-			if useCopyComposer || viper.GetBool(internal.UseCopyComposerSetting) {
+			useCopyComposer = useCopyComposer || viper.GetBool(internal.UseCopyComposerSetting)
+			if useCopyComposer {
 				fullBackup = true
 				tarBallComposerType = postgres.CopyComposer
 			}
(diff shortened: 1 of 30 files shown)


krnaveen14:
@usernamedt I feel naming of `--reduce-memory-usage` is pretty naive. Any suggestions regarding this would be welcome.

cmd/pg/backup_push.go
```diff
@@ -26,6 +26,7 @@ const (
 	deltaFromUserDataFlag     = "delta-from-user-data"
 	deltaFromNameFlag         = "delta-from-name"
 	addUserDataFlag           = "add-user-data"
+	reduceMemoryUsageFlag     = "reduce-memory-usage"
```
usernamedt, line 29: Actually, we plan to extract the Files and TarFileSets to the separate file in the future. It can be named, for example, `files_metadata.json`.  So I guess we can name this feature like `without-files-metadata`.

krnaveen14: @usernamedt `without-files-metadata` feels spot on than the vague `reduce-memory-usage`. I'll proceed with it.

internal/databases/postgres/backup_fetch_handler.go
```diff
@@ -62,9 +62,8 @@ func chooseTablespaceSpecification(sentinelDtoSpec, spec *TablespaceSpec) *Table
 
 // TODO : unit tests
 // deltaFetchRecursion function composes Backup object and recursively searches for necessary base backup
-func deltaFetchRecursionOld(backupName string, folder storage.Folder, dbDataDirectory string,
+func deltaFetchRecursionOld(backup Backup, folder storage.Folder, dbDataDirectory string,
```
usernamedt, line 65: This change is not related to the files metadata, but helps to avoid downloading it twice.

x4m (approved):

PR #950, opened 2021-04-29
rename .ready archive status files

gsmol:
Hello!
I would like to propose a PR, which introduces the walE-like behavior of  renaming `.ready` status files into `.done` for successfully upladed WAL segments.
It allows to save time by not attempting to upload already uploaded files.
This behaivor is enabled by `--pg-ready-rename` flag.
Feedback and thought on possible improvements are welcome!

PR #1105, opened 2021-10-01
Fix permissions on pg_test scenario file

Xaspy:
POSTGRESQL.

Found this problem when trying to run postgres tests. Permissions on `archiving_ready_rename.sh` are not equal to other files permissions in this directory. Changed permission to 775 (like other files in that directory).


usernamedt:
Thanks! Can you also include this test to the CI pipeline?

usernamedt:
@gsmol looks like the test introduced in [this PR](https://github.com/wal-g/wal-g/pull/950) is actually failing, can you take a look?

internal/databases/postgres/bguploader.go
```diff
@@ -220,7 +220,8 @@ func (b *BgUploader) upload(walStatusFilename string) bool {
 		return false
 	}
 
-	if err := b.uploader.ArchiveStatusManager.MarkWalUploaded(walFilename); err != nil {
+	err = b.uploader.ArchiveStatusManager.MarkWalUploaded(walFilename)
```
serprex, line 223: Revert this. You're overwriting the error from `uploadWALFile`

serprex: Oh I see, `uploadWALFile` has return false, so this makes sense. So if this is breaking a test the test is wrong

serprex (approved):

PR #1129, opened 2021-10-19
Swift: Ignore files that were not found during ListFolder

tnt-dev:
Some files can disappear during ListFolder execution - they can be deleted by another process for example.
We can ignore that and return only files that really exist.

### DIFF
--- a/pkg/storages/swift/folder.go
+++ b/pkg/storages/swift/folder.go
@@ -101,6 +101,9 @@ func (folder *Folder) ListFolder() (objects []storage.Object, subFolders []stora
 			} else {
 				//It is a storage object name
 				obj, _, err := folder.connection.Object(folder.container.Name, objectName)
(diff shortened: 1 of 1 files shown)


pkg/storages/swift/folder.go
```diff
@@ -101,6 +101,9 @@ func (folder *Folder) ListFolder() (objects []storage.Object, subFolders []stora
 			} else {
 				//It is a storage object name
 				obj, _, err := folder.connection.Object(folder.container.Name, objectName)
+				if err == swift.ObjectNotFound {
```
usernamedt, line 104:
```suggestion
				// Some files can disappear during ListFolder execution - they can be deleted by another process
				// for example. We can ignore that and return only files that really exist.
				if err == swift.ObjectNotFound {
```

usernamedt (approved):
LGTM

PR #1131, opened 2021-10-20
Improve MySQL restore performance

ostinru:
This patch  improves wal-g-mysql backup-push/backup-fetch  by splitting single-stream backup into multi-stream backups:

In order to provide database-agnostic approach we don't make assumptions about backup-stream wal-g getting from stdin.
Instead, wal-g splits backup stream in blocks of `WALG_STREAM_SPLITTER_BLOCK_SIZE` bytes (in `storages/splitmerge/splitreader.go`). Then blocks are sent to one of `WALG_STREAM_SPLITTER_PARTITIONS` output streams. Each stream compressed, encrypted and uploaded separately.
Restore process is doing the opposite: it fetches multiple streams (utilize network by using multiple connections), then each stream can be decrypted and decompressed independently (utilize more CPU cores). Then it merges (in `storages/splitmerge/mergewriter.go`) all streams into one stream.

In my benchmarks `wal-g backup-fetch LATEST --turbo`:
* WAL-G VANILLA - produces 70 mbs of data written to ssd
* WAL-G patched (6 streams in blocks of 64kb) - produces 270mbs of data written to ssd


internal/stream_push_helper.go
```diff
@@ -24,6 +27,52 @@ func (uploader *Uploader) PushStream(stream io.Reader) (string, error) {
...
+		}(readers[i])
+	}
+
+	// Wait for upload finished:
+	var lastErr error
+	for i := 0; i < uploader.partitions; i++ {
+		err := <-errCh
+		if err != nil {
+			tracelog.WarningLogger.Printf("Failed to upload part of backup: %v", err)
+			lastErr = err
```
usernamedt, line 55: Why not to abort immediately on the first error? You can utilise the `sync.ErrGroup` to cancel all goroutines on receiving error in one of them.

pkg/storages/splitmerge/channelreader.go
```diff
@@ -0,0 +1,53 @@
+package splitmerge
```
usernamedt, line 1: As I can see, this functional is not storages-specific, why need to put it in `storages`? `internal/splitmerge` or `internal/ioextensions/splitmerge` looks like a good destination.

usernamedt (changes requested):

internal/stream_fetch_helper.go
```diff
@@ -51,9 +52,104 @@ func downloadAndDecompressStream(backup Backup, writeCloser io.WriteCloser) erro
...
+		return err
+	}
+
+	errorsPerWorker := make([]chan error, 0)
+	writers := splitmerge.MergeWriter(EmptyWriteIgnorer{WriteCloser: writeCloser}, partitions, blockSize)
+
+	for i := 0; i < partitions; i++ {
+		fileName := GetPartitionedStreamName(backup.Name, decompressor.FileExtension(), i)
+		errCh := make(chan error)
+		errorsPerWorker = append(errorsPerWorker, errCh)
```
usernamedt, line 84: again, maybe use the `sync.ErrGroup` instead?

internal/stream_metadata.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/wal-g/tracelog"
+)
+
+const (
+	SplitMergeStreamBackup   = "SPLIT_MERGE_STREAM_BACKUP"
+	SingleStreamStreamBackup = "STREAM_BACKUP"
+)
+
+type BackupStreamMetadata struct {
+	Type        string `json:"type"`
```
usernamedt, line 20: If this metadata does not consume much space, maybe just add it to the sentinel?

pkg/storages/splitmerge/splitmerge_test.go
```diff
@@ -0,0 +1,79 @@
...
+	var readers = SplitReader(dataReader, partitions, blockSize)
+
+	// out:
+	var sink bytes.Buffer
+	writers := MergeWriter(&sink, partitions, blockSize)
+
+	errCh := make(chan error)
+	defer close(errCh)
+	for i := 0; i < partitions; i++ {
+		go func(idx int, reader io.Reader, writer io.WriteCloser, buffSize int) {
```
usernamedt, line 33: `sync.ErrGroup`?

internal/uploader.go
```diff
@@ -155,3 +196,19 @@ func (uploader *Uploader) UploadMultiple(objects []UploadObject) error {
 	}
 	return nil
 }
+
+func (uploader *Uploader) ChangeDirectory(relativePath string) {
```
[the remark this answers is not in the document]
ostinru: `SetFolder` sounds like absolute path. Change Directory is like uniux `cd`

internal/stream_metadata.go
ostinru: it will increase coupling between uploader metadata and sentinel

pkg/storages/splitmerge/channelreader.go
mialinx: +1

internal/stream_metadata.go
```diff
@@ -0,0 +1,60 @@
...
+		}, nil
+	case SingleStreamStreamBackup, "":
+		return DownloadAndDecompressStream, nil
+	}
+	tracelog.ErrorLogger.Fatalf("Unknown backup type %s", metadata.Type)
+	return nil, nil // unreachable
+}
+
+func UploadBackupStreamMetadata(uploader UploaderProvider, metadata interface{}, backupName string) error {
+	sentinelName := MetadataNameFromBackup(backupName)
```
usernamedt, line 52: Well, this name interferes with the Postgres `metadata.json` file. Maybe name the file like it is done in the code? For example, `stream_metadata.json`.

internal/uploader.go
```diff
@@ -62,6 +77,32 @@ func NewUploader(
...
+	partitions int,
+	blockSize int,
+) UploaderProvider {
+	if partitions <= 1 {
+		// Fallback to old implementation in order to skip unneeded steps:
+		return NewUploader(compressor, uploadingLocation)
+	}
+
+	uploader := &SplitStreamUploader{
+		Uploader: &Uploader{
```
usernamedt, line 92: Just pass the `Uploader` to the `NewSplitStreamUploader` to avoid constructing the internal uploader.

perekalov (approved):

mialinx (approved):

usernamedt (approved):

PR #1139, opened 2021-11-04
Add pgbackrest backup-list and backup-fetch commands

Dm17r1y:
Write restoring pgbackrest backups by wal-g

Use `wal-g pgbackrest backup-list` to check current backups, made by pgbackrest
Use `wal-g pgbackrest backup-fetch pg_data backup_name` to restore backup, made by pgbackrest (for now works only with full backups)


internal/config.go
```diff
@@ -168,7 +170,8 @@ var (
 	}
 
 	PGDefaultSettings = map[string]string{
-		PgWalSize: "16",
+		PgWalSize:        "16",
+		PgbackrestStanza: "main",
```
usernamedt, line 174: I think that it is better to respect the original case, i.e. `PgBackRestStanza`.

internal/extract.go
```diff
@@ -55,6 +58,36 @@ func (err UnsupportedFileTypeError) Error() string {
 	return fmt.Sprintf(tracelog.GetErrorFormatter(), err.error)
 }
 
+type FileExtractor interface {
```
usernamedt, line 61: I see some existing logic duplication there. Why we can't use the regular `TarInterpreter`? As I can see, the main difference is only to append the relative path to the destination file path. I guess that we can modify the `tar.Header` and insert the relative path there instead of introducing the new `FileExtractor`.

internal/pgbackrest/pgbackrest_importer.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func getTime(timestamp int64) time.Time {
+	return time.Unix(timestamp, 0)
+}
+
+func getLsn(lsn string) (uint64, error) {
+	var first uint64
+	var second uint64
+	if _, err := fmt.Sscanf(lsn, "%x/%x", &first, &second); err != nil {
```
usernamedt, line 87: `pgx.ParseLSN`?

internal/pgbackrest/pgbackrest_settings.go
```diff
@@ -0,0 +1,147 @@
...
+const (
+	BackupPath        = "backup"
+	BackupInfoIni     = "backup.info"
+	BackupManifestIni = "backup.manifest"
+
+	BackupFolderName    = "backup"
+	BackupDataDirectory = "pg_data"
+)
+
+type BackupSettings struct {
```
usernamedt, line 18: What is the difference between `BackupSettings` and `PgbackrestBackupSettings`?

usernamedt (changes requested):

cmd/pg/pgbackrest.go
```diff
@@ -0,0 +1,31 @@
...
+
+func init() {
+	Cmd.AddCommand(pgbackrestCmd)
+}
+
+func configurePgbackrestSettings() (folder storage.Folder, stanza string) {
+	folder, err := internal.ConfigureFolder()
+	tracelog.ErrorLogger.FatalOnError(err)
+	stanza, ok := internal.GetSetting(internal.PgbackrestStanza)
+	if !ok {
```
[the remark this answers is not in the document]
Dm17r1y: Remove check

internal/compression/gzip/compressor.go
```diff
@@ -0,0 +1,26 @@
+package gzip
```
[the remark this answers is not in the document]
Dm17r1y: Move code to decompressor, also add simple compressor

internal/config.go
Dm17r1y: Rename PgbackrestStanza -> PgBackRestStanza

internal/pgbackrest/pgbackrest_importer.go
Dm17r1y: Use it insead of getLsn

internal/pgbackrest/pgbackrest_settings.go
Dm17r1y: Not so much differences, remove BackupSettings and use PgbackrestBackupSettings instead

internal/extract.go
Dm17r1y: Remove `FileExtractor` and use `tarInterpreter.Interpret(..., &tar.Header{...})` in raw files. Also add some info in `ReaderMarker` to understand: is that a tar file or regular file

docker/pg_tests/scripts/configs/s3cfg
```diff
@@ -0,0 +1,8 @@
+[default]
```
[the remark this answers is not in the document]
Dm17r1y: I found an another same `s3cfg` in mysql docker section. I think better use one `s3cfg` and put it in some common folder
Also I think better to install `s3cmd` [when building](https://github.com/wal-g/wal-g/blob/36c588e632a300028c9c8c7c6acb5cd7fef6aff2/docker/pg/Dockerfile#L15) `wal-g/pg` image: you dont need to reinstall s3cmd on every `go build` command

docker/pg_tests/scripts/tests/pgbackrest_backup_fetch_test.sh
```diff
@@ -0,0 +1,47 @@
...
+echo "archive_mode = on" >> ${PGDATA}/postgresql.conf
+echo "archive_command = '${archive_command}'" >> ${PGDATA}/postgresql.conf
+echo "archive_timeout = 600" >> ${PGDATA}/postgresql.conf
+
+/usr/lib/postgresql/10/bin/pg_ctl -D ${PGDATA} -w start
+/tmp/scripts/wait_while_pg_not_ready.sh
+
+mkdir -m 770 /tmp/pgbackrest-backups
+
+pgbackrest --stanza=main --pg1-path=${PGDATA} --repo1-path=/tmp/pgbackrest-backups stanza-create
```
usernamedt, line 24: Why not to use the `repo1-s3-*` pgbackrest settings instead of manual copy?

usernamedt (changes requested):

internal/extract.go
```diff
@@ -202,6 +216,23 @@ func ExtractAllWithSleeper(tarInterpreter TarInterpreter, files []ReaderMaker, s
 	return nil
 }
 
+func extractFile(tarInterpreter TarInterpreter, extractingReader io.Reader, fileClosure ReaderMaker) error {
+	switch fileClosure.FileType() {
```
usernamedt, line 220: I think this deserves a couple of comments, also I would rename `extractOne` to `extractOneTar` to avoid confusion.

docker/pg_tests/scripts/tests/pgbackrest_backup_fetch_test.sh
Dm17r1y: Pgbackrest requires minio with https. We can change minio settings and enable https or we can copy backup to minio with s3cmd

internal/extract.go
Dm17r1y: Rename and add comment

PR #1144, opened 2021-11-11
Introduce WALG_TAR_DISABLE_FSYNC to allow configuring wal-g to skip fsync

### LINKED ISSUE #1046: backup-fetch slowness : long (blocking) time spent on file.Sync()
### Database name
postrgesql

# Issue description

I was able to get a 2x speedup locally by skipping calls to `file.Sync()`.

These calls are currently blocking calls in the sequence of actions to extract a tar archive, it may be worth testing if parallelizing these calls can speed up things.

### Describe your problem

When tesing `backup-push / backup-fetch` cycles on my laptop, even on a reasonably small postgre database, the time taken by `backup-fetch` to restore a backup seemed suspiciously long.

In my test setup, wal-g was configured to backup/restore to/from a local directory, so no network issues are at hand here.

A profile on wal-g code pointed me to `file.Sync()` ; by skipping the `file.Sync()` call at the following call site (code version is v1.0, commit  2611d11da1) :

https://github.com/wal-g/wal-g/blob/2611d11da12c89f1ac6c20afb98acaa2d42e0ea0/internal/databases/postgres/tar_interpreter.go#L79

By skipping this call to `file.Sync()` :

```go
// replace :
    	err = file.Sync()
	return errors.Wrap(err, "Interpret: fsync failed")

// with :
	return nil
```

The total time spent on `wal-g backup-fetch` fell from a consistent 54s to a consistent 25s.

Obviously : when dealing with local files on my own laptop (which isn't as sollicitated as a production server), the resulting files (with or without `file.Sync()`) are the same.

Extra notes on my test setup to describe the timings above :

* restoring the postgresql backup results in a ~370M directory, which contains ~17k individual files (this gives an indication of the size on my test postgresql database)
* my local system is a linux : Ubuntu 20.10
* postgresql version is 13.2 (it may affect the number of files in a backup, the reported timing is `wal-g` only)
* I additionally used `lzma` for compression and pgp encryption, the 25s part will probably be way lower with a simpler setup (e.g : `lz4` and no encryption)

* take an existing database, or set up such a database
* set up `wal-g` for local backups : `export WALG_FILE_PREFIX=/path/to/backupdir`
* run `wal-g backup-push`
* run `wal-g backup-fetch targetdir LATEST`
* compare the time taken by `backup-fetch` with and without the call to `file.Sync()` mentioned above

I'm certainly not advocating to *remove* these calls to `file.Sync()`, however it looks like, the way the code is currently written, these calls are sequential, and their success.

It may be worth investigating if resolving these calls asynchronously (e.g: start extracting a new individual file from the tar archive while the previous one is `.Sync()`ed) could speed up the overall restore process.

serprex:
Disabling fsync can double write throughput

fsync for tar files is only used for restoring backups,
an unexpected system crash can always be remedied by restarting the restoration

This setting only affects postgres for now, as only the postgres logic calls fsync

Fixes #1046

### DIFF
--- a/utility/utility.go
+++ b/utility/utility.go
@@ -30,13 +30,15 @@ func LoggedClose(c io.Closer, errmsg string) {
 	}
 }
 
-func LoggedSync(file *os.File, errmsg string) {
-	err := file.Sync()
-	if errmsg == "" {
-		errmsg = "Problem with file sync"
-	}
-	if err != nil {
-		tracelog.ErrorLogger.Printf("%s: %v", errmsg, err)
+func LoggedSync(file *os.File, errmsg string, fsync bool) {
+	if fsync {
+		err := file.Sync()
+		if errmsg == "" {
+			errmsg = "Problem with file sync"
+		}
+		if err != nil {
+			tracelog.ErrorLogger.Printf("%s: %v", errmsg, err)
+		}
 	}
 }
 
--- a/docs/PostgreSQL.md
+++ b/docs/PostgreSQL.md
@@ -104,6 +104,10 @@ To configure base for next delta backup (only if `WALG_DELTA_MAX_STEPS` is not e
 
 To configure the size of one backup bundle (in bytes). Smaller size causes granularity and more optimal, faster recovering. It also increases the number of storage requests, so it can costs you much money. Default size is 1 GB (`1 << 30 - 1` bytes).
 
+* `WALG_TAR_DISABLE_FSYNC`
+
+Disable calling fsync after writing files when extracting tar files.
+
 * `WALG_PG_WAL_SIZE`
 
 To configure the wal segment size if different from the postgres default of 16 MB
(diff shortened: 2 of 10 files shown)


utility/utility.go
```diff
@@ -31,6 +32,9 @@ func LoggedClose(c io.Closer, errmsg string) {
 }
 
 func LoggedSync(file *os.File, errmsg string) {
+	if viper.GetBool("WALG_DISABLE_FSYNC") {
```
serprex, line 35: Couldn't import `internal.DisableFsyncSetting` because it'd create an import loop

Other option would be to pass the setting as a parameter. Arguably this bool could be passed down through all these calls rather than repeatedly calling `viper.GetBool`

I was also noticing that `LoggedSync` can be redundant, since `writeLocalFile`  is used by `catchup_file_unwrapper.go` which also calls `file.Sync`. So if the default file unwrapper also called `file.Sync` this defer wouldn't be needed in tar_interpreter_new, leaving it only used by `pagefile.go`

usernamedt:
I guess that this will work fine for the full backups, but there might be some problems with the delta backups since WAL-G do read some previously fetched paged files sometimes (for example, in [RestoreMissingPages](https://github.com/wal-g/wal-g/blob/0261341f03bafb48960dbb5835d2b921cbb92b14/internal/databases/postgres/pagefile_new.go#L149)).

serprex:
@usernamedt under normal circumstances os should be making fsync transparent (ie if crash occurs during restore, then maybe delta needs to restart from last checkpoint), is the issue that for delta restore the sync is being used as a flush?

usernamedt (approved):

PR #1147, opened 2021-11-15
all: distinguish between storage errors and no backups

mialinx:
Scripts should be able to distinguish between absence of backup and real storage/configurations errors.

In case of real errors - exit code should be 1.


usernamedt (commented):
Maybe it would be more natural to return the empty backups list on no backups found? For example, if called like 'wal-g backup-list --json' it seems to be more natural to return an empty list like '[]'

PR #1149, opened 2021-11-17
more straightforward decompression logic

mialinx:
`Decompressor` interface is poorly designed.
Typical design pattern for decryption, decrompression is wrapping streams.
So `Decompressor` should receive `io.Reader` (with compressed data) and return `io.Reader` (with decompressed data)


usernamedt:
For some reason, `lzo_test.go` was not in the `make unittest` target. Added the little commit to fix it.

internal/extract.go
```diff
@@ -214,9 +218,12 @@ func tryExtractFiles(files []ReaderMaker,
 		}()
 		go func() {
 			defer writingSemaphore.Release(1)
+			defer utility.LoggedClose(extractingReader, "")
 			err := extractOne(tarInterpreter, extractingReader)
+			if err == nil {
+				err = readTrailingZeros(extractingReader)
```
usernamedt, line 224:
```suggestion
				// Fixes trailing bytes https://github.com/wal-g/wal-g/issues/22
				err = readTrailingZeros(extractingReader)
```

internal/extract.go
```diff
@@ -242,3 +249,21 @@ func tryExtractFiles(files []ReaderMaker,
 	})
 	return failed
 }
+
+func readTrailingZeros(r io.Reader) error {
+	b := make([]byte, 1024*1024)
```
serprex, line 254: Why 1MB?

internal/extract.go
```diff
@@ -242,3 +249,27 @@ func tryExtractFiles(files []ReaderMaker,
...
+		}
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
+		if len(b) < utility.CompressedBlockMaxSize {
+			// but if we found zeroes, allocate large buffer to speed up reading
+			b = make([]byte, utility.CompressedBlockMaxSize)
```
serprex, line 272: This condition can fire multiple times, so it's possible to read much more than `utility.CompressedBlockMaxSize`

A `totalBytesRead` counter should be kept, breaking from loop when it reaches `utility.CompressedBlockMaxSize`

Then `utility.CompressedBlockMaxSize - totalBytesRead` would be the allocation size

mialinx: `b` allocated only two times: first with `1024` len, second - with `utility.CompressedBlockMaxSize` len.
Golang zeroes `b` during allocation.  `b`'s len is not affected by `Read` calls.

As we only need to check any non-zero byte, we may omit typical golang Read pattern and always read to the begining of the `b`.

Looks safe to me..

serprex: Yes, the issue I was considering was the reader giving off more zeroes than `utility.CompressedBlockMaxSize`

PR #1154, opened 2021-11-19
Add `--until-binlog-last-modified-time` option to `wal-g-mysql binlog…

ostinru:
Add `--until-binlog-last-modified-time` option to `wal-g-mysql binlog-replay`  this may be useful to achieve exact clones of the same database in scenarios when  new binlogs are uploaded during replay process of different hosts.

### DIFF
--- a/cmd/mysql/binlog_fetch.go
+++ b/cmd/mysql/binlog_fetch.go
@@ -12,9 +12,12 @@ import (
 
 const fetchSinceFlagShortDescr = "backup name starting from which you want to fetch binlogs"
 const fetchUntilFlagShortDescr = "time in RFC3339 for PITR"
+const fetchUntilBinlogLastModifiedFlagShortDescr = "time in RFC3339 that is used to prevent wal-g from replaying" +
+	" binlogs that was created/modified after this time"
 
 var fetchBackupName string
 var fetchUntilTS string
+var fetchUntilBinlogLastModifiedTS string
 
 // binlogPushCmd represents the cron command
 var binlogFetchCmd = &cobra.Command{
@@ -24,7 +27,7 @@ var binlogFetchCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		folder, err := internal.ConfigureFolder()
 		tracelog.ErrorLogger.FatalOnError(err)
-		mysql.HandleBinlogFetch(folder, fetchBackupName, fetchUntilTS)
+		mysql.HandleBinlogFetch(folder, fetchBackupName, fetchUntilTS, fetchUntilBinlogLastModifiedTS)
 	},
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		internal.RequiredSettings[internal.MysqlBinlogDstSetting] = true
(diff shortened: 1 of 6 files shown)


cmd/mysql/binlog_fetch.go
```diff
@@ -24,7 +27,7 @@ var binlogFetchCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		folder, err := internal.ConfigureFolder()
 		tracelog.ErrorLogger.FatalOnError(err)
-		mysql.HandleBinlogFetch(folder, fetchBackupName, fetchUntilTS)
+		mysql.HandleBinlogFetch(folder, fetchBackupName, fetchUntilTS, fetchUntilBinlogLastModifiedTS)
```
mialinx, line 30: what's the difference between fetchUntilTS and fetchUntilBinlogLastModifiedTS ?
can we use fetchUntilTs instead of fetchUntilBinlogLastModifiedTS

internal/databases/mysql/mysql.go
```diff
@@ -271,6 +271,9 @@ func getLogsCoveringInterval(folder storage.Folder, start time.Time, includeStar
 	})
 	var logsToFetch []storage.Object
 	for _, logFile := range logFiles {
+		if logFile.GetLastModified().After(endBinlogTS) {
+			continue // don't fetch binlogs from future
+		}
```
mialinx, line 276: This shold not affect the final result, as mysqbinlog started as
```
mysqlbinlog --stop-datetime="$WALG_MYSQL_BINLOG_END_TS"
```
and should not replay transactions happened after `fetchUntilTS`

mialinx (commented):
didn't understand how this PR fixes the problem

internal/databases/mysql/mysql.go
ostinru: This patch fixes following issue:

Live MySQL cluster **A** constantly uploads binlogs. We are trying to restore new cluster (let's call it **B**) from S3 on some point in time, e.g. `NOW()` (or future time).
Then we can have following scenario:
'Fast' servers restore MySQL on Snapshot + binlog.123 (Binlog 123 is the latest observed binlog for 'fast' hosts).
'Slow' servers may restore MySQL on Snapshot+binlog.124 (Becase cluster **A** have uploaded one more binlog during 'slow' restore).
As a result we have different GTIDs in a cluster. It is easier to prevent such scenario rather than fixing it afterwads.

To prevent different GTID-sets we should:
 * introduce hard limits on which binlogs we are replaying
 * point in time for this limits should be `NOW()` or in past.

cmd/mysql/binlog_fetch.go
ostinru: Here I am introducing new  should distinguish two timestamps:
`fetchUntilTS` - is about time of individual transactions
`fetchUntilBinlogLastModifiedTS` - is about time when binlog was created

There are 2 properties:
`fetchUntilTS < fetchUntilBinlogLastModifiedTS` - because transactions dumped to binlog _before_ binlog upload
'--stop-datetime < fetchUntilBinlogLastModifiedTS' - it is new rule that fixes multi-host restores (check detailed explanation in another comment)

mialinx (approved):

PR #1157, opened 2021-11-22
SQL Server - skipping logs that already been applied

trootnev:
### Database name
SQL Server

# Pull request description

### Describe what this PR fix
Image the case - db is backup is restored, some logs applied and then something bad happens and you're stuck. To proceed with logs restore you basically have to drop database and start over. If you try to run logs-restore again you'll get and error, as wal-g will try to apply all the logs since last full backup.
Additional check can mitigate this. Check the LSNs of the log backup and the database in the 'restoring' state and skipping those whose LSN is somewhere in the past for the current sate of the recovery.

Restore database to a point in time with norecovery, then try to continue restoring further.


mialinx (approved):
LGTM

internal/databases/sqlserver/sqlserver.go
```diff
@@ -551,3 +552,30 @@ func RunOrReuseProxy(ctx context.Context, cancel context.CancelFunc, folder stor
 	}
 	return &LockWrapper{lock}, nil
 }
+
+func GetDBRestoreLSN(db *sql.DB, databaseName string) (int64, error) {
+	query := fmt.Sprintf(`SELECT MAX(redo_start_lsn) 
+							FROM sys.master_files
+							WHERE database_id=DB_ID('%s') 
+							`, databaseName)
```
ostinru, line 560: It is looks like a potential SQL Injections.
I would propose:
`
query := "SELECT MAX(redo_start_lsn)  FROM sys.master_files WHERE database_id=DB_ID(?)"
db.QueryRow(query, databaseName).Scan(...)
`

ostinru (approved):

PR #1161, opened 2021-11-26
sqlserver compression / encryption

mialinx:
### SQLServer

Enables WAL-G standard compression and encryption methods for SQLServer.
Option `WALG_COMPRESSION_METHOD=sqlserver`  has special meaning: WAL-G will request SQLServer to compress data itself.

### DIFF
--- a/internal/databases/sqlserver/blob/server.go
+++ b/internal/databases/sqlserver/blob/server.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/wal-g/wal-g/internal/compression"
+	"github.com/wal-g/wal-g/internal/crypto"
 	"golang.org/x/xerrors"
 
 	"github.com/gofrs/flock"
@@ -40,6 +42,11 @@ type Server struct {
 	leasesMutex  sync.Mutex
 	downloadSem  chan struct{}
 	uploadSem    chan struct{}
+	compression  string
+	compressor   compression.Compressor
+	decompressor compression.Decompressor
+	encryption   string
+	crypter      crypto.Crypter
 }
 
 func NewServer(folder storage.Folder) (*Server, error) {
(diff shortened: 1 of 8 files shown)


internal/databases/sqlserver/blob/server.go
```diff
@@ -72,6 +80,21 @@ func NewServer(folder storage.Folder) (*Server, error) {
 	bs.server = http.Server{Addr: bs.endpoint, Handler: bs}
 	bs.indexes = make(map[string]*Index)
 	bs.leases = make(map[string]Lease)
+	compressor, err := internal.ConfigureCompressor()
+	if err != nil {
+		if _, ok := err.(internal.UnknownCompressionMethodError); !ok {
```
usernamedt, line 85: Maybe add some warning log to notify the user that WAL-G failed to configure the compressor?

internal/databases/sqlserver/blob/server.go
```diff
@@ -698,3 +742,13 @@ func (bs *Server) AcquireLock() (io.Closer, error) {
 	}
 	return lock, nil
 }
+
+func (bs *Server) validateBlobCompressionEncryption(idx *Index) error {
+	if idx.Encryption != bs.encryption {
+		return fmt.Errorf("blob encryption (%s) does not match configured (%s)", idx.Encryption, bs.encryption)
+	}
+	if idx.Compression != bs.compression {
```
usernamedt, line 750: Let's assume the following case: Index with some encryption configured but without compression. Server has both encryption and compression enabled. Encryption check is passed, but compressors are obviously mismatched. In this setup, WAL-G will fail to download the uncompressed encrypted file. Maybe allow this scenario (with some warnings)?

Also, maybe go even further and allow all currently supported compressors without the requirement that it should match the currently configured one?

mialinx: Changing compression algorithm - one part of the problem. The second one is changing encryption.
The proper solution is to keep metainfo with files (as file extension or in some JSON) and **choose** decompressor/crypter by it's name. But it's not possible for the crypter at the moment.

I wouldn't like to solve only half of the problem, especially it's pretty hypothetical for the SQLServer.

internal/databases/sqlserver/blob/server.go
mialinx: Agree.
Will return error unless **special** builtin compression method specified.

usernamedt (approved):
LGTM

ostinru (approved):
no criminality

PR #1173, opened 2021-12-09
sqlserver: fix blob index race

mialinx:
After creating new blob with PUT, SQLServer checks it with HEAD.
If storage is not read-after-write capable, we could get NotFound error and return 404 to SQLServer. It will break backup process.

Fix: cache newly created blob indexes in in-memory cache

### DIFF
--- a/internal/databases/sqlserver/blob/server.go
+++ b/internal/databases/sqlserver/blob/server.go
@@ -624,12 +624,15 @@ func (bs *Server) HandleBlobPut(w http.ResponseWriter, req *http.Request) {
 			return
 		}
 	}
+	bs.indexesMutex.Lock()
+	defer bs.indexesMutex.Unlock()
 	err = idx.Save()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/databases/sqlserver/blob/server.go
```diff
@@ -624,12 +624,15 @@ func (bs *Server) HandleBlobPut(w http.ResponseWriter, req *http.Request) {
 			return
 		}
 	}
+	bs.indexesMutex.Lock()
+	defer bs.indexesMutex.Unlock()
```
ostinru, line 628: It is seems that we can get this lock after `idx.Save` execution.

ostinru (approved):

internal/databases/sqlserver/blob/server.go
mialinx: We would like to be more conservative here.
And prevent succesful GET requests before PUT accomplished.

Issue #1106, opened 2021-10-02
wal-g 1.10 remote backup for postgreSQL 13 misses files


t83714:
### Database name
wal-g version: 1.10
postgreSQL version: 13 (using docker image `bitnami/postgresql:13.3.0-debian-10-r79` )
backup method: remote backup via postgres BASE_BACKUP protocol
Storage: Google cloud storage

# Issue description

For the same database instance, the backup created remotely is missing files and causing `ERROR:  could not open file "base/13397/17003": No such file or directory` error after the recovery on a new server (same configuration).

Please check the [dev_basebackups_005_base_000000010000000000000762_backup_stop_sentinel.json.zip](https://github.com/wal-g/wal-g/files/7270958/dev_basebackups_005_base_000000010000000000000762_backup_stop_sentinel.json.zip) file.

The backup did miss `base/13397/17003` and only contains the following:

```json
   "base/13397/17003.9": {
      "IsIncremented": false,
      "IsSkipped": false,
      "MTime": "2021-10-01T10:18:07Z",
      "UpdatesCount": 0
    },
    "base/13397/17003_fsm": {
      "IsIncremented": false,
      "IsSkipped": false,
      "MTime": "2021-10-01T10:18:07Z",
      "UpdatesCount": 0
    },
    "base/13397/17003_vm": {
      "IsIncremented": false,
      "IsSkipped": false,
      "MTime": "2021-10-01T10:18:07Z",
      "UpdatesCount": 0
    }
```

On the the other hand, the backup made locally does contain `base/13397/17003` and didn't cause any recovery issue.

Please check [dev_basebackups_005_base_000000010000000700000067_backup_stop_sentinel.json.zip](https://github.com/wal-g/wal-g/files/7270963/dev_basebackups_005_base_000000010000000700000067_backup_stop_sentinel.json.zip).

The backup contains all files the remote backup has plus missing files:

```json
     "/base/13397/17003": {
      "IsIncremented": false,
      "IsSkipped": false,
      "MTime": "2021-10-02T05:31:27.39205458Z",
      "UpdatesCount": 0
    },
    "/base/13397/17003.1": {
      "IsIncremented": false,
      "IsSkipped": false,
      "MTime": "2021-10-01T09:15:05.430237269Z",
      "UpdatesCount": 0
    },
    "/base/13397/17003.2": {
      "IsIncremented": false,
      "IsSkipped": false,
      "MTime": "2021-10-01T09:17:01.320815142Z",
[...]

Issue #1078, opened 2021-08-25
wal-show command errors out with: Incorrect logSegNoLo in WAL file name


siepkes:
### Database name

PostgreSQL

# Issue description

### Describe your problem

The `wal-show` command runs in to an "Incorrect logSegNoLo in WAL file name" error when trying to parse the name of the base backup:

```
/opt/walg/wal-g --config /etc/walg/walg.json wal-show
ERROR: 2021/08/25 14:26:28.706255 Failed to add backups info: Incorrect logSegNoLo in WAL file name: 00000019000000000000177F
```

Ultimately the check here triggers the error: https://github.com/wal-g/wal-g/blob/cadf598e1c2a345915a21a44518c5a4d5401e2e3/internal/databases/postgres/timeline.go#L134-L137

If I deduced it correctly `xLogSegmentsPerXLogID` is always `256`. The value of `logSegNoLo` would be way higher then that with a base backup file name `00000019000000000000177F`. 

I'm a bit confused by `logSegNoLo >= xLogSegmentsPerXLogID` since that seems like a sane check for an actual WAL but not for an base backup?

### Please provide steps to reproduce

Make a full backup with `wal-g backup-push`

### Please add config and wal-g stdout/stderr logs for debug purpose

```
----8<---------------------
DEBUG: 2021/08/25 14:03:03.396137 WALG_USE_REVERSE_UNPACK=false
DEBUG: 2021/08/25 14:03:03.396141 WALG_USE_WAL_DELTA=false
DEBUG: 2021/08/25 14:03:03.396145 WALG_VERIFY_PAGE_CHECKSUMS=false
DEBUG: 2021/08/25 14:03:03.396149 YC_CSE_KMS_KEY_ID=
DEBUG: 2021/08/25 14:03:03.396153 YC_SERVICE_ACCOUNT_KEY_FILE=
DEBUG: 2021/08/25 14:03:03.396157 _=/opt/walg/wal-g
DEBUG: 2021/08/25 14:03:03.740582 Selected crypter: Opengpg/Crypter
DEBUG: 2021/08/25 14:03:03.893006 Selected crypter: Opengpg/Crypter
ERROR: 2021/08/25 14:03:04.756975 Failed to add backups info: Incorrect logSegNoLo in WAL file name: 00000019000000000000177F
github.com/wal-g/wal-g/internal/databases/postgres.newIncorrectLogSegNoError
        /root/wal-g-fork/internal/databases/postgres/timeline.go:40
[...]

PR #1175, opened 2021-12-10
Improve and fix remote backup

sebasmannem:
### Database name
Postgres

# Pull request description

### Describe what this PR fix
There are two main issues that this PR will address:

- #1106 
- #1078 

More info in the comments of these issues...

### DIFF
--- a/internal/databases/postgres/tarball_streamer.go
+++ b/internal/databases/postgres/tarball_streamer.go
@@ -190,7 +190,7 @@ func (streamer *TarballStreamer) readFileData() (err error) {
 		return nil
 	}
 	if err == io.EOF && streamer.bufDataSize > 0 {
-		// stream reached end of file. Bytes where read, but . Let's ignore on this pass.
+		// stream reached end of file. Bytes where read, but let's ignore on this pass.
 		return nil
 	}
 	return err
--- a/internal/databases/postgres/tarball_streamer.go
+++ b/internal/databases/postgres/tarball_streamer.go
@@ -41,7 +41,7 @@ func NewTarballStreamerRemap(from string, to string) (tsr *TarballStreamerRemap,
 // TarballStreamer is used to modify tar files which are received streaming.
 // Two modifications are:
 // * remap: change (some of) the paths for files in the tar file, and
-// * tee: copy some of the files to a second tar file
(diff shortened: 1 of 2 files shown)


internal/databases/postgres/tarball_streamer.go
```diff
@@ -188,7 +190,7 @@ func (streamer *TarballStreamer) readFileData() (err error) {
 		return nil
 	}
 	if err == io.EOF && streamer.bufDataSize > 0 {
-		// stream reached end of file. Bytes where read, but . Lets ignore on this pass.
+		// stream reached end of file. Bytes where read, but . Let's ignore on this pass.
```
serprex, line 193:
```suggestion
		// stream reached end of file. Bytes where read, but let's ignore on this pass.
```

internal/databases/postgres/tarball_streamer.go
```diff
@@ -91,24 +91,22 @@ func NewTarballStreamer(input io.Reader, maxTarSize int64, bundleFiles BundleFil
...
 	// First output tar, or switching to next
 	if streamer.outputTar == nil {
 		streamer.outputTar = tar.NewWriter(streamer.outputIo)
 		streamer.tarFileReadIndex = 0
 		if streamer.curHeader != nil {
-			err = streamer.outputTar.WriteHeader(streamer.curHeader)
-			if err != nil {
-				return err
-			}
+			return streamer.addFile()
```
serprex, line 101: Can you expand on why this is necessary? I see `addFile` starts with WriteHeader, so what was causing the rest to sometimes be necessary but usually safe to skip?

sebasmannem: We should always use addFile, because that also adds to the Tee file if needed and all.
But the bigger issue why this was breaking was there was no return, meaning we continued in this method, which does a streamer.inputTar.Next() skipping the body of the current file, and moving straight to the next file...

serprex (approved):

sebasmannem:
I have tested a backup and restore with the pg12 bitnami image wit a database with larger files and ran into both situations, where multiple tar files where created, and where a file is larger then the tar size. 
The restore worked fine.
We could do a beta build and have it tested with the cluster where the original issue was at hand.

PR #1177, opened 2021-12-16
Fix missing unit tests

usernamedt:
Currently, there are some explicit paths configured in the `make unittests` Makefile target. This results in some missing tests. I propose to run all of the `_test.go` files by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,25 +188,9 @@ st_integration_test: load_docker_common
 
 unittest:
 	go list ./... | grep -Ev 'vendor|submodules|tmp' | xargs go vet
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/compression/
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/crypto/openpgp/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -188,25 +188,9 @@ st_integration_test: load_docker_common
 
 unittest:
 	go list ./... | grep -Ev 'vendor|submodules|tmp' | xargs go vet
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/compression/
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/crypto/openpgp/
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/crypto/awskms/
-	go test -mod vendor -v $(TEST_MODIFIER) -tags "$(BUILD_TAGS)" ./internal/abool
-	@if [ ! -z "${USE_LIBSODIUM}" ]; then\
```
serprex, line 196: Is testing with USE_LIBSODIUM disabled no longer supported, or does `...` skip when disabled?

usernamedt: Yes, it should skip these tests since they have the `// +build libsodium` tag.

usernamedt:
Looks like that I've already fixed the LZO issue [in this PR](https://github.com/wal-g/wal-g/pull/1149#issuecomment-975572548), but it accidentally got overwritten :) Now unit tests should pass.

serprex (approved):

PR #1179, opened 2021-12-17
Update brotli to v1.0.9 from v1.0.7

serprex:
xxHash removed as it's no longer required after updating lz4 earlier


x4m (approved):

x4m:
BTW maybe one day we can consider Zstd again? Back in 2019 we removed it due to data corruption issues but since then they fixed some bugs...

PR #1181, opened 2021-12-18
DecryptAndDecompressTar: return ReadCloser instead of taking Writer

serprex:
This allows tryExtractFiles to drop io.Pipe in favor of feeding the decompression reader directly into the tar interpreter

In production we've seen a few databases which'll continuously fail to download some parts from their base backups. This issue persists across base backups & forks

This issue has been reproduced across:
- Before & after decompression refactor
- With lz4, lz4/v4, & brotli
- Before & after azure sdk update

The issue is currently mitigated using curl & tar to complete missing parts

After this change the issue was resolved

### DIFF
--- a/internal/extract.go
+++ b/internal/extract.go
@@ -38,14 +38,6 @@ type UnsupportedFileTypeError struct {
 	error
 }
 
-type DecompressionError struct {
-	error
-}
-
-func newDecompressionError(err error) DecompressionError {
-	return DecompressionError{err}
-}
-
 func newUnsupportedFileTypeError(path string, fileFormat string) UnsupportedFileTypeError {
 	return UnsupportedFileTypeError{errors.Errorf("WAL-G does not support the file format '%s' in '%s'", fileFormat, path)}
 }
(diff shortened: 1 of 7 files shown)


internal/extract.go
```diff
@@ -218,34 +171,30 @@ func tryExtractFiles(files []ReaderMaker,
 			tracelog.ErrorLogger.Println(err)
 			return files //Should never happen, but if we are asked to cancel - consider all files unfinished
 		}
-		err = writingSemaphore.Acquire(downloadingContext, 1)
```
usernamedt, line 221: Why did you remove the `writingSemaphore` logic?

usernamedt (changes requested):

internal/extract.go
serprex: There is now only 1 goroutine which decompresses & writes, instead of sending decompresssed bytes to pipe which gets written

usernamedt: > There is now only 1 goroutine which decompresses & writes, instead of sending decompresssed bytes to pipe which gets written
> 
> These are all still capped by a `semaphore.NewWeighted(int64(downloadingConcurrency))`

Oh, looks like I've missed it...

usernamedt (approved):

PR #1184, opened 2021-12-20
sqlserver: backup import and export to the external storage

mialinx:
### SQLServer

Export/import backups to/from external storage.

Backups blobs are stored as a set of blocks with index in default storage.
During export to external storage, backup blobs are serialized to single .bak files, suitable for downloading and manual restoring using `RESTORE DATABASE FROM DISK='filename.bak'`.
During import from external storage, backup blobs are sliced into the set of blocks appropriate to be restored with wal-g.

### DIFF
--- a/internal/databases/sqlserver/backup_export_handler.go
+++ b/internal/databases/sqlserver/backup_export_handler.go
@@ -1,0 +1,116 @@
+package sqlserver
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+	"syscall"
+
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+
+	"github.com/wal-g/tracelog"
+	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/utility"
+)
+
+func prepareBackupExportSpec(d map[string]string) (dbnames []string, mapping map[string]string) {
+	mapping = make(map[string]string, len(d))
+	for k, v := range d {
+		dbnames = append(dbnames, k)
+		if v == "" {
+			v = k
+		}
+		mapping[k] = v
+	}
+	return
+}
+
+func HandleBackupExport(externalConfig string, exportPrefixes map[string]string) {
+	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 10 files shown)


internal/databases/sqlserver/blob/server.go
```diff
@@ -603,36 +604,52 @@ func (bs *Server) HandleBlobPut(w http.ResponseWriter, req *http.Request) {
...
+					return
+				}
+			}
+			if n == 0 {
+				continue
+			}
+			id := fmt.Sprintf("data_%05d", i)
+			name := idx.PutBlock(id, uint64(n))
+			encryptedReader := internal.CompressAndEncrypt(bytes.NewReader(buf[:n]), bs.compressor, bs.crypter)
+			err = folder.PutObject(name, encryptedReader)
```
ostinru, line 630: I am just curious about why do you split blob into 4Mb chunks.

internal/databases/sqlserver/backup_export_handler.go
```diff
@@ -0,0 +1,116 @@
...
+	tracelog.ErrorLogger.FatalOnError(err)
+	defer lock.Close()
+
+	var backupName string
+	var sentinel *SentinelDto
+	backup, err := internal.GetBackupByName(internal.LatestString, utility.BaseBackupPath, folder)
+	tracelog.ErrorLogger.FatalfOnError("can't find latest backup: %v", err)
+	backupName = backup.Name
+	sentinel = new(SentinelDto)
+	err = backup.FetchSentinel(&sentinel)
```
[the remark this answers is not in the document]
mialinx: I don't.  But it works too: https://go.dev/play/p/_yTrVgnKpXq

internal/databases/sqlserver/blob/server.go
mialinx: SQLServer uses 4Mb blocks by itself. We mimic.

internal/databases/sqlserver/blob/server_test.go
```diff
@@ -0,0 +1,47 @@
+package blob
+
+/*
```
serprex, line 3: Why commented?

mialinx: Sorry, an unnecessary file was added to commit.  Removed it.

ostinru (approved):
LGTM

PR #1191, opened 2021-12-30
Update to azure-sdk-for-go 0.3.0 with fixed ListBlobsHierarchy

serprex:
There's a particularly important fix:
https://github.com/Azure/azure-sdk-for-go/pull/16354 fixes ListBlobsHierarchy when paging is necessary
which turns out to be a broken fix, so a fork is used in the meantime which includes a proper fix:
https://github.com/Azure/azure-sdk-for-go/pull/16992

On some containers paging happens much sooner than MaxResults would imply,
causing restoring LATEST to either fail because wal-g can't find backups, or wal-g selects an old backup

Two other notable PRs:
https://github.com/Azure/azure-sdk-for-go/pull/16067 fixes a chunk pooling bug
https://github.com/Azure/azure-sdk-for-go/pull/15958 removes the need for FakeIoReadCloser


serprex:
Looks like a release should happen soon: https://github.com/Azure/azure-sdk-for-go/pull/16827

Will update PR when that happens, & wait until then

serprex:
@x4m turns out the paging bug's priority is higher since it can crop up in containers with only a few blobs

serprex:
azure-sdk-for-go has released 0.3.0, will update this PR soon

pkg/storages/azure/folder.go
```diff
@@ -91,10 +90,10 @@ func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder,
 		environmentName = defaultEnvName
 	}
 
-	var credential azcore.Credential
+	var credential *azblob.SharedKeyCredential
 	var err error
 	if usingToken {
-		credential = azcore.NewAnonymousCredential()
+		credential = nil
```
usernamedt, line 96: `credential` is already `nil`, isn't it?

usernamedt (approved):

pkg/storages/azure/folder.go
serprex: Yes. I'm used to languages requiring assignment before use

Issue #1082, opened 2021-09-03
Cannot use latest (v1) release as code dependency in golang


maboehm:
# Issue description
Hi, I'm importing "github.com/wal-g/wal-g/utility" in my code, and cannot update the dependency to the latest tags, since they are not complete semVer (i.e. `v1.1` instead of `v1.1.0`, see also https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions)

### Please provide steps to reproduce
a minimal example:
```
$ cat go.mod           
module example.com/test
go 1.16
require github.com/wal-g/wal-g v1.1

$ go mod tidy                       
go: errors parsing go.mod:
/tmp/tmp.07rCLs2fep/go.mod:5: no matching versions for query "v1.1"
```
similarly:
```
$ go get github.com/wal-g/wal-g@v1.1
go get github.com/wal-g/wal-g@v1.1: no matching versions for query "v1.1"
```

Can you please in the future use complete major.minor.patch tags for releases? Thank you

usernamedt:
Hi! Thanks, this looks like a good idea. I think we have no reasons not to use this approach.

usernamedt:
> @DaniilKurlovich can you create an additional tag `v1.1.0` that points to the `v1.1` version?

I've created an [additional v1.1.0 tag](https://github.com/wal-g/wal-g/releases/tag/v1.1.0).

usernamedt:
Looks like the extra tag is not an option since it results in failing builds. I've removed the `v1.1.0` tag.

PR #1192, opened 2022-01-03
Fix broken build by ensuring a single unique version is returned.

darora:
### Database name
n/a

# Pull request description

[v1.1.0](https://github.com/wal-g/wal-g/releases/tag/v1.1.0) was created as a tag pointing to
cadf598e1c2a345915a21a44518c5a4d5401e2e3 , which already has tag v1.1

This results in multiple version strings being returned during the build process,
which breaks the build.

git checkout v1.1 # or git checkout v1.1.0
make deps
make pg_build

```                              
...:/tmp/wal-g$ USE_LIBSODIUM=true make pg_build                                                                                                       
(cd main/pg && go build -mod vendor -tags "brotli libsodium" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/c
md/pg.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/pg.gitRevision=`git rev-parse --short 
HEAD` -X github.com/wal-g/wal-g/cmd/pg.walgVersion=`git tag -l --points-at HEAD`")
# github.com/wal-g/wal-g/main/pg
usage: link [options] main.o 
  -B note
        add an ELF NT_GNU_BUILD_ID note when using ELF
  -E entry
[...]
```

Underlying cause:
```
/tmp/wal-g# git tag -l --points-at HEAD
v1.1
v1.1.0
/tmp/wal-g# git describe --tags --always --first-parent 
v1.1.0
```

Alternatives: delete the tag, or select one of the entries returned by the current git invocation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,8 @@ GOLANGCI_LINT_VERSION ?= "v1.37.0"
 REDIS_VERSION ?= "5.0.8"
 TOOLS_MOD_DIR := ./internal/tools
 
+WALG_VERSION := $(shell git describe --tags --always --first-parent)
+
 BUILD_TAGS:=brotli
 
 ifdef USE_LIBSODIUM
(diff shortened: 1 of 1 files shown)


usernamedt:
Since this PR would not fix the failing build I think it is reasonable to delete the extra tag [discussed in this issue](https://github.com/wal-g/wal-g/issues/1082).

PR #1193, opened 2022-01-03
Postgres: Remove old file unwrap logic

serprex:
It's been slated for removal for a year & a half now

I am seeing some new error messages: when a part fails to download, Sync/Close log that they failed because "file already closed", tracked that only caller of `writeLocalFile` is `unwrapRegularFile`, & removed Close from `writeLocalFile`


serprex:
Trying to read through what separates these implementations, but seems it the new implementation doesn't work for non-reverse deltas. Closing for now, but it'd be good to simplify extraction logic & getting rid of this duplicity would be the first step

Issue #1126, opened 2021-10-18
Errors during wal-fetch return exit code 1


fischerman:
### Database name

Postgres

# Issue description

### Describe your problem

Postgres uses `restore_command` to fetch WAL segments from the archive. This is implemented by `wal-g wal-fetch`. If `restore_command` fails, Postgres distinguishes two cases:

1. Postgres has reached the end of the timeline (no more WAL segments to process, i.e. recovery target time is set to a point in the future). Postgres completes recovery.
2. The restore_command received a signal or failed to execute (shell error). The recovery fails.

The way that Postgres distinguishes those cases is not well documented. The best explanation is in the code (e.g. see [code](https://github.com/postgres/postgres/blob/96b665083eb72570e226cf2d25c960b3acc62040/src/backend/access/transam/xlogarchive.c#L233) or [StackOverflow](https://stackoverflow.com/a/40244365/1863595)). TL;DR For most exit codes incl. exit code 1 Postgres assumes case 1.

WAL-G behaves correctly in this case:
```
# wal-g wal-fetch 000000020000000000000003 /tmp/test1
ERROR: 2021/10/18 12:07:12.891104 Archive '000000020000000000000003' does not exist.
# echo $?
1
```

However, if WAL-G is misconfigured or there are any network related errors, it will return the same exit code.
```
# wal-g wal-fetch 001 001
ERROR: 2021/10/18 13:57:27.599022 Failed to find any configured storage
# echo $?
1
```

Postgres will assume that it has reached the end of the timeline and will complete recovery. I think WAL-G should indicate to Postgres that the WAL fetching failed by mimicking a shell error.

sebasmannem:
@fischerman , this is an awesome question, as it is architectural, practical and philosophical.
So, this will be a verbose answer, still trying to find if and what we should do.

My first comment is that this issue is not exclusive to wal-g. As an example, [rsync](https://linux.die.net/man/1/rsync) (scroll down or search for Exit Values), also does not have exit codes between 126 and 255.
Furthermore, just looking for a generic approach at exit codes I ran into [this](https://tldp.org/LDP/abs/html/exitcodes.html) summary. Everything above 126 seems more like outside of the process.
And I ran into [this](https://unix.stackexchange.com/questions/110348/how-do-i-get-the-list-of-exit-codes-and-or-return-codes-and-meaning-for-a-comm) page.
The page also states:

[...]

PR #1195, opened 2022-01-05
Returns error code 74 on WAL fetch when WAL file does not exist

sebasmannem:
### Database name
Postgres

# Pull request description

### Describe what this PR fix
When running wal-g in restore_command, postgres has 3 ways to handle a exit code:
0: No problem, next WAL file...
1 - 125: End of timeline? Ok, lets stop recovery and go online
>=126: Ouch, big problem. Better not proceed, but error out with a FAIL instead

It would be great if wal-g would be able to distinguish between these cases with a proper exit code too.
Mostly it is so, but config errors, tcp errors, dns errors, etc have the same error code as 'archive does not exist (both 1).

This small change will allow wal-g to return exit code 74 when the wal file does not exist.

* create a postgres backup with wal archive backups
* restore a file that already exists and print the exit code `wal-g wal-fetch 000000010000000300000099 /tmp/000000010000000300000099`. It is 1.
* restore a file with missing config: `WALG_S3_PREFIX=123 wal-g wal-fetch 000000010000000300000099 /tmp/000000010000000300000099`. It is also 1. We should distinguish, but this not different to Postgres either...
* restore a file that does not exist int the backupset `wal-g wal-fetch 00000001000000030000009A /tmp/00000001000000030000009A`and print the exit code. It is also 1, but should be distinguishable

Please see #1126 for more info.

Even with this fix, we still need Postgres to act differently too.
We have requested a change but it is not considered by the PG community yet.
For now we could make this work with a bash script:
```bash
#!/bin/bash
wal-g wal-fetch "$2" "$3"
RET=$?
if [ $RET -eq 1 ]; then
  # whoops, no good. exit code 254 halts postgres with a FAILURE state
  exit 254
fi
exit $RET
```

### DIFF
--- a/internal/databases/postgres/wal_fetch_handler.go
+++ b/internal/databases/postgres/wal_fetch_handler.go
@@ -17,6 +17,9 @@ import (
 	"github.com/wal-g/wal-g/utility"
 )
 
+// Looking at sysexits.h, EX_IOERR (74) is defined as a generic exit code for input/output errors
+const exIoError = 74
+
 type InvalidWalFileMagicError struct {
 	error
 }
@@ -98,7 +101,7 @@ func HandleWALFetch(folder storage.Folder, walFileName string, location string, 
 	err := internal.DownloadFileTo(folder, walFileName, location)
 	if _, isArchNonExistErr := err.(internal.ArchiveNonExistenceError); isArchNonExistErr {
 		tracelog.ErrorLogger.Print(err.Error())
-		os.Exit(74)
+		os.Exit(exIoError)
 	} else {
 		tracelog.ErrorLogger.FatalOnError(err)
 	}
--- a/internal/databases/postgres/wal_fetch_handler.go
+++ b/internal/databases/postgres/wal_fetch_handler.go
@@ -96,7 +96,12 @@ func HandleWALFetch(folder storage.Folder, walFileName string, location string, 
 	}
 
 	err := internal.DownloadFileTo(folder, walFileName, location)
-	tracelog.ErrorLogger.FatalOnError(err)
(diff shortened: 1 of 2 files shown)


internal/databases/postgres/wal_fetch_handler.go
```diff
@@ -96,7 +96,12 @@ func HandleWALFetch(folder storage.Folder, walFileName string, location string,
 	}
 
 	err := internal.DownloadFileTo(folder, walFileName, location)
-	tracelog.ErrorLogger.FatalOnError(err)
+	if _, isArchNonExistErr := err.(internal.ArchiveNonExistenceError); isArchNonExistErr {
+		tracelog.ErrorLogger.Print(err.Error())
+		os.Exit(74)
```
serprex, line 101: Ideally there'd be a link to relevant docs for this magic number

x4m (approved):
I don’t like that we advise to use shell script on top of WAL-G. WAL-G should be that script. Though, let’s merge, I think it’s still for good :)

sebasmannem:
Yes.
One of my hesitations too. 
It is just nasty that we need to set to 126+ to get the behavior we should have from postgres.
If this works as we hope, I will see if I can expand the solution.

haslersn:
>  For PostgreSQL that should be any error code between 126 and 255, which can be achieved with a simple wrapper script.

Where is this documented?

Also, why do we necessarily want to stop PostgreSQL, if the `restore_command` running wal-g exits with an error? I thought PostgreSQL will next try to restore WAL from the `pg_wal` directory, then from the primary server, and then from the archive location again. In a cycle.

PR #1205, opened 2022-01-31
MySQL > Check binlog GTIDs before pushing binlogs to S3

ostinru:
* allow users to reset wal-g's knowledge about uploaded binlog GTID set


ostinru:
Feel free to suggest better naming or better placement for this command

mialinx (commented):
This could be done with  `wal-g st rm` and standard `rm` commands

PR #1206, opened 2022-02-02
Fix dead lock in DownloadAndDecompressSplittedStream

ostinru:
close writers before blocking in error-handling, so MergeWriter will be able to have progress.


mialinx (approved):
LGTM

Issue #562, opened 2020-02-13
wal-g doesn't seem to respect aws.credentials.ChainProvider


kujon:
Hi 👋 

I understand that there was an exactly the same [issue](https://github.com/wal-g/wal-g/issues/5) raised a while ago and a [PR](https://github.com/wal-g/wal-g/pull/12) that addressed it, but it seems it has been broken since moving to a separate [storages package](https://github.com/wal-g/storages/blob/master/s3/session.go#L85-L111).

A simple example to reproduce would be:
```bash
# ~/.aws/credentials
[default]
aws_access_key_id = redacted
aws_secret_access_key = redacted

# ~/.aws/config
[default]
region = eu-west-1
output = json

[profile dev_admin_cli]
role_arn = arn:aws:iam::redacted:role/AdminCLI
region = eu-west-1
source_profile = default
external_id = redacted
```

```bash
> export AWS_PROFILE=dev_admin_cli
> aws sts get-caller-identity
{
    "Account": "redacted",
    "UserId": "redacted:botocore-session-redacted",
    "Arn": "arn:aws:sts::redacted:assumed-role/AdminCLI/botocore-session-redacted"
}
```

```bash
> wal-g backup-list
DEBUG: 2020/02/13 17:17:44.240330 --- COMPILED ENVIRONMENT VARS ---
DEBUG: 2020/02/13 17:17:44.241627 HOSTNAME=redacted
DEBUG: 2020/02/13 17:17:44.242767 PGPORT=5432
DEBUG: 2020/02/13 17:17:44.243837 AWS_PROFILE=dev_admin_cli
DEBUG: 2020/02/13 17:17:44.244858 PGPASSWORD=redacted
DEBUG: 2020/02/13 17:17:44.245814 AWS_REGION=eu-west-1
DEBUG: 2020/02/13 17:17:44.246788 PWD=/
DEBUG: 2020/02/13 17:17:44.247164 WALG_S3_PREFIX=s3://redacted
DEBUG: 2020/02/13 17:17:44.247274 HOME=/root
DEBUG: 2020/02/13 17:17:44.247841 PGUSER=redacted
DEBUG: 2020/02/13 17:17:44.248862 TERM=xterm
DEBUG: 2020/02/13 17:17:44.249351 WALG_LOG_LEVEL=DEVEL
DEBUG: 2020/02/13 17:17:44.249779 SHLVL=1
DEBUG: 2020/02/13 17:17:44.250353 PGHOST=redacted
[...]

PR #1209, opened 2022-02-04
fix: AWS auth fixes

### LINKED ISSUE #886: AWS authentication using OIDC web identity token
WAL-G doesn't support native AWS authentication using OIDC web identity token and role arn environmental variables, `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN`.

This is mainly useful on k8s, if you're using [IAM roles linked with Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)

This feature is natively supported by the GO AWS-SDK https://github.com/aws/aws-sdk-go/pull/2667 and [referenced](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html) even in the documentation

---

The current workaround is to assume web identity using `aws sts assume-role-with-web-identity  --role-arn $AWS_ROLE_ARN  --role-session-name wal-g  --web-identity-token file://$AWS_WEB_IDENTITY_TOKEN_FILE` and parse the JSON output to export variables necessary for WAL-G. Not really ideal if the SDK supports it out of the box.
AWSCLI (v2) is pretty expensive for container image as a 110MB dependency...

Happy to help with this issue, just need a little bit of direction on where should I look. I'm assuming it's https://github.com/wal-g/storages/blob/master/s3/session.go right? From a quick look, it looks like the problem is with the default Config handler. That might also explain the freezing problem I'm encountering (https://github.com/wal-g/wal-g/issues/246#issuecomment-621463846).

Fixing this would cover https://github.com/wal-g/wal-g/issues/246 as well as many more modes that the SDK supports out of the box.

kujon:
### Database name
Any

# Pull request description
Fixes #886, #562 

### Describe what this PR fix
wal-g supports a very limited number of methods of authenticating with AWS. This PR does 2 things:
- ~it deprecates explicitly passing in aws credentials~
- it makes use of aws.credentials.ChainProvider to quack the same as e.g. aws cli.

It's a remake of [this PR from a now deprecated repo](https://github.com/wal-g/storages/pull/15).

### DIFF
--- a/pkg/storages/s3/folder.go
+++ b/pkg/storages/s3/folder.go
@@ -22,11 +22,11 @@ const (
 	EndpointSetting          = "AWS_ENDPOINT"
 	RegionSetting            = "AWS_REGION"
 	ForcePathStyleSetting    = "AWS_S3_FORCE_PATH_STYLE"
-	AccessKeyIdSetting       = "AWS_ACCESS_KEY_ID"     // deprecated
-	AccessKeySetting         = "AWS_ACCESS_KEY"        // deprecated
-	SecretAccessKeySetting   = "AWS_SECRET_ACCESS_KEY" // deprecated
-	SecretKeySetting         = "AWS_SECRET_KEY"        // deprecated
-	SessionTokenSetting      = "AWS_SESSION_TOKEN"     // deprecated
+	AccessKeyIdSetting       = "AWS_ACCESS_KEY_ID"
+	AccessKeySetting         = "AWS_ACCESS_KEY"
+	SecretAccessKeySetting   = "AWS_SECRET_ACCESS_KEY"
+	SecretKeySetting         = "AWS_SECRET_KEY"
+	SessionTokenSetting      = "AWS_SESSION_TOKEN"
 	SseSetting               = "S3_SSE"
 	SseCSetting              = "S3_SSE_C"
 	SseKmsIdSetting          = "S3_SSE_KMS_ID"
--- a/pkg/storages/s3/folder.go
+++ b/pkg/storages/s3/folder.go
@@ -22,11 +22,11 @@ const (
 	EndpointSetting          = "AWS_ENDPOINT"
 	RegionSetting            = "AWS_REGION"
 	ForcePathStyleSetting    = "AWS_S3_FORCE_PATH_STYLE"
-	AccessKeyIdSetting       = "AWS_ACCESS_KEY_ID"
-	AccessKeySetting         = "AWS_ACCESS_KEY"
-	SecretAccessKeySetting   = "AWS_SECRET_ACCESS_KEY"
-	SecretKeySetting         = "AWS_SECRET_KEY"
-	SessionTokenSetting      = "AWS_SESSION_TOKEN"
+	AccessKeyIdSetting       = "AWS_ACCESS_KEY_ID"     // deprecated
+	AccessKeySetting         = "AWS_ACCESS_KEY"        // deprecated
+	SecretAccessKeySetting   = "AWS_SECRET_ACCESS_KEY" // deprecated
+	SecretKeySetting         = "AWS_SECRET_KEY"        // deprecated
+	SessionTokenSetting      = "AWS_SESSION_TOKEN"     // deprecated
 	SseSetting               = "S3_SSE"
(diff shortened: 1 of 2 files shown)


usernamedt:
Hi! As I can see, this issue has been already addressed in https://github.com/wal-g/storages/pull/45 and https://github.com/wal-g/wal-g/pull/1183. Can you please clarify the difference between this PR and the master branch (except the warning messages)?

mialinx:
There are two moments:
* S3 is more than AWS object, storage there are other S3-compatible (Yandex Cloud for instance), which provides S3 REST API, but not AWS-like configuration.  So, `AWS_ACCESS_KEY_ID` and others can't be deprecated at the moment.
* There is a scenario with copying backups from one storage to another. In this case, we have to configure multiple storage with different bucket_names (and may be access keys). How can we handle it without `AWS_*` options in config ?

mialinx:
It looks like the problem is that we use storage settings (such as `AWS_ACCESS_KEY_ID`) to choose configured storage. 
Maybe we need to add explicit storage type setting like `WALG_STORAGE=S3` or `WALG_STORAGE=SSH`, leaving current method as fallback. We will be able to use S3 storage without any (`AWS_...`) options in config at all, just as @kujon suggests.

@usernamedt ?

usernamedt:
> It looks like the problem is that we use storage settings (such as `AWS_ACCESS_KEY_ID`) to choose configured storage. Maybe we need to add explicit storage type setting like `WALG_STORAGE=S3` or `WALG_STORAGE=SSH`, leaving current method as fallback. We will be able to use S3 storage without any (`AWS_...`) options in config at all, just as @kujon suggests.
> 
> @usernamedt ?

But we already don't rely on any `AWS_*` auth setting to determine that storage type is AWS. WAL-G  uses the `WALG_*_PREFIX` for this purpose.

mialinx:
> > It looks like the problem is that we use storage settings (such as `AWS_ACCESS_KEY_ID`) to choose configured storage. Maybe we need to add explicit storage type setting like `WALG_STORAGE=S3` or `WALG_STORAGE=SSH`, leaving current method as fallback. We will be able to use S3 storage without any (`AWS_...`) options in config at all, just as @kujon suggests.
> > @usernamedt ?
> 
> We already don't rely on any `AWS_*` auth setting to determine that storage type is AWS. WAL-G uses the `WALG_*_PREFIX` for this purpose, i.e. `WALG_S3_PREFIX` in the case of AWS.

Oh, my bad )
Well, then we could just remove deprecation comments warnings (as we plan to `AWS_ACCESS_KEY_ID` for a long time) and merge this PR.

kujon:
> > > It looks like the problem is that we use storage settings (such as `AWS_ACCESS_KEY_ID`) to choose configured storage. Maybe we need to add explicit storage type setting like `WALG_STORAGE=S3` or `WALG_STORAGE=SSH`, leaving current method as fallback. We will be able to use S3 storage without any (`AWS_...`) options in config at all, just as @kujon suggests.
> > > @usernamedt ?
> > 
> > We already don't rely on any `AWS_*` auth setting to determine that storage type is AWS. WAL-G uses the `WALG_*_PREFIX` for this purpose, i.e. `WALG_S3_PREFIX` in the case of AWS.
> 
> Oh, my bad )
> Well, then we could just remove deprecation comments warnings (as we plan to `AWS_ACCESS_KEY_ID` for a long time) and merge this PR.

Will do, give me a day!

martin31821:
Any chance we can get a release containing this PR?

x4m:
@usernamedt do you plan release\prerelease soon? Do we have something to show to the community? :)

Jasper-Ben:
:wave: @x4m, I am working with @martin31821. The good news is: v2.0.0-rc1 works for our usecase. See: https://github.com/zalando/postgres-operator/issues/1124#issuecomment-1110788917
@usernamedt a GA release would be great! :+1:

usernamedt:
> @usernamedt do you plan release\prerelease soon? Do we have something to show to the community? :)

Hi! Yes, I guess that we can cut a release soon. Most probably, the next week.

PR #1210, opened 2022-02-08
wal-restore remote target feature

Xaspy:
PostgreSQL. Added remote cluster support to wal-restore.

### DIFF
--- a/cmd/pg/wal_restore.go
+++ b/cmd/pg/wal_restore.go
@@ -9,23 +9,64 @@ import (
 
 const (
 	WalRestoreUsage            = "wal-restore target-pgdata source-pgdata"
-	WalRestoreShortDescription = "Restores WAL segments from storage."
-	WalRestoreLongDescription  = "Restores the missing WAL segments that will be needed to perform pg_rewind from storage."
+	WalRestoreShortDescription = "Restores WAL segments from storage to the source to perform pg_rewind with the target."
+	WalRestoreLongDescription  = "Restores the missing WAL segments that will be needed in source cluster to perform pg_rewind " +
+		"with target cluster from storage. If you mark the target cluster as remote you should specify requisites to connect by" +
+		" flags."
+	RemoteFlag            = "remote"
+	sshHostFlag           = "host-ssh"
+	sshPortFlag           = "port-ssh"
+	sshUsernameFlag       = "username-ssh"
+	sshPasswordFlag       = "password-ssh"
+	sshPrivateKeyPathFlag = "private-key-path-ssh"
 )
 
 // walRestoreCmd represents the walRestore command
-var walRestoreCmd = &cobra.Command{
-	Use:   WalRestoreUsage,
-	Short: WalRestoreShortDescription,
-	Long:  WalRestoreLongDescription,
-	Args:  cobra.ExactArgs(2),
-	Run: func(cmd *cobra.Command, args []string) {
-		folder, err := internal.ConfigureFolder()
-		tracelog.ErrorLogger.FatalfOnError("Error on configure external folder %v\n", err)
-		postgres.HandleWALRestore(args[0], args[1], folder)
-	},
-}
+var (
+	walRestoreCmd = &cobra.Command{
+		Use:   WalRestoreUsage,
+		Short: WalRestoreShortDescription,
+		Long:  WalRestoreLongDescription,
+		Args:  cobra.ExactArgs(2),
+		Run: func(cmd *cobra.Command, args []string) {
+			folder, err := internal.ConfigureFolder()
+			tracelog.ErrorLogger.FatalfOnError("Error on configure external folder %s\n", err)
+
+			sshRequisites := postgres.SshRequisites{}
+			if sshHost != "" {
+				sshRequisites.Host = sshHost
+			}
+			if sshPort != "" {
+				sshRequisites.Port = sshPort
+			}
+			if sshUsername != "" {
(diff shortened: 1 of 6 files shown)


cmd/pg/wal_restore.go
```diff
@@ -9,23 +9,64 @@ import (
 
 const (
 	WalRestoreUsage            = "wal-restore target-pgdata source-pgdata"
-	WalRestoreShortDescription = "Restores WAL segments from storage."
-	WalRestoreLongDescription  = "Restores the missing WAL segments that will be needed to perform pg_rewind from storage."
+	WalRestoreShortDescription = "Restores WAL segments from storage to the source to perform pg_rewind with the target."
+	WalRestoreLongDescription  = "Restores the missing WAL segments that will be needed in source cluster to perform pg_rewind " +
+		"with target cluster from storage. If you mark the target cluster as remote you should specify requisites to connect by" +
```
usernamedt, line 14: Add the note that in the case of remote mode, `wal-restore`  should be run from the `source` cluster.

internal/databases/postgres/wal_restore_handler.go
```diff
@@ -46,30 +53,30 @@ func HandleWALRestore(targetPath, sourcePath string, cloudFolder storage.Folder)
...
-	tracelog.ErrorLogger.FatalfOnError("Failed to get WAL directory name: %v\n", err)
+	tracelog.ErrorLogger.FatalfOnError("Failed to get WAL directory name: %s\n", err)
 	sourceWalDir, err := getWalDirName(sourcePath)
-	tracelog.ErrorLogger.FatalfOnError("Failed to get WAL directory name: %v\n", err)
+	tracelog.ErrorLogger.FatalfOnError("Failed to get WAL directory name: %s\n", err)
 
 	tgtHistoryRecords, err := getLocalTimelineHistoryRecords(targetPgData.GetCurrentTimeline(), targetWalDir)
-	tracelog.ErrorLogger.FatalfOnError("Failed to get history data on target cluster: %v\n", err)
+	tracelog.ErrorLogger.FatalfOnError("Failed to get history data on target cluster: %s\n", err)
 	srcHistoryRecords, err := getLocalTimelineHistoryRecords(sourcePgData.GetCurrentTimeline(), sourceWalDir)
```
usernamedt, line 62: I am curious how does it work in the case of remote `targetWalDir`?

pkg/storages/sh/folder.go
```diff
@@ -3,15 +3,13 @@ package sh
...
-	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 
-	"github.com/pkg/sftp"
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/pkg/storages/storage"
-	"golang.org/x/crypto/ssh"
+	"github.com/wal-g/wal-g/utility"
```
usernamedt, line 12: `wal-g/wal-g/pkg/storages` is a standalone package and it should not rely on any other `wal-g/wal-g/*` package.

usernamedt (changes requested):

internal/databases/postgres/wal_restore_handler.go
Xaspy: My bad. I was too hasty. Will be fixed.

cmd/pg/wal_restore.go
Xaspy: Added.

internal/databases/postgres/pg_control_data_test.go
```diff
@@ -28,13 +28,13 @@ func TestExtractPgControlData_OldVersion(t *testing.T) {
...
 
-	pgControlData, err := extractPgControlData(bytes2.NewReader(bytes))
+	pgControlData, err := parsePgControlData(bytes2.NewReader(bytes))
 	assert.Nil(t, err)
 	assert.Equal(t, uint64(9876), pgControlData.GetSystemIdentifier())
 	assert.Equal(t, uint32(7), pgControlData.GetCurrentTimeline())
 }
 
-func TestExtractPgControlData_NewVersion(t *testing.T) {
+func TestParsePgControlData_NewVersion(t *testing.T) {
```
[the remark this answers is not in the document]
Xaspy: Added explanation.

pkg/storages/sh/folder.go
Xaspy: Now SFTP client locates in `wal-g/pkg/sftp`.

pkg/storages/sh/folder.go
```diff
@@ -45,49 +43,17 @@ func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder,
 		return nil, err
 	}
 
-	user := settings[Username]
```
usernamedt, line 48: Ok, why we can't instead configure it by creating some config struct (`viper.New()`) with SSH options and init a folder? 

It will allow to use the existing storages implementation without any modification.

You can use this code as a reference
https://github.com/wal-g/wal-g/blob/3b3ec011f2c17bc46397046a83e6cb2671261d87/internal/config.go#L702-L720

usernamedt (changes requested):

usernamedt:
Looks cool, I guess we also need an integration test for remote target scenario :)

pkg/storages/sh/folder.go
Xaspy: I don't quite understand. This code, if dive into, works with the code presented above.

PR #1223, opened 2022-02-22
fix mysql backup of a lagging replica

mialinx:
# fix MySQL backup of a lagging replica

### Describe what this PR fix
Suppose we have a standard primary => secodary replication schema.
Primary works fine, and wal-g have uploaded `primary.0001` - `primary.0099` binlogs to storage.
Secondary instance is lagging and 5 hours behind the master, it's currently applying transactions from `primary.0050` binlog.
When we're making a full backup of a secondary, we should mark that it corresponds to `primary.0050` (currently applying) not `primary.0099` (last uploaded) binlog.
Otherwise, PITR will not work due to gap in transactions.
  
</p>
</details>

### DIFF
--- a/internal/databases/mysql/binlog_find_handler.go
+++ b/internal/databases/mysql/binlog_find_handler.go
@@ -1,0 +1,23 @@
+package mysql
+
+import (
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+	"github.com/wal-g/wal-g/utility"
+
+	"github.com/wal-g/tracelog"
+)
+
+func HandleBinlogFind(folder storage.Folder, gtid string) {
+	db, err := getMySQLConnection()
+	tracelog.ErrorLogger.FatalOnError(err)
+	defer utility.LoggedClose(db, "")
+	flavor, err := getMySQLFlavor(db)
+	tracelog.ErrorLogger.FatalOnError(err)
+	if gtid == "" {
+		gtid, err = getMySQLGTIDExecuted(db, flavor)
+		tracelog.ErrorLogger.FatalOnError(err)
+	}
+	name, err := getLastUploadedBinlogBeforeGTID(folder, gtid, flavor)
+	tracelog.ErrorLogger.FatalOnError(err)
+	tracelog.InfoLogger.Println(name)
+}
--- a/internal/databases/mysql/mysql.go
+++ b/internal/databases/mysql/mysql.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/wal-g/wal-g/internal/compression"
 
-	flavors "github.com/go-mysql-org/go-mysql/mysql"
+	gomysql "github.com/go-mysql-org/go-mysql/mysql"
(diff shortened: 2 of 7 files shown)


internal/databases/mysql/binlog_find_handler.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func HandleBinlogFind(folder storage.Folder, gtid string) {
+	db, err := getMySQLConnection()
+	tracelog.ErrorLogger.FatalOnError(err)
+	defer utility.LoggedClose(db, "")
+	flavor, err := getMySQLFlavor(db)
+	tracelog.ErrorLogger.FatalOnError(err)
+	if gtid == "" {
+		gtid, err = getMySQLGTIDExecuted(db, flavor)
```
ostinru, line 17: Is there at least single valid usage for this fallback?
I think that users may use it in a way that will break their backup-recovery process.

internal/databases/mysql/mysql.go
```diff
@@ -61,6 +56,21 @@ func getMySQLCurrentBinlogFileLocal(db *sql.DB) (fileName string) {
...
 
+func getMySQLGTIDExecuted(db *sql.DB, flavor string) (gtid string, err error) {
+	query := ""
+	switch flavor {
+	case gomysql.MySQLFlavor:
+		query = "SELECT @@global.gtid_executed"
+	case gomysql.MariaDBFlavor:
+		query = "SELECT @@global.gtid_current_pos"
+	default:
+		return "", fmt.Errorf("nuknown MySQL flavor: %s", flavor)
```
ostinru, line 67: typo: `nuknown` -> `unknown`

internal/databases/mysql/mysql.go
```diff
@@ -80,6 +90,34 @@ func getLastUploadedBinlog(folder storage.Folder) (string, error) {
...
+	for i := len(logFiles) - 1; i > 0; i-- {
+		prevGtid, err := GetBinlogPreviousGTIDsRemote(folder, logFiles[i].GetName(), flavor)
+		if err != nil {
+			return "", err
+		}
+		if gtidParsed.Contain(prevGtid) {
+			return utility.TrimFileExtension(logFiles[i].GetName()), nil
+		}
+	}
+	return "", nil
```
ostinru, line 118: Fallback to first binlog?

internal/databases/mysql/mysql.go
mialinx: fixed

internal/databases/mysql/mysql.go
mialinx: > Fallback to first binlog?

First one it not better than the last one. 
In the sense, that there are still some transactions missed in base backup.
But we will need to apply all uploaded binlogs during PITR and will still fail slowly. 
Let's fail fast and keep empty `BinLogStart` to simplify debugging.

BTW, added warning for this case.

ostinru (approved):
Looks good to me

PR #1229, opened 2022-03-03
Make mysql bilong-push status observalble to 3rd party tools

ostinru:
Make mysql bilong-push status observalble to 3rd party tools
 * write cache file when no files uploaded (so, monitoring scripts will be able to check it)
 * write cache file from time to time, just to make upload progress visible

### DIFF
--- a/internal/databases/mysql/binlog_push_handler.go
+++ b/internal/databases/mysql/binlog_push_handler.go
@@ -38,12 +38,13 @@ func HandleBinlogPush(uploader internal.UploaderProvider, untilBinlog string, ch
 	binlogsFolder, err := getMySQLBinlogsFolder(db)
 	tracelog.ErrorLogger.FatalOnError(err)
 
-	binlogs, err := getMySQLSortedBinlogs(db)
-	tracelog.ErrorLogger.FatalOnError(err)
-
 	if untilBinlog == "" || untilBinlog > getMySQLCurrentBinlogFileLocal(db) {
+		// select safe to upload binlog before acquiring list of binlogs [avoid race]
 		untilBinlog = getMySQLCurrentBinlogFileLocal(db)
 	}
+
+	binlogs, err := getMySQLSortedBinlogs(db)
+	tracelog.ErrorLogger.FatalOnError(err)
 
 	var binlogSentinelDto BinlogSentinelDto
 	err = FetchBinlogSentinel(rootFolder, &binlogSentinelDto)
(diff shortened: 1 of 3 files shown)


internal/databases/mysql/binlog_push_handler.go
```diff
@@ -68,32 +69,45 @@ func HandleBinlogPush(uploader internal.UploaderProvider, untilBinlog string, ch
...
+
 		// Upload binlogs:
 		err = archiveBinLog(uploader, binlogsFolder, binLog)
 		tracelog.ErrorLogger.FatalOnError(err)
 
 		for _, filter := range filters {
 			filter.onUpload(&cache)
 		}
-		putCache(cache)
+		if i%10 == 0 {
```
mialinx, line 103: it's a cheap operation, we can write cache file on every binlog

internal/databases/mysql/binlog_push_handler.go
```diff
@@ -68,32 +69,45 @@ func HandleBinlogPush(uploader internal.UploaderProvider, untilBinlog string, ch
...
 				tracelog.DebugLogger.Printf("Skip binlog %v (%s check)\n", binLog, filter.name())
-				continue outer
+				upload = false
+				break
 			}
 		}
 
+		if !upload {
+			for _, filter := range filters {
+				filter.onSkip(&cache)
```
mialinx, line 91: We should not trigger onSkip and modify `LastArchivedBinlog` if binlog was skipped by  `archivedBinlogFilter`

Suppose we have `bl.001` - `bl.099`, and all binlogs up to `bl.050` were already uploaded and `LastArchivedBinlog` is set to `bl.050` in cache.
During next run we will check `bl.001` to be uploaded, `archivedBinlogFilter` will discard it, and we will update `data.LastArchivedBinlog`  back to `bl.001`.

mialinx (changes requested):

internal/databases/mysql/binlog_push_handler.go
```diff
@@ -53,47 +53,67 @@ func HandleBinlogPush(uploader internal.UploaderProvider, untilBinlog string, ch
...
+	var filter gtidFilter
 	if checkGTIDs {
 		flavor, err := getMySQLFlavor(db)
 		if flavor == "" || err != nil {
 			flavor = mysql.MySQLFlavor
 		}
 		if flavor == mysql.MySQLFlavor {
-			filters = append(filters, &gtidFilter{BinlogsFolder: binlogsFolder, Flavor: flavor})
+			gtid, _ := mysql.ParseMysqlGTIDSet(cache.GTIDArchived)
+			gtidArchived, _ := gtid.(*mysql.MysqlGTIDSet)
```
mialinx, line 64: It's not safe in conjunction with flavor defaults on L60 ^^.
Let's return error if we are not sure with flavor.

internal/databases/mysql/binlog_push_handler.go
```diff
@@ -53,47 +53,67 @@ func HandleBinlogPush(uploader internal.UploaderProvider, untilBinlog string, ch
...
+				gtidArchived:  gtidArchived,
+				lastGtidSeen:  nil,
+			}
 		}
 	}
-	for _, filter := range filters {
-		filter.init(cache)
-	}
 
 outer:
```
mialinx, line 74: we don't need this label anymore, can just `continue`

mialinx (changes requested):

mialinx (approved):
LGTM

PR #1230, opened 2022-03-05
Optimize fsync

HolyPrapor:
# Pull request description

Another solution of #1046


internal/databases/postgres/tar_interpreter.go
```diff
@@ -24,14 +24,55 @@ type FileTarInterpreter struct {
 	UnwrapResult    *UnwrapResult
 
 	createNewIncrementalFiles bool
+	explicitFsync             *bool
 }
 
+// func NewFileTarInterpreter(
```
HolyPrapor, line 30: Previous implementation

internal/databases/postgres/tar_interpreter.go
```diff
@@ -24,14 +24,55 @@ type FileTarInterpreter struct {
...
+// 	dbDataDirectory string, sentinel BackupSentinelDto, filesMetadata FilesMetadataDto,
+// 	filesToUnwrap map[string]bool, createNewIncrementalFiles bool,
+// ) internal.TarInterpreter {
+// 	return &FileTarInterpreter{dbDataDirectory, sentinel, filesMetadata,
+// 		filesToUnwrap, newUnwrapResult(), createNewIncrementalFiles, nil}
+// }
+
+// func NewFileTarInterpreter(
```
HolyPrapor, line 38: Implementation that uses errgroup to wait for all file syncs

internal/databases/postgres/tar_interpreter.go
```diff
@@ -24,14 +24,55 @@ type FileTarInterpreter struct {
...
+// 	dbDataDirectory string, sentinel BackupSentinelDto, filesMetadata FilesMetadataDto,
+// 	filesToUnwrap map[string]bool, createNewIncrementalFiles bool,
+// ) *GroupFileTarInterpreter {
+// 	return NewGroupFileTarInterpreterWithGlobalFsync(dbDataDirectory, sentinel, filesMetadata, filesToUnwrap, createNewIncrementalFiles)
+// }
+
 func NewFileTarInterpreter(
```
HolyPrapor, line 45: Implementation that uses global file sync to flush all caches at the end

internal/extract.go
```diff
@@ -78,6 +83,10 @@ var _ io.Writer = &DevNullWriter{}
 func extractOneTar(tarInterpreter TarInterpreter, source io.Reader) error {
 	tarReader := tar.NewReader(source)
 
+	if groupInterpreter, ok := tarInterpreter.(GroupTarInterpreter); ok {
```
HolyPrapor, line 86: Little abstraction leak here, but I think that's more appropriate than duplication of all methods for GroupTarInterpreter

HolyPrapor:
Benchmarking required in order to determine what is the best solution.

Will finish this PR after performance tests 😃

serprex:
Does this only apply to NewFile functions?

The old functions are still used quite a bit, NewFile doesn't work with non-reverse delta, & old file is still used for non-delta backups

HolyPrapor:
The `NewFile` are just constructors. The 'Old/New' that you're talking about (as far as I understand you) are hidden inside the implementation.
That means this solution should apply to old functions as well.

serprex:
Is there a reason to keep the fsync-after-every-file logic?

HolyPrapor:
> Is there a reason to keep the fsync-after-every-file logic?

Fair enough. 
I can do this after someone reviews global-fsync part.

usernamedt (changes requested):

internal/config.go
```diff
@@ -53,6 +53,7 @@ const (
 	LogLevelSetting              = "WALG_LOG_LEVEL"
 	TarSizeThresholdSetting      = "WALG_TAR_SIZE_THRESHOLD"
 	TarDisableFsyncSetting       = "WALG_TAR_DISABLE_FSYNC"
+	TarDisableGlobalFsyncSetting = "WALG_TAR_DISABLE_GLOBAL_FSYNC"
```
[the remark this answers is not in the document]
HolyPrapor: What's the point of having `WALG_TAR_DISABLE_GLOBAL_FSYNC` then?
Let's not introduce it, `WALG_TAR_FSYNC_MODE` is enough.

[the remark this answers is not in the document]
serprex: I think he meant `WALG_TAR_DISABLE_FSYNC`, & that users with that option enabled should be on `DISABLED` until they switch to `WALG_TAR_FSYNC_MODE=GLOBAL` in their config

serprex (approved):

PR #1236, opened 2022-03-19
Support better profiling by not stripping binaries

matt-smiley:
### Database name

This is purely a `go build` config change, and is equally applicable to all variants of the `wal-g` build.

# Pull request description

Here we propose to stop stripping symbols and debug sections from the `wal-g` binary.

Stripping the binary saves a little disk space (about 14 MB) but severely inhibits profiling support (see demo below).  The trade-off is generally not worth it.

### Describe what this PR fix

When building the `wal-g` binary, prior to this pull request we tell the go linker to strip the symbol table and debug sections from the binary.  Doing this badly inhibits observability. It breaks profiling and dynamic instrumentation, making it more difficult to analyze and improve performance bottlenecks.

Keeping the binary unstripped fixes this.

### Example

The following pair of flamegraphs are profiles of a toy workload.  The take-home message is that the profile of the stripped binary tells us very little, whereas profiling the unstripped binary tells us a lot.

#### Stripped (current behavior, produces nearly useless profile results)

Every `wal-g` frame in this flamegraph lacks a function name.  Consequently this profile tells us nothing about which code paths are hot for this particular workload.

walg_stripped_binary backup_list flamegraph

#### Unstripped (proposed behavior, produces useful profile results)

In contrast, this flamegraph has function names for all of the `wal-g` frames, so we can see which code paths are prominent for this workload.  Being able to observe the hot code paths supports bottleneck analysis and can help inform optimization decisions, both from a configuration standpoint and a design enhancement standpoint.

walg_unstripped_binary backup_list flamegraph

### What if someone really wants a stripped binary?

Stripping the binary saves roughly 14 MB.  In general that seems like a small cost in exchange for greatly improved observability. If anyone using wal-g would prefer to have that space savings rather than the ability to profile, they can reapply the stripping either:
* during the build, by setting the make variable STRIP_BINARIES
* after the build, by using "strip" or "objcopy --strip-debug"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,10 @@ ifdef USE_LZO
 	BUILD_TAGS:=$(BUILD_TAGS) lzo
 endif
 
+ifdef STRIP_BINARIES
+	EXTRA_LDFLAGS:=$(EXTRA_LDFLAGS) -s -w
+endif
+
 .PHONY: unittest fmt lint clean install_tools
 
 test: deps unittest pg_build mysql_build redis_build mongo_build gp_build unlink_brotli pg_integration_test mysql_integration_test redis_integration_test fdb_integration_test gp_integration_test
@@ -36,7 +40,7 @@ test: deps unittest pg_build mysql_build redis_build mongo_build gp_build unlink
 pg_test: deps pg_build unlink_brotli pg_integration_test
 
 pg_build: $(CMD_FILES) $(PKG_FILES)
-	(cd $(MAIN_PG_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/pg.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/pg.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/pg.walgVersion=`git tag -l --points-at HEAD`")
+	(cd $(MAIN_PG_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "$(EXTRA_LDFLAGS) -X github.com/wal-g/wal-g/cmd/pg.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/pg.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/pg.walgVersion=`git tag -l --points-at HEAD`")
 
 install_and_build_pg: deps pg_build
 
@@ -82,10 +86,10 @@ mysql_base: deps mysql_build unlink_brotli
 mysql_test: deps mysql_build unlink_brotli mysql_integration_test
 
 mysql_build: $(CMD_FILES) $(PKG_FILES)
-	(cd $(MAIN_MYSQL_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/mysql.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/mysql.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/mysql.walgVersion=`git tag -l --points-at HEAD`")
+	(cd $(MAIN_MYSQL_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "$(EXTRA_LDFLAGS) -X github.com/wal-g/wal-g/cmd/mysql.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/mysql.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/mysql.walgVersion=`git tag -l --points-at HEAD`")
 
 sqlserver_build: $(CMD_FILES) $(PKG_FILES)
-	(cd $(MAIN_SQLSERVER_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/sqlserver.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/sqlserver.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/sqlserver.walgVersion=`git tag -l --points-at HEAD`")
+	(cd $(MAIN_SQLSERVER_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "$(EXTRA_LDFLAGS) -X github.com/wal-g/wal-g/cmd/sqlserver.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/sqlserver.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/sqlserver.walgVersion=`git tag -l --points-at HEAD`")
 
 load_docker_common:
 	@if [ "x" = "${CACHE_FILE_UBUNTU}x" ]; then\
@@ -117,7 +121,7 @@ mariadb_integration_test: load_docker_common
 mongo_test: deps mongo_build unlink_brotli
 
 mongo_build: $(CMD_FILES) $(PKG_FILES)
-	(cd $(MAIN_MONGO_PATH) && go build $(BUILD_ARGS) -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/mongo.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/mongo.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/mongo.walgVersion=`git tag -l --points-at HEAD`")
+	(cd $(MAIN_MONGO_PATH) && go build $(BUILD_ARGS) -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "$(EXTRA_LDFLAGS) -X github.com/wal-g/wal-g/cmd/mongo.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/mongo.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/mongo.walgVersion=`git tag -l --points-at HEAD`")
 
 mongo_install: mongo_build
 	mv $(MAIN_MONGO_PATH)/wal-g $(GOBIN)/wal-g
@@ -132,7 +136,7 @@ clean_mongo_features:
 	cd tests_func/ && MONGO_MAJOR=$(MONGO_MAJOR) MONGO_VERSION=$(MONGO_VERSION) go test -v -count=1  -timeout 5m -tf.test=false -tf.debug=false -tf.clean=true -tf.stop=true -tf.database=mongodb
 
 fdb_build: $(CMD_FILES) $(PKG_FILES)
-	(cd $(MAIN_FDB_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "-s -w")
+	(cd $(MAIN_FDB_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "$(EXTRA_LDFLAGS)")
 
 fdb_install: fdb_build
 	mv $(MAIN_FDB_PATH)/wal-g $(GOBIN)/wal-g
@@ -145,7 +149,7 @@ fdb_integration_test: load_docker_common
 redis_test: deps redis_build unlink_brotli redis_integration_test
 
 redis_build: $(CMD_FILES) $(PKG_FILES)
-	(cd $(MAIN_REDIS_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/redis.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/redis.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/redis.walgVersion=`git tag -l --points-at HEAD`")
+	(cd $(MAIN_REDIS_PATH) && go build -mod vendor -tags "$(BUILD_TAGS)" -o wal-g -ldflags "$(EXTRA_LDFLAGS) -X github.com/wal-g/wal-g/cmd/redis.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/redis.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/redis.walgVersion=`git tag -l --points-at HEAD`")
 
 redis_integration_test: load_docker_common
 	docker-compose build redis redis_tests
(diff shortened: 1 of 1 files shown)


serprex (approved):

serprex:
I'm +1 on this change, but I'll let someone else merge in case there's a discussion to be had

x4m:
We stripped binaries per @CyberDem0n request. Maybe we could also ship unstripped binary along with stripped?
Ideally we should provide Lego-like executable constructor. User can request 1 compression method, 1 storage, stripped\nonstripped binary. All this can significantly reduce size.
Hopefully daemon mode could alleviate the problem of executable size...

matt-smiley:
Thank you both!  This change to support better profiling is going to be very helpful for folks trying to discover their bottlenecks, optimize their configuration, or propose and evaluate new ideas.

For context, in case it helps, this observability gap came up while working on a saturation problem for `wal-g wal-push`.  The effective throughput can be influenced by interactions between several factors, including:
* Design decisions: The concurrency model for the background uploader goroutines implicitly allows a new background uploader to start right before the main file upload finishes.  This extends the lifespan of the process and may erode throughput by reducing mean concurrency.
* Configuration decisions: The number of concurrent background uploaders and the limit on the max number of files to upload in the background both influence the effective concurrency level
* Environmental factors: Many runtime conditions can potentially affect throughput and even shift where the constraining bottleneck is.  Examples include: the network throughput to the destination storage service, the number of available OS threads relative to the number of goroutines, the CPU consumption by other competing processes, memory management behavior such as foreground hugepage defrag, etc.

Having profiling support makes it much more feasible to discover such interactions and improve efficiency, either through config changes, machine provisioning adjustments, workload changes, or new design ideas.  I'm really looking forward to getting this in place, both for my team's near-term goals and for the broader community's benefit.

CyberDem0n:
Just look at these numbers:
```bash
root@6a8677bfcf14:/home/postgres# time wal-g > /dev/null

real    0m0.066s
user    0m0.059s
sys     0m0.040s
root@6a8677bfcf14:/home/postgres# time wal-g > /dev/null

real    0m0.080s
user    0m0.064s
sys     0m0.047s
root@6a8677bfcf14:/home/postgres# time wal-g > /dev/null

real    0m0.051s
user    0m0.040s
sys     0m0.035s
root@6a8677bfcf14:/home/postgres# time wal-g > /dev/null

real    0m0.057s
user    0m0.055s
sys     0m0.017s
```

It takes at least 50ms just to start wal-g without doing anything useful. I call this number startup time. Simple math tells that it can't push more ~20 files/second under such conditions. Also, I observed that things get worse if the system is under stress (CPU//IO).

The startup time nearly lineraly depends on the size of the binary. It is very simple to check it by taking for instance `v0.1.0`, which binary is 12MB. Its startup time is ~15ms-17ms.

If we go back to the not stripped binary (by default) we immediately worsen the performance.

x4m:
Let's do daemon mode for wal-push?
I'd propose even inheriting protocol from pgBackRest.

NikolayS:
@x4m 

Do we really need some protocol for Postgres at all? Couldn't it be a walg daemon that listens via a UNIX socket and the `wal-push` command that just tells it what file to archive?

fdr:
Are we sure binary size is the dominant factor in startup time? I at at least equally suspicious about library initialization. mmap and faulting a few pages for instructions for the nominal entry point of wal-g I doubt is so much more expensive.

If it is in (hot, cached) binary size, there can be external debug symbol files.

NikolayS:
(thinking out of loud) I guess we should profile the startup and see where the time is spent, compared with the stripped version.

matt-smiley:
A lot of factors can affect a process's startup time.  For a small binary like `wal-g`, I don't think stripping is one of them (see test results below).

Profiling can tell us empirically where the time is spent.  In the hope that we can take some of the uncertainty and guesswork out of the discussion, here are some observations and interpretations and (perhaps most importantly) a repeatable test case.

First, let's verify that stripping a binary does not affect startup time.

Then let's compare a profile for a stripped versus unstripped binary, so we can see discretely that they spend their time in the same code paths.

### TLDR

This is a long post, so let me summarize:
* Stripping the binary does not appear to significantly affect the benchmarked `wal-g` process duration.
* Profiling can help us understand where the time is spent, so we can efficiently form and test ideas about how to reduce that duration.

Read on for more details.

### Startup time is not affected by stripping the binary

Stripping a binary means removing 2 things:
* the symbol table (and its associated strings table), which contains references that have already been statically linked into the binary.  (Note: The dynamic symbol table is always preserved, because the dynamic linker uses it at runtime to resolve dependencies via dynamically linked libraries.)
* the debug sections, which are typically prefixed with `.debug` or `.zdebug`.

During process startup, neither of those types of sections even get mapped into the process's virtual memory space.  They lack the "alloc" flag (would include `A` in the flags column of the following output) and have no virtual address values.

```
msmiley@saoirse:~$ readelf --sections --wide wal-g.unstripped | grep -e 'Name' -e 'debug' -e '\.symtab' -e '\.strtab'
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [37] .zdebug_aranges   PROGBITS        0000000000000000 2cdbf30 00031c 00      0   0 16
  [38] .zdebug_info      PROGBITS        0000000000000000 2cdc24c 3e4838 00      0   0  1
  [39] .zdebug_abbrev    PROGBITS        0000000000000000 30c0a84 002207 00      0   0  1
  [40] .zdebug_line      PROGBITS        0000000000000000 30c2c8b 1d479b 00      0   0  1
  [41] .zdebug_frame     PROGBITS        0000000000000000 3297426 06dc76 00      0   0  1
  [42] .zdebug_str       PROGBITS        0000000000000000 330509c 0048be 01  MS  0   0  1
  [43] .zdebug_loc       PROGBITS        0000000000000000 330995a 27bb09 00      0   0  1
  [44] .zdebug_ranges    PROGBITS        0000000000000000 3585463 082b31 00      0   0  1
  [45] .debug_gdb_scripts PROGBITS        0000000000000000 3607f94 00002a 00      0   0  1
  [47] .symtab           SYMTAB          0000000000000000 25e2108 1ae908 18     48 72686  8
  [48] .strtab           STRTAB          0000000000000000 2790a10 54b51d 00      0   0  1
```

So during process creation and initialization, the presence of those sections in the binary file should not have any influence on the sequence of events during early process startup.

That's my expectation.  Let's do some measurements to test it.

For context, the following benchmark and profile are running on an idle host (no CPU contention) that has 8 logical processors (4 physical cores each having 2 hyperthreads) and plenty of free memory (no memory pressure).

#### Benchmark

Here we compare 100 runs each of 2 versions of the binary: stripped versus unstripped.

For reference, here are the two versions of the binary:

```
msmiley@saoirse:~$ ./wal-g.unstripped --version
wal-g version v1.1	cadf598	2022.03.16_00:43:43	PostgreSQL

msmiley@saoirse:~$ strip -o wal-g.stripped wal-g.unstripped 

msmiley@saoirse:~$ file wal-g.*
wal-g.stripped:   ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, Go BuildID=nYgJlHQW1CPSDbH-EIL_/220cNtASYETMhk3wBfz4/68HgNRmZ2nPpx_pIn3lg/m3wXyinRW1xzB2hcST7x, BuildID[sha1]=566bd48e75a280d64de670c180b36209daa3b00e, stripped
wal-g.unstripped: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, Go BuildID=nYgJlHQW1CPSDbH-EIL_/220cNtASYETMhk3wBfz4/68HgNRmZ2nPpx_pIn3lg/m3wXyinRW1xzB2hcST7x, BuildID[sha1]=566bd48e75a280d64de670c180b36209daa3b00e, not stripped
```

Benchmark the 2 binaries.

Script:

```
#!/bin/bash

BINARY=$1
[ ! -x "$BINARY" ] && echo "Usage: $0 [walg_binary]" && exit 1

echo "Binary: $BINARY"
# Warm-up run to avoid spurious outlier on first run.
$BINARY --version
for i in $( seq 100 )
do
    time GOMAXPROCS=2 $BINARY --version > /dev/null
done 2>&1 | grep 'real' | sort -Vr | uniq -c
```

Results:

```
msmiley@saoirse:~$ for BINARY in ./wal-g.{,un}stripped ; do echo ; ./benchmark_wal-g.sh $BINARY ; done

Binary: ./wal-g.stripped
wal-g version v1.1	cadf598	2022.03.16_00:43:43	PostgreSQL
      2 real	0m0.019s
      1 real	0m0.018s
     24 real	0m0.017s
     38 real	0m0.016s
     33 real	0m0.015s
      2 real	0m0.014s

Binary: ./wal-g.unstripped
wal-g version v1.1	cadf598	2022.03.16_00:43:43	PostgreSQL
      1 real	0m0.019s
      8 real	0m0.018s
     37 real	0m0.017s
     42 real	0m0.016s
     12 real	0m0.015s
```

The above results show that both binaries have comparable runtime distributions.

### Profiling the `wal-g` startup time

Since stripping does not appear to affect process startup duration, we can more confidently use profiling of the unstripped binary to discover where the time is being spent.

There are several good options for profiling.  Here I will use `perf` rather than `pprof` since we want to include capturing the go runtime's early initialization.

Capture a CPU sampling profile at a rate of approximately one sample per 10 milliseconds.

```
msmiley@saoirse:~$ sudo perf record -g --freq 99 -- ./wal-g.unstripped > /dev/null
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.055 MB perf.data (137 samples) ]

msmiley@saoirse:~$ sudo perf script | stackcollapse-perf.pl --kernel | flamegraph.pl --hash --colors=perl > /tmp/startup_overhead_for_wal-g.unstripped.flamegraph.svg

msmiley@saoirse:~$ firefox /tmp/startup_overhead_for_wal-g.unstripped.flamegraph.svg
```

The resulting flamegraph (below) shows most of the short-lived `wal-g` process's CPU time is spent initialization for OS threads and background GC.  It seems pretty plausible that runtime initialization would drive most of the CPU activity in a case like this where the process exits almost immediately after starting the application code.

startup_overhead_for_wal-g unstripped flamegraph

Does the stripped binary exhibit similar hot code paths?  Yes.

Normally this question would be difficult to answer directly.  But in this case, because the stripped binary's executable and data sections are identical to the unstripped binary, we can borrow the symbols from the unstripped binary to identify the function names in the stack traces of the stripped binary.

Tangent (for reference): It is not usually reliable to assume that the symbols from one binary are usable in another one even if built from the same source code, because compile-time and link-time differences can lead to two binaries having slightly different layouts.  But in this case, both binaries are derived from the same linker output, so we can be confident the symbols are correct.  Also, they share the same gnu-build-id, so the symbols from the unstripped binary are already stored in perf's buildid cache and can be reused by the stripped binary because it shares the same build-id.

```
msmiley@saoirse:~$ sudo perf record -g --freq 99 -- ./wal-g.stripped > /dev/null
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.049 MB perf.data (103 samples) ]

msmiley@saoirse:~$ sudo perf script | stackcollapse-perf.pl --kernel | flamegraph.pl --hash --colors=perl > /tmp/startup_overhead_for_wal-g.stripped.flamegraph.svg
```
startup_overhead_for_wal-g stripped flamegraph

matt-smiley:
In case anyone was wondering why the above profiles do not show any wal-g code...

Here is confirmation that the wal-g application code's `main` function completes very quickly (typically 0.3 ms) for that invocation.

This is another benefit of not stripping the binary -- in addition to generic profiling, we can dynamically add instrumentation to individual functions via uprobe/uretprobe hooks.  This allows us to answer a wide variety of questions.

The following measures wallclock duration [from entry to exit of the `main` function of package `main`](https://github.com/wal-g/wal-g/blob/v1.1/main/pg/main.go#L7-L9).

Start tracing the wal-g main function:

```
msmiley@saoirse:~$ sudo /usr/share/bcc/tools/funcslower --min-ms 0.01 "./wal-g.unstripped:main.main"
Tracing function calls slower than 0.01 ms... Ctrl+C to quit.
COMM           PID    LAT(ms)             RVAL FUNC
wal-g.unstripp 2164323    0.34          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164335    0.27          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164346    0.25          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164358    0.37          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164369    0.24          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164380    0.23          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164391    0.27          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164403    0.28          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164414    0.23          2aba280 ./wal-g.unstripped:main.main 
wal-g.unstripp 2164426    0.27          2aba280 ./wal-g.unstripped:main.main 
```

In another shell, run a toy workload for the uprobe instrumentation to observe:

```
msmiley@saoirse:~$ for i in $( seq 10 ) ; do ./wal-g.unstripped --version > /dev/null ; done
```

fdr:
Nice analysis. But it raises one last question: why did wal-g start so much faster ages ago?

x4m:
> @x4m
> 
> Do we really need some protocol for Postgres at all? Couldn't it be a walg daemon that listens via a UNIX socket and the `wal-push` command that just tells it what file to archive?

The thing is it's not Postgres telling WAL-G which file to archive.
It's Posgres asking WAL-G if the file is archived.
The contract is not that Postgres tells: "Go archive that file." Postgres tells: "Please exit-with-nonzero-code when that files is safely archived."
Surely WAL-G daemon should be able to reply to UNIX socket.

usernamedt:
I've done a little testing using the wal-push daemon mode prototype from [this PR](https://github.com/wal-g/wal-g/pull/1239).

Regular wal-push:
```bash
~ # time $(for i in $(seq 1 100); do bash -c 'wal-g wal-push /var/lib/postgresql/14/data/pg_wal/000000010000000000000064 --config /etc/wal-g/wal-g.yaml'; done)
INFO: 2022/03/22 17:38:13.519623 FILE PATH: 000000010000000000000064.br
INFO: 2022/03/22 17:38:13.962445 FILE PATH: 000000010000000000000064.br
INFO: 2022/03/22 17:38:14.381900 FILE PATH: 000000010000000000000064.br
...
INFO: 2022/03/22 17:38:58.877897 FILE PATH: 000000010000000000000064.br

real	0m45.757s
user	0m12.653s
sys	0m5.050s
```

Daemon mode wal-push:
```
~ # time $(for i in $(seq 1 100); do bash -c '[[ $(nc -U /tmp/wal-push.sock <<< '000000010000000000000064') == "OK" ]] && exit 0 || exit 127'; done)

real	0m35.921s
user	0m0.270s
sys	0m0.102s
```

So I guess that it is definitely possible to remove some overhead using the daemon mode.

CyberDem0n:
> Startup time is not affected by stripping the binary

Interesting, thank you for checking it!

> For a small binary like wal-g

I am sorry, but 40MB (57MB non-striped)  isn't a small binary.
For example, all PostgreSQL v14 binaries (executable files and libraries) with many additional extensions are also taking 57MB.

Yes, profiling could be useful, but let's be honest, what is the percentage of people that do it regularly in production?
Also, in the containerized environment profiling is hardly possible because it requires a privileged mode, which often isn't allowed. That's it, non-stripped binary will make the image bigger without being useful.

fdr:
> So I guess that it is definitely possible to remove some overhead using the daemon mode.
> 
> Also, I think that it is possible to introduce some lightweight wal-g command without any dependencies to get rid of this nc ugliness. Or use the archive module, as @x4m proposes.

Is it overhead, or superior/finer-grained pipelining that has uses more resources available? Granted, both may increase throughput, but cause may be misattributed.

x4m:
I think daemonized version could employ parallelism. While usual WAL-G interface can do parallel upload only for sequential file names, not uploading same file many times.

NikolayS:
Development of daemon sounds really good but it can take long.

There are arguments for both stripping and not stripping the binaries. The common approach for packages is to provide a "debug" package additionally - to keep the main package / binary optimized, since indeed not everyone wants or needs to do it (I personally would prefer to have this ability always, but I also understand those who run everyone in containers these days and has to optimize even if it's ~20 MiB).

So, how about building 2 versions of the binary, always - say, `walg` and `walg-debug`?

xxorde:
> Hopefully daemon mode could alleviate the problem of executable size...

@x4m I really hope to see a good integration of the soon to come interface [archive modules](https://www.postgresql.org/docs/devel/archive-modules.html), this has the potential to be much more performant and reliable than `archive_command`.

But for now, getting the profiling capabilities back would be great as well. :sweat_smile:

x4m:
I think that shipping many binaries is just fine. We already ship many versions, why not add one more?
@matt-smiley can you plz adjust PR so that CI would build both versions on release?

chipitsine:
I came to the similar idea.


I'm fine to merge your PR, however couple of thoughts

1) can we keep default behaviour unchanged, i.e. stripped
2) more definitive flag, not just general EXTRA_FLAG, but a flag with precise purpose. it improves readability

chipitsine:
I added "no stripping" support on make level  https://github.com/wal-g/wal-g/pull/2464

PR #1239, opened 2022-03-22
[draft / prototype] daemon mode for wal-push

usernamedt:
This PR is intended to compare the wal-push effectiveness when using wal-push in daemon mode. 

Usage:
```bash
wal-g wal-push /path/to/pg_wal --config /path/to/config/wal-g.yaml --daemon

# some sort of archive_command
[[ $(nc -U /tmp/wal-push.sock <<< 'WAL_FILE_NAME') == "OK" ]] && exit 0 || exit 127`
```

Binary with this draft is available here https://github.com/usernamedt/wal-g/releases/tag/v0-test-daemon-mode

### DIFF
--- a/cmd/pg/wal_push.go
+++ b/cmd/pg/wal_push.go
@@ -1,6 +1,12 @@
 package pg
 
 import (
+	"net"
+	"os"
+	"path"
+
+	"github.com/wal-g/wal-g/utility"
+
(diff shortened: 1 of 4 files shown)


usernamedt:
Implemented in https://github.com/wal-g/wal-g/pull/1353

Issue #323, opened 2019-07-08
Metrics endpoint


SuperQ:
We (GitLab) have a reasonably large database set (~3TB) that takes a while (4+ hours) to backup with wal-e. We would like to use wal-g, so it would be useful for us to have a Prometheus metrics monitoring endpoint in order to track the status of backups as they run.

We're willing to contribute this code if it would be accepted.

x4m:
Hi! That's very cool!
After review, with tests and docs, it will be happily accepted!

SuperQ:
One question about configuration. We need to listen on a port for metrics requests. Would it be OK to configure one by default and always try to listen? Or should this be off by default?

x4m:
wal-g process life is short, spontaneous and dictated by PostgreSQL archive design. You can create "demonize" command for opening tcp port. I've done so in the prototype of "backup-serve" - command to send wal-g backup to pg_basebackup
See
https://github.com/x4m/wal-g/blob/6e9d0d9491606b3cc3f50a7e1121b6430cadb1a3/internal/backup_fetch_handler.go#L170
https://github.com/x4m/wal-g/blob/6e9d0d9491606b3cc3f50a7e1121b6430cadb1a3/internal/backup_fetch_handler.go#L357
(keep in mind - that functionality if pure PoC and, probably, never will be merged - hence code quality is low)
I think that this "server" command could answer queries about stats, while other calls are updating this stats somewhere..

SuperQ:
"short lived" for us will probably still be many hours for us, as our backup push will be large.

x4m:
I think, eventually, we want to monitor the speed of archival, restoration, the effectiveness of prefetch etc. Also, there may be many PG instances on the same machine.
In our environment, WAL-G can do backups on different hosts: most underutilized replica first, unless it is too lagging.
And we want WAL-G stats from all hosts that possibly use WAL-G.
I'm mostly bothered about unsuccessful restores, cost of archiving and speed of restoration between WAL-G versions: if at some release wal-g is getting slower - I want to see this on the scale of 10k hosts.

Sorry that I bother you with problems of Yandex, but I believe we should take it into design considerations.

x4m:
There is a feature - prefault during WAL apply. I still do not have statistically meaningful data - does this feature actually make sense. I'd setup Prometheus just to see some nice charts about it...

SuperQ:
Thanks, I do need to understand the various design considerations in order to build this properly.

[...]

PR #1240, opened 2022-03-22
Metrics forwarding via statsd protocol

igorwwwwwwwwwwwwwwwwwwww:
### Database name
Mainly targeting postgres, but this feature is database independent.

# Pull request description

It is currently somewhat tricky to monitor the behaviour of wal-g. While it is possible to get a lot of information from the logs, it would be really useful to have proper metrics.

For example, being able to directly monitor the rate at which `wal-g wal-push` is uploading files would be very useful in order to detect stalls, performance issues, and other unexpected behaviour.

This MR can be considered a functional proof-of-concept with a very small set of initial metrics. The main goal is to gather feedback on the approach and flesh this out a bit if there is agreement on the general idea + implementation.

This should also address https://github.com/wal-g/wal-g/issues/323.

Design goals:

- Build on existing libararies and infra.
- Make it easy to add more metrics.
- Interoperate with prometheus.

Chosen design:

- Use the prometheus client library for instrumentation.
- Convert to statsd protocol, and send to statsd server.
- Use statsd_exporter as a drop-in replacement for statsd, re-expose metrics in prometheus format.

Rejected designs:

- Prometheus HTTP metrics server: This does not really work for wal-g, because it is not a long-running process. So we cannot easily bind to an HTTP port.
- Prometheus [pushgateway](https://github.com/prometheus/pushgateway): The pushgateway does not aggregate, so counters simply overwrite each other instead of incrementing.
- Prometheus [aggregation gateway](https://github.com/weaveworks/prom-aggregation-gateway): Not very well maintained. Non-standard protocol.

TODO:

- Implement more metrics that we may want to track. Instrumenting the inner retry loops of storage folder implementations would be interesting.

### Describe what this PR fix

Introduces metrics forwarding via statsd protocol.

To use it, configure `WALG_STATSD_ADDRESS` to point at a [`statsd`](https://github.com/statsd/statsd) or statsd-compatible server. I mainly tested against the prometheus [`statsd_exporter`](https://github.com/prometheus/statsd_exporter).

Example:

```
$ statsd_exporter &

$ WALG_FILE_PREFIX=tmp WALG_STATSD_ADDRESS=localhost:9125 ./main/pg/wal-g wal-push bar

$ curl -s localhost:9102/metrics | grep walg | grep -v '^#'

walg_uploader_uploaded_files_failed_total 0
walg_uploader_uploaded_files_total 1
```

n/a


usernamedt:
Looks good, but have you actually tested if this doesn't affect the `wal-push` execution time?

internal/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+	"time"
+
+	"github.com/cactus/go-statsd-client/v5/statsd"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/spf13/viper"
+	"github.com/wal-g/tracelog"
+)
+
+var (
```
usernamedt, line 15: Let's introduce some common struct, e.g. `metrics` to hold all the current and future metrics?

internal/metrics.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return nil
+}
+
+func writeMetricFamilyToStatsd(client statsd.Statter, in *dto.MetricFamily) error {
+	name := in.GetName()
+	metricType := in.GetType()
+
+	// only export metrics native to wal-g
+	if !strings.HasPrefix(name, "walg_") {
```
usernamedt, line 82: To enforce the common `walg_` metrics prefix I suggest to extract this to const and use during the metric name creation, i.e.  			
```
Name: WalgMetrixPrefix + "uploader_uploaded_files_total",
```

usernamedt (changes requested):

internal/metrics.go
igorwwwwwwwwwwwwwwwwwwww: This is the only one I wasn't sure about. If we want to introduce more low-level metrics for a particular database, we may want to create a separate `metrics.go` inside that database package.

Of course, if you think we'd want a metrics struct per database in this case, I'm happy to introduce the struct.

usernamedt:
> @usernamedt Thanks for the review.
> 
> I attempted to run a benchmark, but it's hard to be sure the right thing is being measured. These were run on a GCP [`e2-standard-4`](https://cloud.google.com/compute/docs/general-purpose-machines#sharedcore) machine with 4 CPUs. I tried to keep noise to a minimum but have no control over frequency scaling and such.
> 
> Metrics are sent best-effort fire-and-forget via UDP, so if the metrics endpoint isn't listening, we simply lose them without extra delay.
> 
> The benchmark runs `WALG_FILE_PREFIX=archive wal-g wal-push wal000000`, where `wal000000` is 16 megs of random data.
> ## TLDR
> 
> I did not measure any significant performance differences with this patch.
> ## Results
> 
> ```shell
> $ hyperfine --warmup 20 --min-runs 100 --setup 'sleep 10' --prepare 'rm -rf archive/*' -n 'master branch' 'WALG_FILE_PREFIX=archive ./wal-g.master wal-push wal000000' -n 'metrics branch, metrics disabled' 'WALG_FILE_PREFIX=archive ./wal-g.metrics wal-push wal000000' -n 'metrics branch, metrics sent to statsd_exporter' 'WALG_STATSD_ADDRESS=localhost:9125 WALG_FILE_PREFIX=archive ./wal-g.metrics wal-push wal000000' -n 'metrics branch, metrics blackholed' 'WALG_STATSD_ADDRESS=localhost:9126 WALG_FILE_PREFIX=archive ./wal-g.metrics wal-push wal000000'
> 
> Benchmark 1: master branch
>   Time (mean ± σ):      81.9 ms ±   4.2 ms    [User: 55.1 ms, System: 51.6 ms]
>   Range (min … max):    73.3 ms … 102.7 ms    100 runs
> 
> Benchmark 2: metrics branch, metrics disabled
>   Time (mean ± σ):      72.0 ms ±   3.9 ms    [User: 43.9 ms, System: 46.5 ms]
>   Range (min … max):    66.7 ms …  92.5 ms    100 runs
> 
> Benchmark 3: metrics branch, metrics sent to statsd_exporter
>   Time (mean ± σ):      74.0 ms ±   4.8 ms    [User: 44.6 ms, System: 47.9 ms]
>   Range (min … max):    67.2 ms …  90.5 ms    100 runs
> 
> Benchmark 4: metrics branch, metrics blackholed
>   Time (mean ± σ):      75.0 ms ±   4.9 ms    [User: 46.4 ms, System: 47.3 ms]
>   Range (min … max):    68.7 ms …  92.3 ms    100 runs
> 
> Summary
>   'metrics branch, metrics disabled' ran
>     1.03 ± 0.09 times faster than 'metrics branch, metrics sent to statsd_exporter'
>     1.04 ± 0.09 times faster than 'metrics branch, metrics blackholed'
>     1.14 ± 0.08 times faster than 'master branch'
> ```
> 
> ## Methodology
> 
> The full `bench.sh` script for reference:
> ## Logs
> 
> The full output from the script for reference:
> 
> ```
> + set -e
> + go install github.com/prometheus/statsd_exporter@latest
> + killall statsd_exporter
> statsd_exporter: no process found
> + true
> + statsd_exporter
> + command -v hyperfine
> /usr/bin/hyperfine
> ++ mktemp -d
> + WORKDIR=/tmp/tmp.FJ8D74427t
> + git stash
> level=info ts=2022-03-29T11:16:55.766Z caller=main.go:291 msg="Starting StatsD -> Prometheus Exporter" version="(version=, branch=, revision=)"
> level=info ts=2022-03-29T11:16:55.767Z caller=main.go:292 msg="Build context" context="(go=go1.17.2, user=, date=)"
> level=info ts=2022-03-29T11:16:55.767Z caller=main.go:341 msg="Accepting StatsD Traffic" udp=:9125 tcp=:9125 unixgram=
> level=info ts=2022-03-29T11:16:55.767Z caller=main.go:342 msg="Accepting Prometheus Requests" addr=:9102
> No local changes to save
> + git checkout master
> Switched to branch 'master'
> Your branch is up to date with 'upstream/master'.
> + git stash pop
> On branch master
> Your branch is up to date with 'upstream/master'.
> 
> Changes not staged for commit:
>   (use "git add <file>..." to update what will be committed)
>   (use "git restore <file>..." to discard changes in working directory)
> 	modified:   go.mod
> 	modified:   go.sum
> 
> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
> 	bench.sh
> 
> no changes added to commit (use "git add" and/or "git commit -a")
> Dropped refs/stash@{0} (0b4f0aed34f9fdfd4619dc56442f488082764e58)
> + go mod vendor
> + make pg_build
> (cd main/pg && go build -mod vendor -tags "brotli" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/pg.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/pg.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/pg.walgVersion=`git tag -l --points-at HEAD`")
> + git stash
> Saved working directory and index state WIP on master: 2888d487 Merge pull request #1246 from wal-g/dependabot/github_actions/actions/cache-3
> + cp main/pg/wal-g /tmp/tmp.FJ8D74427t/wal-g.master
> + git checkout metrics
> Switched to branch 'metrics'
> Your branch is up to date with 'origin/metrics'.
> + go mod vendor
> + make pg_build
> (cd main/pg && go build -mod vendor -tags "brotli" -o wal-g -ldflags "-s -w -X github.com/wal-g/wal-g/cmd/pg.buildDate=`date -u +%Y.%m.%d_%H:%M:%S` -X github.com/wal-g/wal-g/cmd/pg.gitRevision=`git rev-parse --short HEAD` -X github.com/wal-g/wal-g/cmd/pg.walgVersion=`git tag -l --points-at HEAD`")
> + cp main/pg/wal-g /tmp/tmp.FJ8D74427t/wal-g.metrics
> + cd /tmp/tmp.FJ8D74427t
> + dd if=/dev/urandom of=wal000000 bs=1M count=16
> 16+0 records in
> 16+0 records out
> 16777216 bytes (17 MB, 16 MiB) copied, 0.119451 s, 140 MB/s
> + mkdir -p archive archive_status
> + hyperfine --warmup 20 --min-runs 100 --setup 'sleep 10' --prepare 'rm -rf archive/*' -n 'master branch' 'WALG_FILE_PREFIX=archive ./wal-g.master wal-push wal000000' -n 'metrics branch, metrics disabled' 'WALG_FILE_PREFIX=archive ./wal-g.metrics wal-push wal000000' -n 'metrics branch, metrics sent to statsd_exporter' 'WALG_STATSD_ADDRESS=localhost:9125 WALG_FILE_PREFIX=archive ./wal-g.metrics wal-push wal000000' -n 'metrics branch, metrics blackholed' 'WALG_STATSD_ADDRESS=localhost:9126 WALG_FILE_PREFIX=archive ./wal-g.metrics wal-push wal000000'
> Benchmark 1: master branch
>   Time (mean ± σ):      81.9 ms ±   4.2 ms    [User: 55.1 ms, System: 51.6 ms]
>   Range (min … max):    73.3 ms … 102.7 ms    100 runs
> 
> Benchmark 2: metrics branch, metrics disabled
>   Time (mean ± σ):      72.0 ms ±   3.9 ms    [User: 43.9 ms, System: 46.5 ms]
>   Range (min … max):    66.7 ms …  92.5 ms    100 runs
> 
> Benchmark 3: metrics branch, metrics sent to statsd_exporter
>   Time (mean ± σ):      74.0 ms ±   4.8 ms    [User: 44.6 ms, System: 47.9 ms]
>   Range (min … max):    67.2 ms …  90.5 ms    100 runs
> 
> Benchmark 4: metrics branch, metrics blackholed
>   Time (mean ± σ):      75.0 ms ±   4.9 ms    [User: 46.4 ms, System: 47.3 ms]
>   Range (min … max):    68.7 ms …  92.3 ms    100 runs
> 
> Summary
>   'metrics branch, metrics disabled' ran
>     1.03 ± 0.09 times faster than 'metrics branch, metrics sent to statsd_exporter'
>     1.04 ± 0.09 times faster than 'metrics branch, metrics blackholed'
>     1.14 ± 0.08 times faster than 'master branch'
> + curl -s localhost:9102/metrics
> + grep walg
> + grep -v '^#'
> walg_uploader_uploaded_files_failed_total 0
> walg_uploader_uploaded_files_total 120
> + killall statsd_exporter
> level=info ts=2022-03-29T11:18:26.637Z caller=main.go:541 msg="Received os signal, exiting" signal=terminated
> + wait
> ```

Ok, then I guess we can merge this PR after fixing the issues. Also, I think that eventually, we will make [WAL-G daemon mode](https://github.com/wal-g/wal-g/pull/1239) which should help a little.

igorwwwwwwwwwwwwwwwwwwww:
I've introduced a metrics struct and added some documentation.

usernamedt (approved):
LGTM

usernamedt:
@igorwwwwwwwwwwwwwwwwwwww are we ready to merge?

PR #1247, opened 2022-03-28
Introduce sampled pprof profiler

igorwwwwwwwwwwwwwwwwwwww:
This patch allows WAL-G runs to be profiled with golang's pprof profiler.

Options:

- `PROFILE_SAMPLING_RATIO`: A float value between 0 and 1, defines likelihood of the profiler getting enabled. When set to 1, it will always run. This allows probabilistic sampling of invocations. Since WAL-G processes may get created several times per second (e.g. wal-g wal-push), we do not want to profile all of them.
- `PROFILE_MODE`: The type of pprof profiler to use. Can be one of `cpu`, `mem`, `mutex`, `block`, `threadcreation`, `trace`, `goroutine`. Defaults to `cpu`.
- `PROFILE_PATH`: The directory to store profiles in. Defaults to `$TMPDIR`.

Sample invocation:

```bash
$ PROFILE_SAMPLING_RATIO=1 WALG_FILE_PREFIX=tmp ./main/pg/wal-g wal-push bar

2022/03/28 14:21:57 profile: cpu profiling enabled, /var/folders/7z/l7bjk65s5159bp_k60h6hr940000gn/T/profile357452575/cpu.pprof
INFO: 2022/03/28 14:21:57.413146 FILE PATH: bar.lz4
2022/03/28 14:21:57 profile: cpu profiling disabled, /var/folders/7z/l7bjk65s5159bp_k60h6hr940000gn/T/profile357452575/cpu.pprof
```

Sample pprof analysis:

```bash
$ go tool pprof /var/folders/7z/l7bjk65s5159bp_k60h6hr940000gn/T/profile3628604213/cpu.pprof

Type: cpu
Time: Mar 28, 2022 at 2:22pm (CEST)
Duration: 203.34ms, Total samples = 10ms ( 4.92%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 10ms, 100% of 10ms total
      flat  flat%   sum%        cum   cum%
      10ms   100%   100%       10ms   100%  runtime.scanobject
         0     0%   100%       10ms   100%  runtime.gcBgMarkWorker
         0     0%   100%       10ms   100%  runtime.gcBgMarkWorker.func2
         0     0%   100%       10ms   100%  runtime.gcDrain
         0     0%   100%       10ms   100%  runtime.systemstack
```

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -82,6 +82,10 @@ const (
 	StreamSplitterPartitions     = "WALG_STREAM_SPLITTER_PARTITIONS"
 	StreamSplitterBlockSize      = "WALG_STREAM_SPLITTER_BLOCK_SIZE"
 
+	ProfileSamplingRatio = "PROFILE_SAMPLING_RATIO"
+	ProfileMode          = "PROFILE_MODE"
+	ProfilePath          = "PROFILE_PATH"
+
 	MongoDBUriSetting               = "MONGODB_URI"
 	MongoDBLastWriteUpdateInterval  = "MONGODB_LAST_WRITE_UPDATE_INTERVAL"
 	OplogArchiveAfterSize           = "OPLOG_ARCHIVE_AFTER_SIZE"
@@ -246,6 +250,10 @@ var (
 		DeltaFromUserDataSetting:     true,
 		FetchTargetUserDataSetting:   true,
 		SerializerTypeSetting:        true,
+
+		ProfileSamplingRatio: true,
+		ProfileMode:          true,
+		ProfilePath:          true,
 
 		// Swift
 		"WALG_SWIFT_PREFIX": true,
--- a/cmd/common/common.go
+++ b/cmd/common/common.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/cmd/common/st"
 	"github.com/wal-g/wal-g/internal"
 )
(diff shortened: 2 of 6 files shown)


usernamedt (changes requested):

igorwwwwwwwwwwwwwwwwwwww:
Thanks for the review, feedback has been addressed.

internal/config.go
```diff
@@ -82,6 +82,10 @@ const (
 	StreamSplitterPartitions     = "WALG_STREAM_SPLITTER_PARTITIONS"
 	StreamSplitterBlockSize      = "WALG_STREAM_SPLITTER_BLOCK_SIZE"
 
+	ProfileSamplingRatio = "PROFILE_SAMPLING_RATIO"
+	ProfileMode          = "PROFILE_MODE"
+	ProfilePath          = "PROFILE_PATH"
```
igorwwwwwwwwwwwwwwwwwwww, line 87: @usernamedt Do you think we should add a `WALG_` prefix to these?

usernamedt: > @usernamedt Do you think we should add a `WALG_` prefix to these?

I guess that since this is not directly related to the WAL-G functionality, we can leave it as it is.

usernamedt:
Also, it would be cool to have some notes regarding the profiling steps in the documentation.

igorwwwwwwwwwwwwwwwwwwww:
I've added some docs for the profiling options.

usernamedt (approved):

Issue #1036, opened 2021-07-10
FR: Add completion option


EnochTheWise:
... for bash/zsh 8)

x4m:
Can cobra generate it automagically?

usernamedt:
Fixed in https://github.com/wal-g/wal-g/pull/1250

PR #1250, opened 2022-04-03
Bash/zsh shell completions

hexlify:
Added support for static bash/zsh completions
 #1036

### DIFF
--- a/cmd/pg/copy.go
+++ b/cmd/pg/copy.go
@@ -62,4 +62,7 @@ func init() {
 		withoutHistoryShorthand,
 		false,
 		withoutHistoryDescription)
+
+	_ = backupCopyCmd.MarkFlagRequired(toFlag)
+	_ = backupCopyCmd.MarkFlagRequired(fromFlag)
 }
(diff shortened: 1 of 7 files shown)


cmd/pg/copy.go
```diff
@@ -62,9 +62,4 @@ func init() {
 		withoutHistoryShorthand,
 		false,
 		withoutHistoryDescription)
-
-	_ = backupCopyCmd.MarkFlagFilename(toConfigFile)
```
serprex, line 66: Why remove this?

hexlify: 1. These calls don't affect anything. Both `MarkFlagFilename` and `MarkFlagRequired` require flag name but flag value was provided. Just a bug I guess
2. `MarkFlagFilename` call without extension arg is pointless
3. Seems like toFlag and fromFlag actually aren't required. `ReadConfigFromFile` handles empty `configFile` argument

usernamedt: Well, wal-g copy command definetly requires both `fromFlag` and `toFlag` to be specified. So I don't see any reason to remove this logic. Let's just fix the incorrect argument, e.g. change it to this:
```
	_ = backupCopyCmd.MarkFlagRequired(toFlag)
	_ = backupCopyCmd.MarkFlagRequired(fromFlag)
```

usernamedt (changes requested):

usernamedt (approved):

PR #1268, opened 2022-05-17
sqlserver: blob read cache

mialinx:
Cache last downloaded and decrypted block in memory
to avoid read amplifications during small range reads.

### DIFF
--- a/internal/databases/sqlserver/blob/server.go
+++ b/internal/databases/sqlserver/blob/server.go
@@ -13,12 +13,15 @@ import (
 	"sync"
 	"time"
 
+	"github.com/wal-g/wal-g/utility"
+
 	"github.com/wal-g/wal-g/internal/compression"
 	"github.com/wal-g/wal-g/internal/crypto"
 	"golang.org/x/xerrors"
 
 	"github.com/gofrs/flock"
 	"github.com/google/uuid"
+	lru "github.com/hashicorp/golang-lru"
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
 	"github.com/wal-g/wal-g/pkg/storages/storage"
@@ -29,6 +32,10 @@ const ProxyStartTimeout = 10 * time.Second
 const ReqIDHeader = "X-Ms-Request-Id"
 
 const DefaultConcurrency = 8
+
+const BlockReadCacheSize = 16
+
+const MaxCacheBlockSize = 16 * 1024 * 1024 // 16M
 
 type Server struct {
 	folder       storage.Folder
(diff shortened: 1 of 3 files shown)


internal/databases/sqlserver/blob/server.go
```diff
@@ -783,3 +791,52 @@ func (bs *Server) validateBlobCompressionEncryption(idx *Index) error {
...
+		r, err := folder.ReadObject(s.Path)
+		if err != nil {
+			return nil, err
+		}
+		dr, err := bs.decompressDecryptIfNeeded(idx, r)
+		if err != nil {
+			utility.LoggedClose(r, "failed to close block reader")
+			return nil, err
+		}
+		defer dr.Close()
```
ostinru, line 821: NIT: you can add `utility.LoggedClose` to wrap `dr.Close()`

internal/databases/sqlserver/blob/server.go
```diff
@@ -94,6 +102,18 @@ func NewServer(folder storage.Folder) (*Server, error) {
...
+		tracelog.DebugLogger.Printf("EVICT_CACHE: %s", k)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create lru cache: %v", err)
+	}
+	bs.readCache = c
+	go func() {
+		// to release memory after restore
+		time.Sleep(time.Minute)
+		bs.readCache.RemoveOldest()
```
ostinru, line 115: It is a little bit weird that you are going to remove _one_ record from the cache after one minute...

ostinru (approved):

internal/databases/sqlserver/blob/server.go
mialinx: This is poor man's  expiration for lru cache :)
When database is restored, cache is no longer needed, and we can free some memory.
But it's hard to say when exactly restore is done, so we just purge cache slowly one by one item.

internal/databases/sqlserver/blob/server.go
mialinx: fixed

internal/databases/sqlserver/blob/server.go
ostinru: I think that in LRU cache old pages will be evicted without your explicit actions.
When we are talking about memory - I think that golang is greedy and won't return a byte to the operation system when you free some bytes in your app.

mialinx: They won't. LRU cache doesn't know about time and only evict items when new ones added.
Golang can return memory back to OS (unless it was fragmented, of course).

ostinru: Hm... It seems that you are right: golang authors say that since 1.12
>> The Go runtime now releases memory back to the operating system more aggressively, particularly in response to large allocations that can't reuse existing heap space.

PR #1275, opened 2022-05-24
storage tools: check read / check write tools

mialinx:
Simple tools to verify read / write access to the configured storage.
It's handy to distinguish between storage configuration and actual backup errors


cmd/common/st/check.go
```diff
@@ -0,0 +1,40 @@
...
+
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "check access to the storage",
+}
+
+var checkReadCmd = &cobra.Command{
+	Use:   "read",
+	Short: "check read access to the storage",
+	Args:  cobra.MinimumNArgs(0),
```
usernamedt, line 18: `cobra.MinimumNArgs(1)`?

cmd/common/st/check.go
```diff
@@ -0,0 +1,40 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		folder, err := internal.ConfigureFolder()
+		tracelog.ErrorLogger.FatalOnError(err)
+		storagetools.HandleCheckRead(folder, args)
+	},
+}
+
+var checkWriteCmd = &cobra.Command{
+	Use:   "write",
+	Short: "check write access to the storage",
```
usernamedt, line 28: `Args: cobra.NoArgs?`

internal/storagetools/check.go
```diff
@@ -0,0 +1,61 @@
...
+	"strings"
+	"time"
+
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+
+	"github.com/wal-g/tracelog"
+)
+
+func HandleCheckRead(folder storage.Folder, filenames []string) {
+	objs, _, err := folder.ListFolder()
```
usernamedt, line 16: This will only list the objects of the root folder. I guess that it will not work if the filename is `folder1/subfolder/file`.

internal/storagetools/check.go
```diff
@@ -0,0 +1,61 @@
...
+	"math/rand"
+	"strings"
+	"time"
+
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+
+	"github.com/wal-g/tracelog"
+)
+
+func HandleCheckRead(folder storage.Folder, filenames []string) {
```
usernamedt, line 15: Maybe name it like `check exists`? Because it seems like the actual behavior of this command.

internal/storagetools/check.go
```diff
@@ -0,0 +1,61 @@
...
+		ok, err := folder.Exists(filename)
+		if err != nil {
+			tracelog.ErrorLogger.Fatalf("failed to read from the storage: %v", err)
+		}
+		if !ok {
+			break
+		}
+	}
+	err := folder.PutObject(filename, bytes.NewBufferString("test"))
+	_ = folder.DeleteObjects([]string{filename})
```
usernamedt, line 56: Maybe add a warning to notify the user that we've left some garbage in the storage?

usernamedt (changes requested):

cmd/common/st/check.go
mialinx: no, checking particular files is optional. 
we may check that storage is readable without any particular files

internal/storagetools/check.go
mialinx: file existance check is optional

internal/storagetools/check.go
mialinx: ok

internal/storagetools/check.go
mialinx: well.. yes.
It's not required for my case, as all files are supposed to be in one folder.
```
WALG_S3_PREFIX: "s3://se-test/subfolder"
```
may we implement subfolder checks later ?

mialinx: fixed

internal/storagetools/check.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+	err := folder.PutObject(filename, bytes.NewBufferString("test"))
+	if folder.DeleteObjects([]string{filename}) != nil {
+		tracelog.WarningLogger.Printf("failed to clean temp files, %s left in storage", filename)
+	}
+	if err != nil {
+		tracelog.ErrorLogger.Fatalf("failed to write to the storage: %v", err)
+	}
+	fmt.Printf("OK")
```
usernamedt, line 65:
```suggestion
	tracelog.InfoLogger.Println("Write check OK")
```

usernamedt: We tend not to use `fmt.Print*` stuff in WAL-G output.

internal/storagetools/check.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		ok, err := sf.Exists(fname)
+		if err != nil || !ok {
+			missing = append(missing, name)
+		}
+	}
+	if len(missing) > 0 {
+		tracelog.ErrorLogger.Fatalf("files are missing: %s", strings.Join(missing, ", "))
+	}
+	fmt.Printf("OK")
```
usernamedt, line 36:
```suggestion
	tracelog.InfoLogger.Println("Read check OK")
```

internal/storagetools/check.go
```diff
@@ -0,0 +1,66 @@
...
+		tracelog.ErrorLogger.Fatalf("failed to list the storage: %v", err)
+	}
+	missing := make([]string, 0)
+	for _, name := range filenames {
+		fdir, fname := path.Split(name)
+		sf := folder
+		if fdir != "" && fdir != "/" {
+			sf = folder.GetSubFolder(fdir)
+		}
+		ok, err := sf.Exists(fname)
```
usernamedt, line 28: I guess that the regular `sf.Exists` will do the job without the need for subfolder check.

cmd/common/st/check.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/wal-g/wal-g/internal/storagetools"
+)
+
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "check access to the storage",
+}
+
+var checkReadCmd = &cobra.Command{
+	Use:   "read",
```
usernamedt, line 16:
```suggestion
	Use:   "read [filename1 filename2 ...]",
```

usernamedt (approved):

PR #1281, opened 2022-05-31
mysql: binlog cache fixes

mialinx:
* use numeric binlog comparison
* clear binlogs cache after reset master or host resetup

### DIFF
--- a/internal/databases/mysql/binlog_push_handler.go
+++ b/internal/databases/mysql/binlog_push_handler.go
@@ -9,7 +9,6 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
-	"sort"
 
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


internal/databases/mysql/binlog_push_handler.go
```diff
@@ -37,21 +36,28 @@ func HandleBinlogPush(uploader internal.UploaderProvider, untilBinlog string, ch
...
 
 	var binlogSentinelDto BinlogSentinelDto
 	err = FetchBinlogSentinel(rootFolder, &binlogSentinelDto)
-	// copy MySQLBinlogSentinel to cache:
-	cache := getCache()
 	if err == nil && binlogSentinelDto.GTIDArchived != "" {
 		tracelog.InfoLogger.Printf("fetched binlog archived GTID SET: %s\n", binlogSentinelDto.GTIDArchived)
 	}
+	cache := getCache()
+	if len(binlogs) > 0 && cache.LastArchivedBinlog != "" {
```
ostinru, line 53: `len(binlogs) > 0` is always `true` here - when `binlogs` slice is empty - it will crash in line 43

mialinx: `len(binlogs) > 0` is always `true` here - when `binlogs` slice is empty - it will crash in line 43

teem0n (approved):

PR #1186, opened 2021-12-22
Update LazyCache to be generic

serprex:
Add Key/Value type parameters

Increases minimum go version to 1.18

PR #1282, opened 2022-06-01
Azure authentication using the default credential chain

tommyzli:
### Database name
Postgres

# Pull request description
Support the standard Azure authentication methods. Closes https://github.com/wal-g/wal-g/issues/1284

### Describe what this PR fix
wal-g requires either the `AZURE_STORAGE_ACCESS_KEY` or `AZURE_STORAGE_SAS_TOKEN` environment variables to be set and bails out early if they are not. This breaks two methods of authentication
- [managed identities](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) attached to a vm running inside of azure
- the standard azure [environment variables](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#-option-1-define-environment-variables)

Before my patch:
```
2022-05-31 20:41:52.238 UTC [75] LOG:  archive command failed with exit code 1
2022-05-31 20:41:52.238 UTC [75] DETAIL:  The failed archive command was: AZURE_STORAGE_ACCOUNT=myaccount WALG_AZ_PREFIX="azure://mycontainer/foo" wal-g wal-push "pg_wal/000000010000000000000001"
ERROR: 2022/05/31 20:41:53.274696 failed to configure folder: Azure error : AZURE_STORAGE_ACCESS_KEY setting is not set: Credential error
```

### DIFF
--- a/pkg/storages/azure/folder.go
+++ b/pkg/storages/azure/folder.go
@@ -154,14 +154,13 @@ func configureAuthType(settings map[string]string) (AzureAuthType, string, strin
 }
 
 func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder, error) {
-	var accountName, accountKey, accountToken, storageEndpointSuffix string
+	var accountName, storageEndpointSuffix string
 	var ok bool
-	var authType AzureAuthType
 	if accountName, ok = settings[AccountSetting]; !ok {
 		return nil, NewCredentialError(AccountSetting)
 	}
 
-	authType, accountToken, accountKey = configureAuthType(settings)
+	authType, accountToken, accountKey := configureAuthType(settings)
 
 	var credential *azblob.SharedKeyCredential
 	var err error
@@ -294,6 +293,11 @@ func (folder *Folder) ReadObject(objectRelativePath string) (io.ReadCloser, erro
 
 	get, err := blobClient.Download(context.Background(), nil)
 	if err != nil {
+		var storageError *azblob.StorageError
+		errors.As(err, &storageError)
+		if storageError.ErrorCode == azblob.StorageErrorCodeBlobNotFound {
+			return nil, storage.NewObjectNotFoundError(path)
+		}
 		return nil, NewFolderError(err, "Unable to download blob %s.", path)
 	}
 	reader := get.Body(nil)
@@ -331,11 +335,11 @@ func (folder *Folder) CopyObject(srcPath string, dstPath string) error {
 	var srcClient, dstClient *azblob.BlockBlobClient
 	srcClient, err = folder.containerClient.NewBlockBlobClient(srcPath)
 	if err != nil {
-		return NewFolderError(err, "Unable to init Azure Blob client for copy source %v", srcPath)
+		return NewFolderError(err, "Unable to init Azure Blob client for copy source %s", srcPath)
 	}
 	dstClient, err = folder.containerClient.NewBlockBlobClient(dstPath)
 	if err != nil {
-		return NewFolderError(err, "Unable to init Azure Blob client for copy destination %v", dstPath)
+		return NewFolderError(err, "Unable to init Azure Blob client for copy destination %s", dstPath)
 	}
 	_, err = dstClient.StartCopyFromURL(context.Background(), srcClient.URL(), &azblob.BlobStartCopyOptions{Tier: azblob.AccessTierHot.ToPtr()})
 	return err


pkg/storages/azure/folder.go
```diff
@@ -124,7 +124,7 @@ func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder,
 
 	var containerUrlString string
 	var containerClient azblob.ContainerClient
-	if usingToken {
+	if usingToken && !usingAccessKey {
```
usernamedt, line 127: Can we introduce some type to define the chosen auth method to make this logic more straightforward?

For example:

```
type AzureAuthType string

const ( 
AzureTokenAuth AzureAuthType = "AzureTokenAuth"
AzureAccessKeyAuth AzureAuthType = "AzureAccessKeyAuth"
)
```

So we can separate the auth logic by the `AzureAuthType`.

usernamedt (changes requested):

usernamedt:
> Not sure why the linter triggered on my PR, the `x/crypto/openpgp` package has been deprecated for some time now

Yeah, looks like this is not relevant to the changes from this PR.

usernamedt:
> Not sure why the linter triggered on my PR, the `x/crypto/openpgp` package has been deprecated for some time now



Yeah, looks like this is not relevant to the changes from this PR.

pkg/storages/azure/folder.go
tommyzli: makes sense, pushed a change

serprex (approved):

serprex:
You can disable SA1019: https://github.com/golangci/golangci-lint/issues/2626#issuecomment-1054012392

pkg/storages/azure/folder.go
```diff
@@ -124,7 +133,7 @@ func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder,
 
 	var containerUrlString string
 	var containerClient azblob.ContainerClient
-	if usingToken {
+	if authType == AzureSASTokenAuth {
```
usernamedt, line 136: Now it looks like a good place to extract the logic inside the `if` branches into separate functions to improve readability for future contributors.

pkg/storages/azure/folder.go
```diff
@@ -74,15 +82,16 @@ func NewFolder(
 
 func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder, error) {
 	var accountName, accountKey, accountToken, storageEndpointSuffix string
-	var ok, usingToken bool
+	var ok bool
+	var authType AzureAuthType
```
usernamedt, line 86: Let's extract the `AzureAuthType` decision logic to the separate function, e.g:
```
authType, err := configureAuthType(settings)
```

usernamedt (changes requested):
Looks good to me, but I guess it would be cool to merge this after a little refactoring.

serprex:
@tommyzli for some reason when I tried using the sdk's Download method it would fail to download the parts properly. & ofc the sdk has multiple abstraction layers to obfuscate why direct usage of http methods worked while the sdk's code didn't

serprex:
#1186 updates wal-g to go 1.18

pkg/storages/azure/folder.go
```diff
@@ -384,27 +319,39 @@ func (folder *Folder) PutObject(name string, content io.Reader) error {
...
-	dstClient := folder.containerClient.NewBlockBlobClient(dstPath)
-	_, err := dstClient.StartCopyFromURL(context.Background(), srcClient.URL(), &azblob.StartCopyBlobOptions{Tier: azblob.AccessTierHot.ToPtr()})
+	var srcClient, dstClient *azblob.BlockBlobClient
+	srcClient, err = folder.containerClient.NewBlockBlobClient(srcPath)
+	if err != nil {
+		return NewFolderError(err, "Unable to init Azure Blob client for copy source %v", srcPath)
+	}
+	dstClient, err = folder.containerClient.NewBlockBlobClient(dstPath)
+	if err != nil {
+		return NewFolderError(err, "Unable to init Azure Blob client for copy source %v", srcPath)
```
usernamedt:
```suggestion
		return NewFolderError(err, "Unable to init Azure Blob client for copy destination %s", dstPath)
```

pkg/storages/azure/folder.go
```diff
@@ -215,166 +278,38 @@ func (folder *Folder) ListFolder() (objects []storage.Object, subFolders []stora
...
-
+	get, err := blobClient.Download(context.Background(), nil)
 	if err != nil {
 		return nil, NewFolderError(err, "Unable to download blob %s.", path)
 	}
-
-	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		resp.Body.Close()
-		if resp.StatusCode == 404 {
-			return nil, storage.NewObjectNotFoundError(path)
```
usernamedt, line 363: I don't see the reason why the 404 handling logic has been removed. I guess we still need to distinguish the 404 and return an appropriate `storage.NewObjectNotFoundError()` error.

usernamedt (changes requested):

pkg/storages/azure/folder.go
```diff
@@ -58,40 +63,109 @@ func NewCredentialError(settingName string) storage.Error {
...
 	}
 
+	return authType, accountToken, accessKey
+}
+
+func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder, error) {
+	var accountName, accountKey, accountToken, storageEndpointSuffix string
+	var ok bool
+	var authType AzureAuthType
+	if accountName, ok = settings[AccountSetting]; !ok {
```
usernamedt, line 160: I suggest using the ':=' instead to avoid the unnecessary declaration above.

pkg/storages/azure/folder.go
```diff
@@ -58,40 +63,109 @@ func NewCredentialError(settingName string) storage.Error {
...
+		return nil, NewCredentialError(AccountSetting)
+	}
+
+	authType, accountToken, accountKey = configureAuthType(settings)
```
usernamedt, line 164: Same here

usernamedt (changes requested):

pkg/storages/azure/folder.go
tommyzli: The accountName and ok declarations are actually necessary, but I'm not familiar enough with go to articulate why. The build fails with `accountName declared but not used`, `undefined: accountName`, and `undefined: ok` errors when I change it.

pkg/storages/azure/folder.go
tommyzli: sorry, didn't realize 404s were specifically meaningful

serprex (approved):

usernamedt:
It is sad that we don't (yet) have any Azure storage-related tests in our CI pipeline. It would be cool to add some simple tests, for example with [Azurite](https://github.com/Azure/Azurite) to our CI.

usernamedt (approved):

PR #1285, opened 2022-06-02
Initial mvp mongodb binary backup

KhurtinDN:
Added binary backup for mongodb.


internal/limited_reader.go
```diff
@@ -0,0 +1,33 @@
+package internal
+
+import (
+	"io"
+)
+
+type LimitedReader struct {
```
usernamedt, line 7: We already have an `io.LimitedReader`, let's use it instead?

internal/databases/mongo/logical/storage_utils.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/wal-g/wal-g/internal/databases/mongo/common"
+	"github.com/wal-g/wal-g/internal/databases/mongo/models"
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+)
+
+func DownloadSentinel(folder storage.Folder, backupName string) (*models.Backup, error) {
+	var sentinel models.Backup
+	err := common.DownloadSentinel(folder, backupName, &sentinel)
+	if sentinel.BackupName == "" {
```
usernamedt, line 12: Better check the error and return it early to avoid some malformed struct being returned.
```
if err != nil {
    return nil, err
}
```

internal/databases/mongo/binary/utilities.go
```diff
@@ -0,0 +1,30 @@
+package binary
+
+import (
+	"github.com/pkg/errors"
+	"github.com/wal-g/tracelog"
+	"golang.org/x/mod/semver"
+)
+
+func MajorMinorVersion(v string) string {
```
usernamedt, line 9: Maybe add some comment regarding this logic? It is not completely clear.

internal/databases/mongo/binary/storage_utils.go
```diff
@@ -0,0 +1,12 @@
+package binary
+
+import (
+	"github.com/wal-g/wal-g/internal/databases/mongo/common"
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+)
+
+func DownloadSentinel(folder storage.Folder, backupName string) (*MongodBackupMeta, error) {
+	var sentinel MongodBackupMeta
+	err := common.DownloadSentinel(folder, backupName, &sentinel)
```
usernamedt, line 10: Same issue as in `internal/databases/mongo/logical/storage_utils.go`.

internal/databases/mongo/binary/restore.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return err
+	}
+
+	tracelog.InfoLogger.Println("Download backup files to dbPath")
+	return restoreService.DownloadFilesFromBackup(mongodBackupMeta)
+}
+
+func (restoreService *RestoreService) DownloadFilesFromBackup(backupMeta *MongodBackupMeta) error {
+	err := restoreService.LocalStorage.CreateDirectories(backupMeta.BackupDirectories)
```
usernamedt, line 56: Maybe add a check that the target directory is empty to avoid some unexpected results?

internal/databases/mongo/binary/mongod.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/wal-g/wal-g/internal"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+const adminDB = "admin"
+
+type MongodService struct {
```
usernamedt, line 14: What is the difference between `MongodService` and `MongoClient`? Maybe state explicitly that it is a wrapper and embed the mongo.Client into it? For example:


```
type MongoClientWrapper struct {
ctx context.Context
mongo.Client
}
```

internal/databases/mongo/binary/checksum.go
```diff
@@ -0,0 +1,32 @@
+package binary
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"hash"
+)
+
+type ChecksumCalculator struct {
```
usernamedt, line 9: Can we extract this directly to the `internal` package? It looks like a not mongo-specific code.

internal/databases/mongo/binary/backup.go
```diff
@@ -0,0 +1,319 @@
...
+		UncompressedSize: backupFileStat.Size(), // backupFile.FileSize contains 0 for extended backup cursor
+	}
+	backupMeta.BackupFiles = append(backupMeta.BackupFiles, &backupFileMeta)
+	backupMeta.UncompressedDataSize += backupFile.FileSize
+	backupService.BackupFilesMap[backupFilePath] = &backupFileMeta
+
+	return nil
+}
+
+func processDoubleEncounterTheSameBackupFile(backupFileMeta *BackupFileMeta, backupFile BackupCursorFile) error {
```
usernamedt, line 255: Do I correctly understand that we might encounter some files multiple times during the backup-push? When does it happen?

internal/databases/mongo/binary/backup.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	backupCursor, err := backupService.MongodService.GetBackupCursor()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err != nil {
+			closeErr := backupCursor.Close(backupService.Context)
+			tracelog.ErrorLogger.Printf("Unable to close backup cursor: %+v", closeErr)
```
usernamedt, line 63: You can use `utility.LoggedClose`

internal/databases/mongo/binary/backup.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	closeChannel := make(chan string)
+
+	go func() {
+		ticker := time.NewTicker(time.Minute * 1)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-closeChannel:
```
usernamedt, line 107: Maybe add a check for the `BackupService` context?

internal/databases/mongo/backup_list_handler.go
```diff
@@ -1,35 +1,177 @@
...
+
+func NewLogicalBackupDetail(backupTime internal.BackupTime, sentinel models.Backup) *BackupDetail {
+	return &BackupDetail{
+		BackupName:      backupTime.BackupName,
+		ModifyTime:      backupTime.Time,
+		TsStart:         sentinel.MongoMeta.Before.LastMajTS,
+		TsEnd:           sentinel.MongoMeta.After.LastMajTS,
+		StartLocalTime:  sentinel.StartLocalTime,
+		FinishLocalTime: sentinel.FinishLocalTime,
+		//UncompressedSize: sentinel.UncompressedDataSize,
```
usernamedt, line 62: Uncomment these lines?

cmd/mongo/binary_backup_push.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const binaryBackupPushCommandName = "binary-backup-push"
+
+var binaryBackupPushCmd = &cobra.Command{
+	Use:   binaryBackupPushCommandName,
+	Short: "Creates mongodb binary backup and pushes it to storage without local disk",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx, cancel := context.WithCancel(context.Background())
```
usernamedt, line 21: Too much logic for the command Run function, I guess you might want to move it to a corresponding handler in the `internal/databases/mongo`.

cmd/mongo/binary_backup_fetch.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+const binaryBackupFetchCommandName = "binary-backup-fetch"
+
+var binaryBackupFetchCmd = &cobra.Command{
+	Use:   binaryBackupFetchCommandName + " <backup name> <replSetName> <mongod dbpath> <mongod version>",
+	Short: "Fetches a mongodb binary backup from storage and restores it in mongodb storage dbPath",
+	Args:  cobra.ExactArgs(4),
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx, cancel := context.WithCancel(context.Background())
```
usernamedt, line 21: Too much logic for the command Run function, I guess you might want to move it to a corresponding handler in the internal/databases/mongo.

usernamedt (changes requested):

internal/databases/mongo/binary/mongod.go
KhurtinDN: I want to incapsulate work with mongo server in this Service "class"

internal/databases/mongo/binary/backup.go
KhurtinDN: yes, I know it and use it, but in this case backupCursor have Close method with arg (LoggedClose can call Close method without arg)

internal/databases/mongo/binary/backup.go
KhurtinDN: Sorry, I don't understand. Can you give me fragment of example?

internal/databases/mongo/backup_list_handler.go
KhurtinDN: I can't do it, because for logic backup I should implement it. I don't want to do it right now (in initial mvp).

usernamedt: OK, I would suggest to add comment above to avoid confusion

internal/databases/mongo/binary/mongod.go
```diff
@@ -0,0 +1,214 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func replaceData(ctx context.Context, collection *mongo.Collection, drop bool, insertData bson.M) error {
+	findCursor, err := collection.Find(ctx, bson.D{})
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("unable to find all from %v", collection.Name()))
```
usernamedt, line 151: `errors.Wrap` is deprecated, we use `fmt.Errorf` in the new code.

internal/databases/mongo/binary/mongod_runner.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+
+	cliArgs := []string{"--port", strconv.Itoa(unusedPort), "--config", configFilePath}
+	for _, parameter := range mongodProcess.parameters {
+		cliArgs = append(cliArgs, "--setParameter", parameter)
+	}
+
+	if _, err := os.Stat("/var/log/mongodb"); err == nil {
```
usernamedt, line 83: It is better to make this and other hard-coded paths configurable via the config option.

internal/databases/mongo/binary/mongod_runner.go
```diff
@@ -0,0 +1,117 @@
...
+
+func randomUnusedPort() (int, error) {
+	listen, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return 0, errors.Wrap(err, "unable to choose random unused port")
+	}
+	defer utility.LoggedClose(listen, "unable to close listen")
+
+	address := listen.Addr().String()
+	lastColonIndex := strings.LastIndex(address, ":")
```
usernamedt, line 110: If you know that the address is tcp, you can just write something like this (maybe add some error handling too):
```
addr.(*net.TCPAddr).Port
```

internal/databases/mongo/binary/dto.go
```diff
@@ -0,0 +1,29 @@
...
+	FileMode         os.FileMode `json:"FileMode"`
+	Compression      string      `json:"Compression"`
+	CompressedSize   int64       `json:"CompressedSize"`
+	UncompressedSize int64       `json:"UncompressedSize"`
+	Checksum         Checksum    `json:"Checksum"`
+}
+
+type MongodBackupFilesMetadata struct {
+	BackupDirectories []*BackupDirectoryMeta `json:"BackupDirectories,omitempty"`
+	BackupFiles       []*BackupFileMeta      `json:"BackupFiles,omitempty"`
```
usernamedt, line 28: Why not store it like this since we convert them to maps in the BackupService anyway?

```
	BackupDirectoriesMap map[string]*BackupDirectoryMeta
	BackupFilesMap       map[string]*BackupFileMeta
```

usernamedt: Also, is it really necessary to divide the directories and files? Also, what about symlinks, etc? Why not store everything in the same place? We have a FileMode anyway.

internal/databases/mongo/binary/storage.go
```diff
@@ -0,0 +1,162 @@
...
+func (backupStorage *BackupStorage) CalculateCompressedFiles(backupFiles map[string]*BackupFileMeta) (int64, error) {
+	folder := backupStorage.UploaderProvider.Folder().GetSubFolder(backupStorage.BackupName)
+	return backupStorage.calculateCompressedFiles("", folder, backupFiles)
+}
+
+//nolint: whitespace
+func (backupStorage *BackupStorage) calculateCompressedFiles(relativePath string, folder storage.Folder,
+	backupFiles map[string]*BackupFileMeta) (backupSize int64, err error) {
+
+	objects, subFolders, err := folder.ListFolder()
```
usernamedt, line 105: Why list folder to calculate the size? We can take this information from the appropriate `Uploader`

internal/databases/mongo/models/backup.go
```diff
@@ -1,27 +1,35 @@
...
 
-import "time"
+import (
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
 
 // Backup represents backup sentinel data
+// todo: use `internal.GenericMetadata` as base
```
usernamedt, line 10: It might be useful to write the implementation for [GenericMetaFetcher/GenericMetaSetter](https://github.com/wal-g/wal-g/blob/3ffb8d522d45242f14ef9e69d6b754f15830e5d2/internal/generic_metadata.go#L53), for example like it has been done [in Greenplum](https://github.com/wal-g/wal-g/blob/master/internal/databases/greenplum/generic_meta_interactor.go) to use the common code for `backup-mark`, backup selectors, etc.

usernamedt (changes requested):

internal/databases/mongo/binary/dto.go
usernamedt: On the other hand, if we plan to store the backup files in TARs in the future versions, this might be not so important since we will rewrite this code anyway.

internal/databases/mongo/backup_list_handler.go
KhurtinDN: I fix it

internal/databases/mongo/archive/backup.go
```diff
@@ -103,11 +25,13 @@ type MongoMetaConstructor struct {
...
 	meta := m.Meta()
 	backupSentinel := &models.Backup{
+		BackupName:      meta.BackupName,
+		BackupType:      common.LogicalBackupType,
 		StartLocalTime:  meta.StartTime,
 		FinishLocalTime: meta.FinishTime,
 		UserData:        meta.User,
 		MongoMeta:       meta.Mongo,
-		DataSize:        meta.DataSize,
+		CompressedSize:  meta.CompressedSize,
```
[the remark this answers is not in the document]
KhurtinDN: added it

internal/databases/mongo/archive/backup.go
```diff
@@ -159,8 +89,9 @@ func (m *MongoMetaConstructor) Finalize(backupName string) error {
 		LastTS:    lastTS,
 		LastMajTS: lastMajTS,
 	}
+	m.meta.BackupName = backupName
 	m.meta.FinishTime = utility.TimeNowCrossPlatformLocal()
-	m.meta.DataSize = dataSize
+	m.meta.CompressedSize = dataSize
```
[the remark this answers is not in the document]
KhurtinDN: to do this I should write some code for logical backup, but I write code only for binary. May be this functionality will add by feature request?

internal/databases/mongo/binary/backup.go
KhurtinDN: > I mean, what happens if the backupService.Context gets canceled?
then to closeChannel should send "Game Over"?

internal/databases/mongo/binary/mongod.go
KhurtinDN: Why? errors. Wrap will take stacktrace for good troubleshooting

internal/databases/mongo/binary/dto.go
KhurtinDN: > Why not store it like this since we convert them to maps in the BackupService anyway?

map needed to detect "visited" situation and get visited "value" during backup stage only. key of map is path, that already exists in value (so this redundant information in meta files).
map is small overhead.

> Also, is it really necessary to divide the directories and files? Also, what about symlinks, etc? Why not store everything in the same place? We have a FileMode anyway.

This is for convenience, because of I process files and directories separately. and files meta information contains more fields. and I see so this life =) or it is essential problem? 

> On the other hand, if we plan to store the backup files in TARs in future versions, this might be not so important since we will rewrite this code anyway.

I don't know =)

internal/databases/mongo/models/backup.go
KhurtinDN: I have created new task for me to implement it, but I don't want implement it in this PR, because this functionality was implemented before me and not connected with binary feature directly (it is refactoring and should be in separate pr).

internal/databases/mongo/binary/storage.go
KhurtinDN: I implement like internal.FolderSize, that is used in logical backup. 
Besides I set compressed size for each file to save in meta (for troubleshooting purpose).
I of cause can remove it and use uploader.UploadedDataSize(), but I will be forced to remove compressed size for each my file.

internal/databases/mongo/binary/mongod.go
usernamedt: > Why? errors. Wrap will take stacktrace for good troubleshooting

`errors` package is no longer maintained. If you want to keep stacktraces, use `errors.Errorf`.

internal/databases/mongo/binary/storage.go
usernamedt: Since we upload files one by one in the current version, these S3 listings might be massive.

usernamedt (approved):

perekalov (approved):

docs/MongoDB.md
```diff
@@ -101,14 +101,31 @@ WAL-G mongodb extension currently supports these commands:
...
+
+Creates new binary backup and send it to storage.
+
+This command uses [backup cursors functionality of WiredTiger](https://source.wiredtiger.com/develop/backup.html)
+to get list of files in local file system, that stay locked while backup cursor is opened.
+Applications may continue to read and write the databases while a binary snapshot is taken.
+
+Note: In this mvp version we support replicaset cluster only (non sharded).
+
+This functionality inspired by article
```
dfrolov95, line 122: Hello!
Thanks for the functionality!

I think it's worth pointing out explicitly that this feature will not work for the community version of MongoDB.

dfrolov95:
Hello!
Thanks for the functionality!

I think it's worth pointing out explicitly that this feature will not work for the community version of mongodb.

Issue #1291, opened 2022-06-10
wal-g delete fails on WAL-E backups


OrangeDog:
### Database name
PostgreSQL 12

# Issue description

### Describe your problem
`wal-g delete retain N` fails if any backups are found from WAL-E

### Please provide steps to reproduce
```
wal-e backup-push /var/lib/postgresql/12/main
wal-e backup-push /var/lib/postgresql/12/main
wal-e backup-push /var/lib/postgresql/12/main
wal-e backup-push /var/lib/postgresql/12/main
wal-g delete retain 3
```

### Please add config and wal-g stdout/stderr logs for debug purpose

```json
{
  "PGHOST": "/var/run/postgresql",
  "PGUSER": "postgres",
  "WALG_COMPRESSION_METHOD": "brotli",
  "WALG_DOWNLOAD_CONCURRENCY": 2,
  "WALG_FILE_PREFIX": "/mnt/pg_backup",
  "WALG_PREFETCH_DIR": "/var/lib/postgresql/prefetch",
  "WALG_PREVENT_WAL_OVERWRITE": true,
  "WALG_UPLOAD_CONCURRENCY": 1
}
```

<details><summary>If you can, provide logs</summary>

```
INFO: 2022/06/10 16:31:16.361203 retrieving permanent objects
ERROR: 2022/06/10 16:31:16.374858 failed to fetch backup meta for backup base_000000020000016B0000005E_00000040 with error failed to unmarshal metadata: object '/mnt/pg_backup/basebackups_005/base_000000020000016B0000005E_00000040/metadata.json' not found in storage
github.com/wal-g/wal-g/pkg/storages/storage.NewObjectNotFoundError
        /home/runner/work/wal-g/wal-g/pkg/storages/storage/errors.go:15
github.com/wal-g/wal-g/pkg/storages/fs.(*Folder).ReadObject
        /home/runner/work/wal-g/wal-g/pkg/storages/fs/folder.go:100
github.com/wal-g/wal-g/internal.(*StorageReaderMaker).Reader
        /home/runner/work/wal-g/wal-g/internal/storage_reader_maker.go:31
github.com/wal-g/wal-g/internal.FetchDto
        /home/runner/work/wal-g/wal-g/internal/backup.go:86
github.com/wal-g/wal-g/internal.(*Backup).FetchMetadata
        /home/runner/work/wal-g/wal-g/internal/backup.go:72
[...]

PR #1292, opened 2022-06-11
Fix of deleting WAL-E backups

x4m:
Fix for #1291

### DIFF
--- a/internal/delete_util.go
+++ b/internal/delete_util.go
@@ -22,8 +22,7 @@ func FindPermanentBackups(folder storage.Folder, metaFetcher GenericMetaFetcher)
 		meta, err := metaFetcher.Fetch(
 			backupTime.BackupName, folder.GetSubFolder(utility.BaseBackupPath))
 		if err != nil {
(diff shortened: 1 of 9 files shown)


usernamedt (approved):

internal/delete_util.go
```diff
@@ -10,27 +10,14 @@ import (
...
-			backupTime.BackupName, folder.GetSubFolder(utility.BaseBackupPath))
-		if err != nil {
-			tracelog.ErrorLogger.Printf("failed to fetch backup meta for backup %s with error %s, ignoring...",
-				backupTime.BackupName, err.Error())
-			continue
-		}
-		if meta.IsPermanent {
-			permanentBackups[backupTime.BackupName] = true
-		}
+func PrintMetadataNotFoundError(backupTime BackupTime, err error) {
```
usernamedt, line 13: well, now this function not only prints but exits in case of non-404s so I guess it needs to be renamed

usernamedt (changes requested):

teem0n (approved):

usernamedt (approved):

PR #1297, opened 2022-06-18
[WIP] Multithread downloading directory

hariton27sy:
### Database name
Databases with physical backups

# Pull request description
Second part of working with directories. Downloader


internal/bundle.go
```diff
@@ -110,5 +108,23 @@ func (bundle *Bundle) FinishComposing() (TarFileSets, error) {
 }
 
 func (bundle *Bundle) getFileRelPath(fileAbsPath string) string {
-	return utility.PathSeparator + utility.GetSubdirectoryRelativePath(fileAbsPath, bundle.Directory)
+	return utility.GetSubdirectoryRelativePath(fileAbsPath, bundle.Directory)
+}
+
+func (bundle *Bundle) createTarFileInfoHeader(path string, info os.FileInfo) (header *tar.Header, err error) {
+	var symLinkTarget string
+	if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
usernamedt, line 116: WAL-G default behavior is not to follow symlinks.

internal/bundle.go
```diff
@@ -81,16 +81,14 @@ func (bundle *Bundle) AddToBundle(path string, info os.FileInfo, err error) erro
...
 	if err != nil {
-		return errors.Wrap(err, "addToBundle: could not grab header info")
+		return err
 	}
 
-	fileInfoHeader.Name = bundle.getFileRelPath(path)
 	tracelog.DebugLogger.Println(fileInfoHeader.Name)
 
-	if bundle.FilesFilter.ShouldUploadFile(path) {
+	if bundle.FilesFilter.ShouldUploadFile(path) && info.Mode().IsRegular() {
```
usernamedt, line 91: WAL-G uploads all files, including dirs and symlinks.

usernamedt (changes requested):

internal/bundle.go
```diff
@@ -110,5 +108,15 @@ func (bundle *Bundle) FinishComposing() (TarFileSets, error) {
 }
 
 func (bundle *Bundle) getFileRelPath(fileAbsPath string) string {
-	return utility.PathSeparator + utility.GetSubdirectoryRelativePath(fileAbsPath, bundle.Directory)
+	return utility.GetSubdirectoryRelativePath(fileAbsPath, bundle.Directory)
```
usernamedt, line 111: Why to remove the leading path separator?

internal/directory_downloader.go
```diff
@@ -0,0 +1,83 @@
...
+type DirectoryIsNotEmptyError struct {
+	error
+}
+
+type CommonDirectoryDownloader struct {
+	Folder     storage.Folder
+	BackupName string
+}
+
+func NewCommonDirectoryDownloader(folder storage.Folder, backupName string) DirectoryDownloader {
```
usernamedt, line 24: I guess it is better to pass the explicit list of tars for download to the downloader constructor. Sometimes we don't need to download all the tars from the folder.

hariton27sy: If you have explicit list of tars then you can use `ExtractAll` function from package `internal`

internal/bundle.go
hariton27sy: Make behaviour like in postgres relative path

usernamedt (approved):
LGTM, the only thing left is to fix the failing tests

PR #1302, opened 2022-06-24
Add ListFolderRecursivelyWithFilter to storages & optimize delete handler in Greenplum

usernamedt:
Add the `ListFolderRecursivelyWithFilter` function that accepts folder filters so we can avoid the excessive storage listing calls (for example, ListObjects in S3). This allows for optimizing some functionality, such as delete handlers.


usernamedt:
In this PR, I did only basic optimization for the Greenplum delete handler.

x4m (approved):

PR #1313, opened 2022-07-16
MariaDB. Fix backup-push

ostinru:
### MaridDB
This patch fixes a regression in `backup-push` handler - now it works with MariaDB as well.

In addition fix MariaDB tests and adds MariaDB tests to the default github actions workflow, so we will not break it in future.

Steps to reproduce:
* Setup MariaDB
* run `wal-g backup-push`
Following error raised:
```
ERROR: 2022/07/16 15:42:27.699160 failed to get last uploaded binlog: failed to parse server gtid: 0-1-4112: invalid GTID format, must UUID:interval[:interval]
```

### DIFF
--- a/.github/workflows/dockertests.yml
+++ b/.github/workflows/dockertests.yml
@@ -125,6 +125,7 @@ jobs:
                    'make MYSQL_TEST=mysql_base_tests mysql_integration_test',
                    'make MYSQL_TEST=mysql_delete_tests mysql_integration_test',
                    'make MYSQL_TEST=mysql_copy_tests mysql_integration_test',
+                   'make mariadb_test',
                    'make gp_test',
                    'make st_test',
                    'make TEST="pg_pgbackrest_backup_fetch_test" pg_integration_test'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -65,8 +65,11 @@ services:
       &&  mkdir -p /export/mysqlpitrmysqldumpbucket
       &&  mkdir -p /export/mysqllivereplay
       &&  mkdir -p /export/mysqlpermanentbackupbucket
-      &&  mkdir -p /export/mariadbfullmysqldumpbucket
-      &&  mkdir -p /export/mariadbfullmariabackupbucket
+      &&  mkdir -p /export/mariadb_binlog_push_fetch
+      &&  mkdir -p /export/mariadb_binlog_push_with_gtids_check
+      &&  mkdir -p /export/mariadb_full_mysqldump
+      &&  mkdir -p /export/mariadb_full_maria_backup
+      &&  mkdir -p /export/mariadb_pitr_xtrabackup
       &&  mkdir -p /export/mongostreampushbucket
       &&  mkdir -p /export/mongooplogpushbucket
       &&  mkdir -p /export/mongodeletebeforebucket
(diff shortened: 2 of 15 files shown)


mialinx:
Let's run mariadb tests in `.github/workflows/dockertests.yml`  ?

mialinx:
It looks like the only place we rely on MySQL-specific gtid format is `gtidFilter.shouldUpload` (we use Minus/Add mehtods).
In onther places we use `GTIDSet.Contain` interface method.

Can we replace `*mysql.MysqlGTIDSet` with `GTIDSet` in most places and get rid of type-ifs ?

mialinx (approved):

ostinru:
It seems PiTR had never worked for MariaDB: I haven't found a way to skip transactions already applied when `mysqlbinlog | mysql` approach is used. There are 3 approaches suggested in MariaDB documentation: 
* use `mysqlbinlog` with binlog+position - this can work for simple cases and can cause troubles when we have binlogs from different masters.
* use `mysqlbinlog` with GTIDs (for MariaDB 10.8+) - it can find position in binlog by looking at GTID you provide (so, basically it is the previous approach)
* use master as a source of binlogs - it seems that MariaDB can handle duplicates on its own (at least MariaDB support ring-topology replication)

I think that for now, we should:
1. mark PiTR for MariaDB as not supported
2. suggest workarounds in MySQL.md
3. Implement `binlog-server` feature, so wal-g would be able to work as a replication source. This can be beneficial for MySQL as well.

teem0n (approved):

mialinx:
> It seems PiTR had never worked for MariaDB: I haven't found a way to skip transactions already applied when `mysqlbinlog | mysql` approach is used. There are 3 approaches suggested in MariaDB documentation:

> * use `mysqlbinlog` with GTIDs (for MariaDB 10.8+) - it can find position in binlog by looking at GTID you provide (so, basically it is the previous approach)

Doesn't MariaDB skip transactions with already executed GTID  identifiers ?

ostinru:
### Applying binlogs
It is interesting that `mysqlbinlog` knows about GTID:
```
mysqlbinlog /tmp/mysql-bin.000003253796933
<...>
#220725  9:13:49 server id 1  end_log_pos 427 CRC32 0xf80c51c6  GTID 0-1-5810 ddl
/*!100101 SET @@session.skip_parallel_replication=0*//*!*/;
/*!100001 SET @@session.gtid_domain_id=0*//*!*/;
/*!100001 SET @@session.server_id=1*//*!*/;
/*!100001 SET @@session.gtid_seq_no=5810*//*!*/;
<...>
```

So, I would expect that MariaDB can skip duplicates... There is `gtid_ignore_duplicates`[2] option that is OFF by defalt. Unfortunately it is not fixing an issue for me (probably I am incorrectly set `gtid_slave_pos` or something like this)


### What people from MariaDB community suggests
I have asked people on MariaDB zullip chat - how they working with PiTR and they suggest not to use `mysqlbinlog | mysql` - instead use `CHANGE MASTER TO RELAY_LOG_FILE='mysql1-relay-bin.000001', RELAY_LOG_POS=1, MASTER_HOST='dummy';`[3] as fast(parallel) and reliable way to apply binlogs. Unfortunately, it seems don't have a way to stop at some point in time (only at position / gtid)

[1] https://mariadb.com/kb/en/gtid/#gtid_strict_mode
[2] https://mariadb.com/docs/reference/mdb/system-variables/gtid_ignore_duplicates/
[3] https://lefred.be/content/howto-make-mysql-point-in-time-recovery-faster/

Issue #300, opened 2019-06-05
wal-g releases are incompatible with CentOS 6


jstaf:
Because wal-g uses C/C++ extensions (namely the brotli library), it is dependent on the version of glibc where the binary is built. The current build environment makes wal-g incompatible with older OS versions running on many database servers, like CentOS 6 (this is probably true for older Ubuntu and Debian versions, but I haven't tested those).

```
$ ./wal-g 
./wal-g: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by ./wal-g)
$ ldd wal-g
./wal-g: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by ./wal-g)
	linux-vdso.so.1 =>  (0x00007ffd71f10000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f8ba911c000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f8ba8f18000)
	liblzo2.so.2 => /usr/lib64/liblzo2.so.2 (0x00007f8ba8cf7000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f8ba8ada000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f8ba8746000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f8ba93a0000)
```

To resolve this, you should either provide a release where CGo is disabled/brotli compression is totally absent from the binary, or build on an older OS to avoid the glibc versioning issue. For what it's worth, wal-g works on CentOS 6 when it's built on CentOS 6 (I tried this out just now and was able to get a working binary from the latest master).

x4m:
Thanks for reporting this! I will think about how to make releases better...
We can make brotli optional like lzo is now... but then we will have to ship lzo,br,lzo+br, that's not good...
Or we can make releases on CentOS too...

jstaf:
I think the simplest resolution would be to just build releases on something with an older version of glibc. Any older OS should work for this, CentOS 6 is just one example. Let me see if I can dig up what docker image I used to do the build when I get home (created the Dockerfile as a one-off to test things out). 

Regarding lzo+brotli, after second thought I'd almost just include both in every release once the glibc versioning issue gets ironed out. From my understanding of things, brotli gets built into the binary anyways, and lzo is in the standard repositories for most servers, so having lzo/brotli by default isn't a big deal. (Anyone who's migrating to wal-g from wal-e probably already has lzo installed.)

[...]

PR #1315, opened 2022-07-21
Statically compile wal-g using docker

jstaf:
### Database name

All databases! 🙂 

# Pull request description

This PR allows compiling a statically-linked wal-g binary, compresses the binary using `upx`, and changes the `wal-g/golang` Dockerfile to Alpine Linux. All binaries are now smaller (10-12MB instead of >40MB). The static binary has the important advantage over the normal binary in that it can be used on any Linux system without recompilation.

The static build is not the new default (you can still compile dynamically-linked binaries as before). To compile a statically linked wal-g, set `export USE_STATIC_BUILD=1` before compiling:

```bash
# compile a statically linked binary for MySQL
export USE_LIBSODIUM=1
export USE_STATIC_BUILD=1
docker build -t wal-g/golang -f docker/golang/Dockerfile .
make deps
make mysql_build
```

### Describe what this PR fix

Currently wal-g binaries are not portable between different Linux distributions or versions because it relies on linking against libsodium and brotli. By statically linking instead, the resulting binary can be used on any Linux system. Fixes https://github.com/wal-g/wal-g/issues/300.

### Known issues

WAL-G linked against musl's libc will fail to read users managed by services like FreeIPA/sssd/GCP OSLogin. This is because musl doesn't properly support NSS. Running wal-g with a user created by sssd/OSLogin/etc. will result in the error below:

```
$ wal-g backup-list
ERROR: 2022/07/14 17:16:17.325247 user: unknown userid 1092191234
```

However I ultimately feel this error is a non-issue: you're usually going to be using sudo to execute wal-g as a local UNIX user anyways (and these work just fine). 

Credit for this PR mostly goes to @halsbox who came up with the idea to do this here: https://github.com/wal-g/wal-g/issues/300#issuecomment-1166289794

I've only really tested this for the MySQL build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,12 @@ ifdef USE_LZO
 	BUILD_TAGS:=$(BUILD_TAGS) lzo
 endif
 
+ifdef USE_STATIC_BUILD
+	# build using alpine golang image with musl to allow static linking
+	BUILD_CMD := docker run --rm -it -e GOCACHE=/tmp/gocache -v $(PWD)/tmp/gocache:/tmp/gocache -v $(PWD):/app wal-g/golang
+	BUILD_LDFLAGS := -linkmode external -extldflags \"-static\"
+endif
+
 .PHONY: unittest fmt lint clean install_tools
 
 test: deps unittest pg_build mysql_build redis_build mongo_build gp_build unlink_brotli pg_integration_test mysql_integration_test redis_integration_test fdb_integration_test gp_integration_test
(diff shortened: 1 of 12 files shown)


Doyle-net:
Compiled on alpine OS，for PostgreSQL
`
# 编译wal-g
FROM golang:alpine3.16 AS builder
ENV WALG_VERSION=v2.0.0
ENV _build_deps="wget cmake git build-base bash"
RUN set -ex  \
     && apk add --no-cache $_build_deps \
     && git clone https://github.com/wal-g/wal-g/  $GOPATH/src/wal-g \
     && cd $GOPATH/src/wal-g/ \
     && git checkout $WALG_VERSION \
     && make install_and_build_pg \
     && install main/pg/wal-g / \
     && /wal-g --help

FROM alpine:latest
COPY --from=builder /wal-g /usr/local/bin/
`

x4m:
This looks really awesome!

@usernamedt can you please look into this after vacation?

usernamedt:
> This looks really awesome!
> 
> 
> 
> @usernamedt can you please look into this after vacation?

Sure, I'll have a look in a couple of days

usernamedt:
@jstaf can you have a look why docker tests fail to complete?

Issue #585, opened 2020-03-10
Failed to restore PG cluster from differential backup with several table spaces


banlex73:
wal-g v.2.14
CentOS 7
------------------
Backup extraction complete.
INFO: 2020/03/06 00:02:10.176626 base_00000001000000000000000E fetched. Upgrading from LSN e000028 to LSN 10000028
ERROR: 2020/03/06 00:02:10.184098 Failed to fetch backup: Error creating tablespace symkink symlink /pgsql/index/index_wal_g /pgsql/cluster/data_walg/pg_tblspc/16386: file exists
------------------
Taking a full backup
./wal-g backup-push /pgsql/cluster/data_walg/
INFO: 2020/03/06 00:05:20.161925 Doing full backup.
INFO: 2020/03/06 00:05:20.172496 Calling pg_start_backup()
INFO: 2020/03/06 00:05:20.268884 Walking ...
INFO: 2020/03/06 00:05:20.269233 Starting part 1 ...
INFO: 2020/03/06 00:05:20.567598 Finished writing part 1.
INFO: 2020/03/06 00:05:20.769631 Starting part 2 ...
INFO: 2020/03/06 00:05:20.769833 /global/pg_control
INFO: 2020/03/06 00:05:20.770321 Finished writing part 2.
INFO: 2020/03/06 00:05:20.771293 Calling pg_stop_backup()
INFO: 2020/03/06 00:05:21.822094 Starting part 3 ...
INFO: 2020/03/06 00:05:21.822512 backup_label
INFO: 2020/03/06 00:05:21.822534 tablespace_map
INFO: 2020/03/06 00:05:21.822991 Finished writing part 3.
INFO: 2020/03/06 00:05:22.108895 Wrote backup with name base_000000010000000000000014
-------------------
Success with full backup
./wal-g backup-fetch  /pgsql/cluster/data_walg/ LATEST
INFO: 2020/03/06 00:05:45.970066 LATEST backup is: 'base_000000010000000000000014'
INFO: 2020/03/06 00:05:47.303329 Finished decompression of part_003.tar.br
INFO: 2020/03/06 00:05:47.303347 Finished extraction of part_003.tar.br
INFO: 2020/03/06 00:05:49.937295 Finished decompression of part_001.tar.br
INFO: 2020/03/06 00:05:49.937315 Finished extraction of part_001.tar.br
INFO: 2020/03/06 00:05:50.287711 Finished decompression of pg_control.tar.br
[...]

Issue #499, opened 2019-12-11
backup-fetch: Error creating tablespace symkink


tatodorov:
I have a database with few tablespaces, each in a separate location.
I am able to restore successfully from a full backup.
Unfortunately, when I try to restore from a delta backup I am encountering the following error:

-bash-4.2$ wal-g backup-fetch /var/lib/pgsql/data base_000000010000000000000058_D_000000010000000000000056
INFO: 2019/12/11 11:01:31.960831 Delta from base_000000010000000000000056 at LSN 56000028
INFO: 2019/12/11 11:01:32.117140 Finished extraction of part_003.tar.br
INFO: 2019/12/11 11:01:32.117413 Finished decompression of part_003.tar.br
INFO: 2019/12/11 11:02:39.146508 Finished extraction of part_001.tar.br
INFO: 2019/12/11 11:02:39.146802 Finished decompression of part_001.tar.br
INFO: 2019/12/11 11:02:39.217374 Finished extraction of pg_control.tar.br
INFO: 2019/12/11 11:02:39.217392
Backup extraction complete.
INFO: 2019/12/11 11:02:39.217400 base_000000010000000000000056 fetched. Upgrading from LSN 56000028 to LSN 58000028
INFO: 2019/12/11 11:02:39.217424 Finished decompression of pg_control.tar.br
ERROR: 2019/12/11 11:02:39.267744 Failed to fetch backup: Error creating tablespace symkink symlink /var/lib/pgsql/testdb/tablespace/testdb_data /var/lib/pgsql/data/pg_tblspc/16386: file exists

wal-g first extracts the full backup and later the delta backups.
The symbolic link is created when the full backup is extracted, and when wal-g tries to extract the delta, it complains that the symbolic link to the tablespace location already exists.
So, it fails.

g0djan:
Hi, we'll take a look

kos12345:
it doesn't work. same problem.
wal-g backup-list
name                                                     last_modified             wal_segment_backup_start
base_00000037000000010000000A                            2021-04-09T15:16:40+03:00 00000037000000010000000A
base_000000370000000100000060_D_00000037000000010000000A 2021-04-09T15:30:53+03:00 000000370000000100000060
base_00000037000000010000007E                            2021-04-09T16:13:35+03:00 00000037000000010000007E
base_0000003700000001000000E7                            2021-04-12T10:50:07+03:00 0000003700000001000000E7
[...]

PR #1320, opened 2022-08-05
Change error to warning when creating symlink, that already exists

IncubusRK:
### Database name
PostgreSQL

# Pull request description
Fix error when fetching incremental backup of posgresql database, that have multiple table spaces.

### Describe what this PR fix
wal-g trying to create symlink for additional tablespace on every part of backup and exit witch error

Also see
https://github.com/wal-g/wal-g/issues/585
https://github.com/wal-g/wal-g/issues/499

create new postgres database
create additional table space
push-backup
fetch-backup


internal/databases/postgres/backup.go
```diff
@@ -235,7 +235,14 @@ func setTablespacePaths(spec TablespaceSpec) error {
 	for _, location := range spec.tablespaceLocations() {
 		err := fs.NewFolder(location.Location, "").EnsureExists()
 		if err != nil {
-			return fmt.Errorf("error creating folder for tablespace %v", err)
+			// Don't show error when correct symlink already exist
```
x4m, line 238: it would be very good to do this check within EnsureExists() function.

x4m (commented):
Many thanks for fixing this bug! Your fix seems correct. But it would be more robust to refactor this stuff a little. What if we do this check in the EnsureExists() function? just check there that err is "Dir already exists"?

Thank you!

internal/databases/postgres/backup.go
IncubusRK: EnsureExists() used for Folder type. 
I think, I should not create a new type to work with symbolic links if only a few calls are needed.

IncubusRK:
This fix only help to prevent canceling of backup-fetch process. 
May be also need to change the recovery logic.

For example after run next command in /data (original base also backuped from /data) dir:
wal-g backup-fetch ./backup-test LATEST
I got /data/backup-test with recovered database and /data/pg_tblspc with empty directories and sysmlinks to it.
It meants that recovery process use relative path to recovery additional tablespaces, which may not be convenient in terms of free space for a very large database.

usernamedt:
> This fix only help to prevent canceling of backup-fetch process. May be also need to change the recovery logic.
> 
> For example after run next command in /data (original base also backuped from /data) dir: wal-g backup-fetch ./backup-test LATEST I got /data/backup-test with recovered database and /data/pg_tblspc with empty directories and sysmlinks to it. It meants that recovery process use relative path to recovery additional tablespaces, which may not be convenient in terms of free space for a very large database.

I guess it would be cool to add the check for the correct handling of delta backups to this [integration test](https://github.com/wal-g/wal-g/blob/master/docker/pg_tests/scripts/tests/tablespace_support_test.sh) to make sure it won't accidentally break in the future.

internal/databases/postgres/backup.go
usernamedt: You can add this logic to the [TablespaceLocation](https://github.com/wal-g/wal-g/blob/7f699a029a4d625c18136ebf9e9d87b022c27022/internal/databases/postgres/tablespace_spec.go#L28) type

IncubusRK:
Added some minor fixes for passing linter checks

serprex (approved):

PR #1326, opened 2022-08-24
expose zstd compression

heinerstilz:
### Database name
PostgreSQL

# Pull request description

### Describe what this PR fix
The zstd compression library is already supported internally in WAL-G. It seems to have been disabled because of a data corruption bug in zstd that has meanwhile been resolved (https://github.com/DataDog/zstd/issues/39).
It has been mentioned that it might be good to bring zstd back (see discussion on https://github.com/wal-g/wal-g/issues/300).

This PR
- exposes the zstd compression option to the user
- upgrades to a more recent version
- adjusts some docs.

In our measurements with PostgreSQL and WAL-G on a ~270 GB dataset, zstd was significantly less CPU intensive than brotli:
1.6x less real time and 1.25x less user time backing up/compressing with WALG_UPLOAD_CONCURRENCY=8
1.6x less real time and 2.8x less user time restoring/decompressing.
Compression ratio was around 2.4x for both.

Another measurement by a different party that seems to support zstd's performance advantage over brotli: https://peazip.github.io/fast-compression-benchmark-brotli-zstandard.html#:~:text=Comparing%20Brotli%20and%20Zstandard%20extraction,twice%20as%20fast%20as%20Brotli.

### Please provide steps to test this PR
Which of the tests makes most sense to be adapted for zstd?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,7 +8,7 @@
 
 WAL-G is an archival restoration tool for PostgreSQL, MySQL/MariaDB, and MS SQL Server (beta for MongoDB and Redis).
 
-WAL-G is the successor of WAL-E with a number of key differences. WAL-G uses LZ4, LZMA, or Brotli compression, multiple processors, and non-exclusive base backups for Postgres. More information on the original design and implementation of WAL-G can be found on the Citus Data blog post ["Introducing WAL-G by Citus: Faster Disaster Recovery for Postgres"](https://www.citusdata.com/blog/2017/08/18/introducing-wal-g-faster-restores-for-postgres/).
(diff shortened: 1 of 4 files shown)


x4m:
FWIW I saw one report of brotli corruption: a file properly decrypted by gpg, but unexctractable. Probably, due to cosmic rays or something.
Zstd is great codec, best on Pareto frontier. It would be very good to expose it. But...

@heinerstilz do you validate you backups? Will you use Zstd? We need someone who will warn us if something is still wrong with Zstd.

heinerstilz:
Implementing our backup solution for PostgreSQL, the idea is to go with zstd unless we find an issue with it.
We'll likely soon have a large volume of frequent Postgres backups in production. Usable restores are needed on a regular basis.
So should zstd still corrupt data, there is quite a good chance we'll notice (and of course flag it here).

PR #1353, opened 2022-10-07
Add daemon mode for wal-push command

rogaliiik:
### Database name
postgresql15

# Pull request description

Added daemon command for archiving wal segments


internal/databases/postgres/wal_push_handler.go
```diff
@@ -32,21 +31,20 @@ func (err CantOverwriteWalFileError) Error() string {
 
 // TODO : unit tests
 // HandleWALPush is invoked to perform wal-g wal-push
-func HandleWALPush(uploader *WalUploader, walFilePath string) {
```
usernamedt, line 35: You should also modify the original ```wal-push``` command handler (cmd/pg/wal_push.go):
1. Check for the returned error
2. Since the uploading folder initialization is removed from ```HandleWALPush```, add it before the function call.

cmd/pg/daemon.go
```diff
@@ -0,0 +1,118 @@
...
+		tracelog.ErrorLogger.Println("OK write fail: ", err)
+	}
+}
+
+// daemonCmd represents the daemon command
+var daemonCmd = &cobra.Command{
+	Use:   "daemon wal_filepath",
+	Short: DaemonShortDescription, // TODO : improve description
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
```
usernamedt, line 74: There is too much logic for a command package.

1. Create a handler in the `databases/postgres` package. 
2. Move this logic to the handler
3. Call this handler from here


You can see how other commands are implemented for reference.

usernamedt (changes requested):

cmd/pg/daemon.go
```diff
@@ -0,0 +1,118 @@
...
+			if err != nil {
+				tracelog.ErrorLogger.Println("accept error:", err)
+			}
+
+			go handleConn(fd, func(walFileName string) error {
+				fullPath := path.Join(args[0], walFileName)
+				tracelog.InfoLogger.Printf("starting wal-push for %s\n", fullPath)
+				return postgres.HandleWALPush(uploader, fullPath)
+			})
+		}
```
ostinru, line 112: I have no idea of the purpose of this daemon =) , however, I have couple of thoughts:
* Would it be beneficial to have an option to gracefully exit from this loop?
* Is it possible that `Accept` will produce upload tasks faster than we can handle? What is expected behavior in this case?

rogaliiik: Interesting thoughts. Appreciate it.

cmd/pg/daemon.go
```diff
@@ -0,0 +1,70 @@
...
+			uploader.PGArchiveStatusManager = asm.NewDataFolderASM(PGArchiveStatusManager)
+		} else {
+			tracelog.ErrorLogger.PrintError(err)
+			uploader.PGArchiveStatusManager = asm.NewNopASM()
+		}
+
+		uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(utility.WalPath)
+
+		_ = os.Remove(DaemonSocketName)
+		l, err := net.Listen("unix", DaemonSocketName)
```
usernamedt, line 48: All the internal socket-related configuration should be inside the command handler.

cmd/pg/daemon.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/wal-g/wal-g/internal/asm"
+	"github.com/wal-g/wal-g/internal/databases/postgres"
+)
+
+const DaemonShortDescription = "Uploads a WAL file to storage"
+const DaemonSocketName = "/tmp/wal-push.sock"
+
+// daemonCmd represents the daemon command
+var daemonCmd = &cobra.Command{
+	Use:   "daemon wal_filepath",
```
usernamedt, line 22: What is the logic behind the `wal_filepath` argument? I guess that the WAL-G daemon command should accept the path to the socket as an argument.

docker/pg_tests/scripts/tests/daemon_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+set -e -x
```
usernamedt, line 2: You need to make sure that this test is included in [the CI pipeline](https://github.com/wal-g/wal-g/blob/e7ffb57e4c5ba385ed596e45295ad8b60eb1ae76/.github/workflows/dockertests.yml#L103). It means that you also need to add the corresponding configuration files. Check how others Postgresql Docker tests are implemented for reference.

usernamedt (changes requested):

docker/pg_tests/scripts/tests/daemon_test.sh
```diff
@@ -0,0 +1,28 @@
...
+cat ${COMMON_CONFIG} >> ${TMP_CONFIG}
+/tmp/scripts/wrap_config_file.sh ${TMP_CONFIG}
+
+WAL=$(ls -l ${PGDATA}/pg_wal | head -n2 | tail -n1 | egrep -o "[0-9A-F]{24}")
+
+wal-g --config=${TMP_CONFIG} daemon "${PGDATA}"/pg_wal
+
+# shellcheck disable=SC2039
+nc -U /tmp/wal-push.sock <<< 'CHECK'
+sleep 1
```
usernamedt, line 20: It is better to check for reply from the socket and remove the `sleep 1` calls. It would help to:
- make sure that WAL-G responded correctly
- avoid race conditions when using the `sleep 1` logic

internal/databases/postgres/daemon_handler.go
```diff
@@ -3,13 +3,64 @@ package postgres
...
+	"errors"
+	"fmt"
 	"net"
 	"os"
 	"path"
 
 	"github.com/wal-g/tracelog"
 )
 
+type CheckType byte
```
usernamedt, line 15:
```go
type SocketMessageType byte

const (
	ArchiveMessage SocketMessageType = 'a'
	CheckMessage  SocketMessageType = 'c'
)
```

internal/databases/postgres/daemon_handler.go
```diff
@@ -3,13 +3,64 @@ package postgres
...
+	case 'C':
+		return CheckType('C')
+	case 'F':
+		return FileType('F')
+	default:
+		return nil
+	}
+}
+
+func (msg CheckType) Handle(messageBody []byte, c net.Conn, f func(string) error) error {
```
usernamedt, line 33: You need to create a separate message handler structure for each message type.

i.e.
```go
type SocketMessageHandler interface {
Handle(messageBody []byte) error
}


type ArchiveMessageHandler struct
{
...
}

func (h * ArchiveMessageHandler) Handle(messageBody []byte) error {}

type CheckMessageHandler struct
{
...
}

func (h * CheckMessageHandler) Handle(messageBody []byte) error {}
```

usernamedt (changes requested):

cmd/pg/daemon.go
```diff
@@ -0,0 +1,55 @@
...
+		PGArchiveStatusManager, err := internal.ConfigurePGArchiveStatusManager()
+		if err == nil {
+			uploader.PGArchiveStatusManager = asm.NewDataFolderASM(PGArchiveStatusManager)
+		} else {
+			tracelog.ErrorLogger.PrintError(err)
+			uploader.PGArchiveStatusManager = asm.NewNopASM()
+		}
+
+		uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(utility.WalPath)
+		pathToData, exist := os.LookupEnv("PGDATA")
```
usernamedt, line 43: We use viper for configuration:

https://github.com/wal-g/wal-g/blob/71fb7e9d110a80d971744963ac5c2cabfd041cc3/tests_func/helpers/walg.go#L124

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,132 @@
...
+		tracelog.ErrorLogger.Printf("Error on writing in socket: %v", err)
+		return err
+	}
+	return nil
+}
+
+func (msg FileType) Handle(messageBody []byte, c net.Conn, f func(string) error) error {
+	if len(messageBody) < 24 {
+		if len(messageBody) > 0 {
+			tracelog.ErrorLogger.Printf("Received incorrect message %s", messageBody)
```
usernamedt, line 45: Create a separate validation function and return an error if validation failed

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,132 @@
...
+func (msg CheckType) Handle(messageBody []byte, c net.Conn, f func(string) error) error {
+	_, err := c.Write([]byte{'O', 0, 3})
+	if err != nil {
+		tracelog.ErrorLogger.Printf("Error on writing in socket: %v", err)
+		return err
+	}
+	return nil
+}
+
+func (msg FileType) Handle(messageBody []byte, c net.Conn, f func(string) error) error {
```
usernamedt, line 42: Why do you need to inject this handler function so far?

Just call the archiver command directly from here:
```go
fullPath := path.Join(pathToWal, walFileName)
tracelog.InfoLogger.Printf("starting wal-push for %s\n", fullPath)
return HandleWALPush(uploader, fullPath)
```

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,132 @@
...
+			fullPath := path.Join(pathToWal, walFileName)
+			tracelog.InfoLogger.Printf("starting wal-push for %s\n", fullPath)
+			return HandleWALPush(uploader, fullPath)
+		})
+	}
+}
+
+func DaemonProcess(c net.Conn, f func(string) error) {
+	defer func(c net.Conn) {
+		err := c.Close()
```
usernamedt, line 86: utility.LoggedClose()

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,132 @@
...
+		err = requestType.Handle(byteBody, c, f)
+		if err != nil {
+			tracelog.ErrorLogger.Println("Failed to handle message, err:", err)
+		}
+		if byteType == 'F' {
+			return
+		}
+		if int(byteLength) < nr {
+			tracelog.InfoLogger.Println("Read remaining buffer...")
+			byteType, byteLength, byteBody = MessageParser(buf[byteLength:])
```
usernamedt, line 111: Why do we need to duplicate the code from above?

usernamedt (changes requested):

internal/databases/postgres/daemon_handler.go
rogaliiik: Duplication for the case when two messages are written to the buffer at the same time and we don't want to block on repeated reading from the buffer, which will not happen. should I implement a function to reuse?

usernamedt: You can create a special protocol reader that implements an `io.Reader` interface and  encapsulate this logic in it.

docker/pg_tests/scripts/tests/daemon_test.sh
```diff
@@ -0,0 +1,46 @@
...
+  echo -en "C\x0\x8"
+  echo -n "CHECK"
+  echo -en "F\x0\x1B"
+  echo -n "${WAL}"
+} | nc -U ${SOCKET} | grep -q "OO"; then
+  echo "WAL-G response is correct"
+else
+  echo "Error in WAL-G response"
+  exit 1
+fi
```
usernamedt, line 46: We also need to check that WAL has been archived correctly.

cmd/pg/wal_push.go
```diff
@@ -35,7 +36,11 @@ var walPushCmd = &cobra.Command{
 			uploader.PGArchiveStatusManager = asm.NewNopASM()
 		}
 
-		postgres.HandleWALPush(uploader, args[0])
+		uploader.UploadingFolder = uploader.UploadingFolder.GetSubFolder(utility.WalPath)
+		err = postgres.HandleWALPush(uploader, args[0])
+		if err != nil {
```
usernamedt, line 41: You can remove this if and leave only the `			tracelog.ErrorLogger.FatalOnError(err)` part.

cmd/pg/daemon.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/wal-g/wal-g/internal/asm"
+	"github.com/wal-g/wal-g/internal/databases/postgres"
+	"github.com/wal-g/wal-g/utility"
+)
+
+const DaemonShortDescription = "Uploads a WAL file to storage"
+
+// daemonCmd represents the daemon archive command
+var daemonCmd = &cobra.Command{
+	Use:   "daemon daemon_socket_path",
```
usernamedt, line 16: Add some information  about this command to the [documentation](https://github.com/wal-g/wal-g/blob/master/docs/PostgreSQL.md).

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	Handle(messageBody []byte) error
+}
+
+type ArchiveMessageHandler struct {
+	messageType SocketMessageType
+	fd          net.Conn
+	uploader    *WalUploader
+}
+
+func (h *CheckMessageHandler) Handle(messageBody []byte) error {
```
usernamedt, line 34: CheckMessageHandler/ArchiveMessageHandler struct declarations and Handle methods are mixed up, you can reorder them

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+type CheckMessageHandler struct {
+	messageType SocketMessageType
+	fd          net.Conn
+	uploader    *WalUploader
+}
+
+func (h *ArchiveMessageHandler) Handle(messageBody []byte) error {
+	err := messageValidation(messageBody)
+	if err != nil {
+		tracelog.ErrorLogger.Printf("Incorrect message: %v\n", err)
```
usernamedt, line 53: You can just log the error in the upper level.
Here you can use the fmt.Errorf("msg: %w", err) to wrap error with some message.

For example:
```go
if err != nil {
    return fmt.Errorf("message validation failed: %w", err)
}
```

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	_, err = h.fd.Write([]byte{'O'})
+	if err != nil {
+		tracelog.ErrorLogger.Printf("failed to write in socket: %v\n", err)
+		return err
+	}
+	return nil
+}
+
+func messageValidation(messageBody []byte) error {
```
usernamedt, line 78: It is preffered to use verbs in function names, e.g. `validateMessage()`

internal/databases/postgres/daemon_handler.go
usernamedt: Same issue in other places below.

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	default:
+		return nil
+	}
+}
+
+type SocketMessageReader struct {
+	c net.Conn
+}
+
+func GetMessageReader(c net.Conn) *SocketMessageReader {
```
usernamedt, line 104: It is preferred to use `NewSmth` for constructors, e.g. `NewMessageReader`

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+		if len(messageBody) > 0 {
+			tracelog.ErrorLogger.Println("incorrect message accepted")
+			return fmt.Errorf("incorrect message accepted: %s", string(messageBody))
+		}
+		return fmt.Errorf("empty message accepted")
+	}
+	return nil
+}
+
+func GetMessageHandler(messageType byte, c net.Conn, uploader *WalUploader) SocketMessageHandler {
```
usernamedt, line 89: func `NewMessageHandler`

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		tracelog.ErrorLogger.Printf("Failed to read from socket, err: %v \n", err)
+		return 'E', nil, err
+	}
+	return messageType, messageBody, err
+}
+
+// HandleDaemon is invoked to perform daemon mode
+func HandleDaemon(uploader *WalUploader, pathToSocket string) {
+	_ = os.Remove(pathToSocket)
```
usernamedt, line 135: do not ignore the errors

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	l, err := net.Listen("unix", pathToSocket)
+	if err != nil {
+		tracelog.ErrorLogger.Fatal("Error on listening socket:", err)
+	}
+	for {
+		fd, err := l.Accept()
```
usernamedt, line 141: What if we would want to terminate the running daemon? 

I suggest to add the signal listener and listend for `SIGTERM` to perform a graceful shutdown. But looks like it is a feature for the next PR, not for the MVP.

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			tracelog.ErrorLogger.Println("Failed to accept, err:", err)
+		}
+		go DaemonProcess(fd, uploader)
```
usernamedt, line 145: same thing with verbs, for example you can name it like `Listen()`

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+
+// DaemonProcess reads data from connection and processes it
+func DaemonProcess(c net.Conn, uploader *WalUploader) {
+	defer utility.LoggedClose(c, fmt.Sprintf("Failed to close connection with %s \n", c.RemoteAddr()))
+	messageReader := GetMessageReader(c)
+	for {
+		messageType, messageBody, err := messageReader.Next()
+		if err != nil {
+			tracelog.ErrorLogger.Printf("Failed to read  message from client %s, err: %v\n", c.RemoteAddr(), err)
+			_, _ = c.Write([]byte{'E'})
```
usernamedt, line 157: do not ignore the possible errors

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+			return
+		}
+		messageHandler := GetMessageHandler(messageType, c, uploader)
+		if messageHandler == nil {
+			tracelog.ErrorLogger.Printf("Unexpected message type: %s", string(messageType))
+			_, _ = c.Write([]byte{'E'})
```
usernamedt, line 163: Create these mesage types as well:
```
type ErrorType SocketMessageType = 'E'
type ... SocketMessageType = 'F'
type ... SocketMessageType = 'O'
...
```

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+			return
+		}
+		err = messageHandler.Handle(messageBody)
+		if err != nil {
+			tracelog.ErrorLogger.Println("Failed to handle message, err:", err)
+			_, _ = c.Write([]byte{'E'})
+			return
+		}
+		if messageType == 'F' {
```
usernamedt, line 172: Log this somehow

usernamedt (changes requested):

internal/databases/postgres/daemon_handler.go
```diff
@@ -0,0 +1,176 @@
...
+	_, err = h.fd.Write([]byte{'O'})
+	if err != nil {
+		tracelog.ErrorLogger.Printf("failed to write in socket: %v\n", err)
+		return err
+	}
+	return nil
+}
+
+func messageValidation(messageBody []byte) error {
+	if len(messageBody) < 24 {
```
usernamedt, line 79: Also, we archive not only WALs but other files with length other than 24.

I think we can remove this validation.

usernamedt (changes requested):

usernamedt (changes requested):

usernamedt (approved):

PR #1357, opened 2022-10-11
Replace mongodb binary backup to tar archive format

KhurtinDN:


internal/databases/mongo/binary/concurrent_downloader.go
```diff
@@ -0,0 +1,65 @@
...
+
+	for _, tarObject := range tarObjects {
+		tarToExtract := internal.NewStorageReaderMaker(tarsFolder, tarObject.GetName())
+		tarsToExtract = append(tarsToExtract, tarToExtract)
+	}
+
+	return tarsToExtract, nil
+}
+
+func IsTarBackupFormat(uploader *internal.Uploader, backupName string) bool {
```
usernamedt, line 61: Maybe store such information in a backup sentinel file instead?

internal/databases/mongo/binary/dto.go
```diff
@@ -2,28 +2,42 @@ package binary
...
 }
 
-type Checksum struct {
-	Algorithm string `json:"Algorithm"`
-	Data      string `json:"Data"`
+type BackupFileMeta struct {
+	Path        string      `json:"Path"`
+	FileMode    os.FileMode `json:"FileMode"`
+	Compression string      `json:"Compression"` // used only for restoring old backups
+	FileSize    int64       `json:"-"`
```
usernamedt, line 18: is it intended?

internal/databases/mongo/binary/storage_local.go
```diff
@@ -120,8 +118,8 @@ func (localStorage *LocalStorage) CreateDirectories(directoryMetas []*BackupDire
 		if err != nil {
 			return err
 		}
-		userName := "mongodb"  // todo: get from sentinel!!!
-		groupName := "mongodb" // todo: get from sentinel!!!
+		userName := "mongodb"
```
usernamedt, line 121: Why not get it from sentinel? Maybe introduce a config option? Also, maybe just remove the chown logic completely and run WAL-G from the appropriate user?

internal/databases/mongo/binary/dto.go
KhurtinDN: yes, I want skip it in json, but it used during backup process.

internal/databases/mongo/binary/storage_local.go
KhurtinDN: it is legacy, need only to restore from old binary backup without tar archive. I want delete all legacy code.

usernamedt (approved):

VGoshev (approved):

PR #1359, opened 2022-10-18
Fix closing mongodb backup cursor

KhurtinDN:


internal/databases/mongo/binary/mongod.go
```diff
@@ -105,7 +106,36 @@ func (mongodService *MongodService) GetBackupCursor() (cursor *mongo.Cursor, err
 			time.Sleep(time.Minute * minutes)
 		}
 	}
-	return cursor, err
+
+	backupContext, stopPingBackupCursor := context.WithCancel(mongodService.Context)
+	var closeBackupCursorWg sync.WaitGroup
```
usernamedt, line 111: Maybe waitgroup is overkill here? we can just use a simple channel instead.

internal/databases/mongo/binary/mongod.go
```diff
@@ -105,7 +106,36 @@ func (mongodService *MongodService) GetBackupCursor() (cursor *mongo.Cursor, err
...
+	closeBackupCursor = func() {
+		stopPingBackupCursor()
+		closeBackupCursorWg.Wait()
+		closeErr := cursor.Close(mongodService.Context)
+		if closeErr != nil {
+			tracelog.ErrorLogger.Printf("Unable to close backup cursor: %+v", closeErr)
+		}
+	}
+
+	return cursor, closeBackupCursor, err
```
usernamedt, line 138: Maybe instead wrap the `*mongo.Cursor` into another struct and implement `io.Closer` for it?

usernamedt (changes requested):

internal/databases/mongo/binary/mongod.go
KhurtinDN: I used channel to notify closing before waitgroup, but I don't "wait" closing backup cursor and I think that backup cursor don't close. Maybe problem was in other places. Here I wait closing backup cursor explicitly

usernamedt (approved):

VGoshev (approved):

PR #1362, opened 2022-10-24
add timeout for s3 upload operations

moridin26:
# Pull request description

Fix will allow to set timeout for s3-upload operations to prevent wal-g hanging for indefinite time in case of some unavailability of s3 server

It's not really a bug, I was able to repeate this wal-g behaviour by editing iptables rules (and allowing/denying requests to s3). The issue is that if s3 server is for some reason unavailable, wal-g wouldn't just drop some error, but will hang for some indefinite time. This may cause databases locks unreleased for some time. 

After fix if wal-g has in it's config S3_UPLOAD_WITH_TIMEOUT = true and non-zero value of S3_UPLOAD_TIMEOUT, wal-g will drop request after timeout passed with error of "context deadline exceeded"

### DIFF
--- a/pkg/storages/s3/folder.go
+++ b/pkg/storages/s3/folder.go
@@ -1,6 +1,7 @@
 package s3
 
 import (
+	"context"
 	"io"
 	"path"
 	"strconv"
@@ -41,6 +42,8 @@ const (
 	UseListObjectsV1         = "S3_USE_LIST_OBJECTS_V1"
 	RangeBatchEnabled        = "S3_RANGE_BATCH_ENABLED"
 	RangeQueriesMaxRetries   = "S3_RANGE_MAX_RETRIES"
+	UploadWithTimeout        = "S3_UPLOAD_WITH_TIMEOUT"
+	UploadTimeout            = "S3_UPLOAD_TIMEOUT"
 	// MaxRetriesSetting limits retries during interaction with S3
 	MaxRetriesSetting = "S3_MAX_RETRIES"
 
@@ -78,6 +81,8 @@ var (
 		RangeBatchEnabled,
 		RangeQueriesMaxRetries,
 		MaxRetriesSetting,
+		UploadWithTimeout,
+		UploadTimeout,
 	}
 )
 
@@ -278,10 +283,20 @@ func (folder *Folder) listObjectsPagesV1(prefix *string, delimiter *string,
 		Prefix:    prefix,
 		Delimiter: delimiter,
 	}
-	return folder.S3API.ListObjectsPages(s3Objects, func(files *s3.ListObjectsOutput, lastPage bool) bool {
-		listFunc(files.CommonPrefixes, files.Contents)
-		return true
-	})
+	uploadTimeout := folder.getUploadTimeout()
+	if uploadTimeout != 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(uploadTimeout * int(time.Second)))
+		defer cancel()
+		return folder.S3API.ListObjectsPagesWithContext(ctx, s3Objects, func(files *s3.ListObjectsOutput, lastPage bool) bool {
+			listFunc(files.CommonPrefixes, files.Contents)
+			return true
+		})
+	} else {
+		return folder.S3API.ListObjectsPages(s3Objects, func(files *s3.ListObjectsOutput, lastPage bool) bool {
+			listFunc(files.CommonPrefixes, files.Contents)
+			return true
+		})
+	}
 }
 
 func (folder *Folder) listObjectsPagesV2(prefix *string, delimiter *string,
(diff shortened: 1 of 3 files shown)


pkg/storages/s3/folder.go
```diff
@@ -41,6 +42,8 @@ const (
 	UseListObjectsV1         = "S3_USE_LIST_OBJECTS_V1"
 	RangeBatchEnabled        = "S3_RANGE_BATCH_ENABLED"
 	RangeQueriesMaxRetries   = "S3_RANGE_MAX_RETRIES"
+	UploadWithTimeout        = "S3_UPLOAD_WITH_TIMEOUT"
```
usernamedt, line 45: Why introduce the two settings?

We can add only the `S3_UPLOAD_TIMEOUT` setting. If it is unset or set to zero it basically means that timeout is disabled.

pkg/storages/s3/folder.go
```diff
@@ -278,10 +283,20 @@ func (folder *Folder) listObjectsPagesV1(prefix *string, delimiter *string,
...
 	}
-	return folder.S3API.ListObjectsPages(s3Objects, func(files *s3.ListObjectsOutput, lastPage bool) bool {
-		listFunc(files.CommonPrefixes, files.Contents)
-		return true
-	})
+	uploadTimeout := folder.getUploadTimeout()
+	if uploadTimeout != 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(uploadTimeout * int(time.Second)))
+		defer cancel()
+		return folder.S3API.ListObjectsPagesWithContext(ctx, s3Objects, func(files *s3.ListObjectsOutput, lastPage bool) bool {
```
usernamedt, line 290: The only difference is the context, so we can just alter this part instead of duplicating the entire call to `folder.S3API.ListObjectsPages`. I mean, for no timeout we can just pass `context.Background()` and `context.WithTimeout()` in other case.

pkg/storages/s3/uploader.go
```diff
@@ -74,8 +77,16 @@ func (uploader *Uploader) createUploadInput(bucket, path string, content io.Read
 }
 
 func (uploader *Uploader) upload(bucket, path string, content io.Reader) error {
+	var err error
+	uploadTimeout := uploader.uploadTimeout
 	input := uploader.createUploadInput(bucket, path, content)
-	_, err := uploader.uploaderAPI.Upload(input)
+	if uploadTimeout != 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(uploadTimeout * int(time.Second)))
```
usernamedt, line 84: This part repeats too many times, just move it to some function, e.g. `configureUploadContext`

pkg/storages/s3/uploader.go
```diff
@@ -139,6 +150,24 @@ func configureUploader(s3Client *s3.S3, settings map[string]string) (*Uploader,
 		maxPartSize = DefaultMaxPartSize
 	}
 
+	uploadWithTimeout := false
+	if strUploadWithTimeout, ok := settings[UploadWithTimeout]; ok {
```
usernamedt, line 154: Same code as in the `getUploadTimeout()` function above.

pkg/storages/s3/folder.go
usernamedt: Also, I see that upload timeout affects not only the upload part but also the listings. Are you sure about the naming?

usernamedt (changes requested):

usernamedt:
Also, it would be great to add some notes about this settings to the docs.

pkg/storages/s3/uploader.go
```diff
@@ -74,8 +77,16 @@ func (uploader *Uploader) createUploadInput(bucket, path string, content io.Read
...
 
 func (uploader *Uploader) upload(bucket, path string, content io.Reader) error {
+	var err error
+	uploadTimeout := uploader.uploadTimeout
 	input := uploader.createUploadInput(bucket, path, content)
-	_, err := uploader.uploaderAPI.Upload(input)
+	if uploadTimeout != 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(uploadTimeout * int(time.Second)))
+		defer cancel()
+		_, err = uploader.uploaderAPI.UploadWithContext(ctx, input)
```
mialinx, line 86: Single timeout to the whole upload is not very useful.
The upload may take 10 minutes or 10 hours in normal situation depending on the file size.

So it will be difficult to choose the one timeout to fit them all.

pkg/storages/s3/folder.go
```diff
@@ -319,6 +344,28 @@ func (folder *Folder) partitionToObjects(keys []string) []*s3.ObjectIdentifier {
 	return objects
 }
 
+func (folder *Folder) getUploadTimeout() int {
```
mialinx, line 347: It would be more convenient to create `getUploadContext` method, returning `context.Context`
Then we could get rid of `if`’s in all methods.

But see another comments first.

mialinx (changes requested):

pkg/storages/s3/folder.go
moridin26: there is also difference in called functions: ListObjectsPagesWithContext and ListObjectsV2PagesWithContext

about namings: yes, I'll change it

Issue #1324, opened 2022-08-23
Debug logs may leak storage secrets


x4m:
Currently WALG_LOG_LEVEL=DEVEL will output all S3\GCP\Azure keys. I think we should mark some storage params as secrets and never under any circumstances output it.

LeGEC:
as mentioned in the PR I opened :

PR #1368 proposes a fix, currently coded on top of :
#1367 - Only dump relevant env variables with LOG_LEVEL=DEVEL

PR #1368 requires a review of the settings marked as "secret" :  
I built that list by trying to make educated guesses from the names of the constants, and their usage in wal-g's code, but I am in no way an expert in many of the involved libraries (sqlserver, gp, aws, swift ...), and someone should check whether the list I built is correct (do we want to hide all of these values ?) and complete (are these all the settings that need to be hidden ?)

usernamedt:
Fixed in https://github.com/wal-g/wal-g/pull/1368 and https://github.com/wal-g/wal-g/pull/1367

PR #1368, opened 2022-11-02
Do not print secrets in debug logs

LeGEC:
### Database name

# Pull request description
Hide secrets from `DEVEL` env dump

### Describe what this PR fix

proposed fix for
#1324 - Debug logs may leak storage secrets
coded on top of 
#1367 - Only dump relevant env variables with LOG_LEVEL=DEVEL

Before :

```
$ export PGPASSWORD=super_secret
$ WALG_LOG_LEVEL=DEVEL ./wal-g help
DEBUG: 2022/11/02 15:03:27.415603 --- COMPILED ENVIRONMENT VARS ---
	PGBACKREST_STANZA=main
	PGPASSWORD=super_secret        # <- leaked secret
	TOTAL_BG_UPLOADED_LIMIT=32
	...
```

After :

```
$ export PGPASSWORD=super_secret
$ WALG_LOG_LEVEL=DEVEL ./wal-g help
DEBUG: 2022/11/02 15:07:53.013343 --- COMPILED ENVIRONMENT VARS ---
	PGBACKREST_STANZA=main
	PGPASSWORD=--HIDDEN--            # <- mentioned, but redacted
	TOTAL_BG_UPLOADED_LIMIT=32
	...
```

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -69,6 +69,7 @@ const (
 	PgUserSetting                = "PGUSER"
 	PgHostSetting                = "PGHOST"
 	PgPasswordSetting            = "PGPASSWORD"
+	PgPassfileSetting            = "PGPASSFILE"
 	PgDatabaseSetting            = "PGDATABASE"
 	PgSslModeSetting             = "PGSSLMODE"
 	PgSlotName                   = "WALG_SLOTNAME"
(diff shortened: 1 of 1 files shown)


LeGEC:
I submitted this PR as an additional two commits on top of #1367, but the last two commits can be cherry-picked without changing the output format of the env variables in Debug logs -- there may be a merge conflict, easy to resolve, to port the following part :

```
	// for secret settings: leave them empty if they are defined but empty, otherwise hide their actual value
	if secretSettings[k] && val != "" {
		val = "--xxxx--"
	}
	fmt.Fprintf(&buff, "\t%s=%s\n", k, val)
```

internal/config.go
```diff
@@ -522,12 +562,31 @@ func Configure() {
...
+
+		for _, k := range keys {
+			val, ok := os.LookupEnv(k)
+			if !ok {
+				continue
+			}
+
+			// for secret settings: leave them empty if they are defined but empty, otherwise hide their actual value
+			if secretSettings[k] && val != "" {
+				val = "--xxxx--"
```
usernamedt, line 584: Maybe state explicitly that it is a hidden setting, e.g. `val = "--HIDDEN--"`?

usernamedt (approved):

PR #1371, opened 2022-11-07
notify_d watchdog for daemon service

rogaliiik:
PostgreSQL_15

Notify daemon service watchdog about working

### DIFF
--- a/internal/databases/postgres/daemon_handler.go
+++ b/internal/databases/postgres/daemon_handler.go
@@ -17,6 +17,8 @@ import (
 type SocketMessageType byte
 
 const (
+	SdNotifyWatchdog = "WATCHDOG=1"
+
 	CheckType    SocketMessageType = 'C'
 	FileNameType SocketMessageType = 'F'
 	OkType       SocketMessageType = 'O'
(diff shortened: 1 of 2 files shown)


internal/databases/postgres/daemon_handler.go
```diff
@@ -166,3 +168,25 @@ func Listen(c net.Conn, uploader *WalUploader) {
 		}
 	}
 }
+
+const SdNotifyWatchdog = "WATCHDOG=1"
+
+func SdNotify(state string) error {
+	socketName, ok := os.LookupEnv("NOTIFY_SOCKET")
```
usernamedt, line 175: Use the viper config setting instead. You can check how other settings are implemented for reference.

internal/databases/postgres/daemon_handler.go
```diff
@@ -166,3 +168,25 @@ func Listen(c net.Conn, uploader *WalUploader) {
 		}
 	}
 }
+
+const SdNotifyWatchdog = "WATCHDOG=1"
```
usernamedt, line 172: We place consts and global vars on the top of the file

usernamedt (changes requested):

internal/databases/postgres/daemon_handler.go
usernamedt: Also, it is better to skip the notification if the config setting is unset.

usernamedt (approved):

usernamedt:
We can merge this after rebase.

PR #1013, opened 2021-06-13
MySQL partial backups

ChizhovAlexey:

PR #1374, opened 2022-11-08
Implement max file size for mysql

bularond:
Add config WALG_STREAM_SPLITTER_MAX_FILE_SIZE witch set maximum one file size for backup


ostinru:
@bularond , could you please elaborate on why do you want to limit file sizes? If I remember correctly, current implementation shouldn't keep whole files locally, so it is unclear what you are trying to improve here.

mialinx:
> @bularond , could you please elaborate on why do you want to limit file sizes? If I remember correctly, current implementation shouldn't keep whole files locally, so it is unclear what you are trying to improve here.

Main idea behind `max size` is to avoid uploading / downloading huge files, as it consumes a lot of memory (s3client multipart specific)

internal/stream_push_helper.go
```diff
@@ -39,13 +39,31 @@ func (uploader *SplitStreamUploader) PushStream(stream io.Reader) (string, error
...
-		dstPath := GetPartitionedStreamName(backupName, uploader.Compressor.FileExtension(), i)
-		tracelog.InfoLogger.Printf("Uploading... %v", dstPath)
-		errGroup.Go(func() error {
-			return uploader.PushStreamToDestination(reader, dstPath)
-		})
+	for partNumber := 0; partNumber < uploader.partitions; partNumber++ {
+		reader := readers[partNumber]
+		if uploader.maxFileSize != 0 {
+			tracelog.DebugLogger.Printf("Start backup part %d\n", partNumber)
+			filesReaders := splitmerge.SplitMaxSizeReader(reader, uploader.blockSize, uploader.maxFileSize)
```
mialinx, line 46: Does all the `fileReaders` read from `reader` at the same time ?

internal/splitmerge/splitmaxsizereader.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/wal-g/tracelog"
+	"io"
+)
+
+func SplitMaxSizeReader(reader io.Reader, blockSize int, maxSize int) <-chan io.ReadCloser {
+	out := make(chan io.ReadCloser)
+
+	go func() {
```
mialinx, line 11: why do we need a separate goroutine here ?

mialinx (changes requested):
redesign needed

ostinru:
FYI: we have PR that limits size of files for single-stream backups: https://github.com/wal-g/wal-g/pull/1013 (PR weirdly named).

docker/mysql_tests/scripts/base_tests/splited_file_stream.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+set -e -x
+
+. /usr/local/export_common.sh
+
+export WALE_S3_PREFIX=s3://mysqlfullxtrabackupbucket
```
ostinru, line 6: Let's use a separate bucket for each test.
Don't forget to create it in `docker-compose.yml`

mialinx:
> FYI: we have PR that limits size of files for single-stream backups: #1013 (PR weirdly named).

Yes, but... it does not support split/merge parallel processing.  And it stores redundant info (Filenames) in sentinel.

internal/stream_metadata.go
```diff
@@ -32,12 +33,23 @@ func GetBackupStreamFetcher(backup Backup) (StreamFetcher, error) {
 		return nil, err
 	}
 
+	limitedFileSize := false
+	if metadata.Type == LimitedFileSizeBackup {
+		limitedFileSize = true
+	}
+
 	switch metadata.Type {
+	case LimitedFileSizeBackup:
```
mialinx, line 42: It looks like thees two cases differ only in the last argument. May be join them together ?  And get rid of temporary  `limitedFileSize` variable

internal/stream_fetch_helper.go
```diff
@@ -132,11 +111,66 @@ func DownloadAndDecompressSplittedStream(backup Backup, blockSize int, extension
 	return lastErr
 }
 
-func detectPartitionsCount(backup Backup, decompressor compression.Decompressor) (int, error) {
+func downloadFile(backup Backup, decompressor compression.Decompressor, fileName string, writer io.WriteCloser) error {
```
mialinx, line 114: We have added 4 extra functions.

May we just modify `detectPartitionsCount` to return `[]int` (number of part for each partition).
And add one loop (over parts) in `DownloadAndDecompressSplittedStream` ?

mialinx (changes requested):

mialinx (changes requested):

mialinx (approved):
LGTM

PR #1379, opened 2022-11-13
Makefile: change compilation and linking of libbrotli to optional (and off by default)

### LINKED ISSUE #1378: option to disable libbrotli in Makefile
### Database name

all databases - affects using libbrotli when building wal-g

# Issue description

In its current state, the Makefile on `master` does not allow to build a pure go version of wal-g : it always builds and link libbrotli.

### Describe your problem

We have a project that uses pure go binaries, and wanted  to keep it that way once we chose to use wal-g. We thus compile our own version of wal-g, where we disabled brotli and chose a pure go implementation of zstd.

I suggest to add an option to completely disable libbrotli in the regular Makefile.

The downside of using that option at compile time is that it creates wal-g binaries which can't handle brotli anymore (especially: that wouldn't be able to read back previous files compressed with brotli).  
This could be addressed by using a pure go broti implementation -- https://github.com/andybalholm/brotli claims to do that but we haven't taken time to review it.

In our case, we are fine with disabling brotli compression/decompression entirely (as said above: we use zstd, and use a pure go library that implements it), but obviously projects that used brotli for previous backups should keep using versions of wal-g that support it.

LeGEC:
### Database name

all databases -- affects libbrotli at build time

# Pull request description

### Describe what this PR fix

fixes #1378 

~~add a `NO_LIBBROTLI` option to opt out of libbrotli at build time.~~

This PR has been modified to :
add a `USE_BROTLI` option in the Makefile to build and link libbrotli

this is a (arguably small) breaking change :
if you previously ran regular make commands :

```
make
make link_brotli
make pg_build
# etc ...
```

and still need to support the brotli compression format (for retrocompatibility), you now need to set `USE_BROTLI=1` at build time :
```
export USE_BROTLI=1
make
make link_brotli
make pg_build
# etc ...
```

or somehow set it in your build pipeline

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,11 @@ GOLANGCI_LINT_VERSION ?= "v1.37.0"
 REDIS_VERSION ?= "5.0.8"
 TOOLS_MOD_DIR := ./internal/tools
 
-BUILD_TAGS:=brotli
+BUILD_TAGS:=
+
+ifdef USE_BROTLI
+	BUILD_TAGS:=$(BUILD_TAGS) brotli
+endif
 
 ifdef USE_LIBSODIUM
 	BUILD_TAGS:=$(BUILD_TAGS) libsodium
@@ -233,7 +237,8 @@ install:
 	@echo "Nothing to be done. Use pg_install/mysql_install/mongo_install/fdb_install/gp_install... instead."
 
 link_brotli:
-	./link_brotli.sh
+	@if [ -n "${USE_BROTLI}" ]; then ./link_brotli.sh; fi
+	@if [ -z "${USE_BROTLI}" ]; then echo "info: USE_BROTLI is not set, skipping 'link_brotli' task"; fi
 
 link_libsodium:
 	@if [ ! -z "${USE_LIBSODIUM}" ]; then\
@@ -242,7 +247,7 @@ link_libsodium:
 
 unlink_brotli:
 	rm -rf vendor/github.com/google/brotli/*
-	mv tmp/brotli/* vendor/github.com/google/brotli/
+	if [ -n "${USE_BROTLI}" ] ; then mv tmp/brotli/* vendor/github.com/google/brotli/; fi
 	rm -rf tmp/brotli
 
 unlink_libsodium:
--- a/.github/workflows/dockertests.yml
+++ b/.github/workflows/dockertests.yml
@@ -62,6 +62,7 @@ jobs:
         if: steps.cache-images.outputs.cache-hit != 'true'
         run: make pg_save_image
     env:
+      USE_BROTLI: 1
       USE_LZO: 1
 
   test:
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -19,7 +19,15 @@ GOLANGCI_LINT_VERSION ?= "v1.37.0"
...
 TOOLS_MOD_DIR := ./internal/tools
 
-BUILD_TAGS:=brotli
+BUILD_TAGS:=
+
+# the original Makefile had 'brotli' always included in BUILD_TAGS
+# for this historical reason:
+#  - brotli is an *opt-out* feature (included by default, use NO_LIBBROTLI to *not* include it),
+#  - the other libs are opt-in features (not included by default, use USE_LIBSODIUM or USE_LZO to include them)
+ifndef NO_LIBBROTLI
```
usernamedt, line 28: I propose to rename it like `USE_BROTLI`, to match the `USE_LZO` and `USE_LIBSODIUM` options

usernamedt (dismissed):

Makefile
LeGEC: we certainly can, but this is a breaking change.

meaning :
* within wal-g project : your test suite and your build pipeline would have to be changed too (I don't think you want to relase versions of wal-g that do not work for existing users, am I correct ?)
* for outside users of the project that compile your project : updating to latest version would silently change the result of their build -- if they intended to keep libbrotli, they too would have to adapt their build scripts / test suites.

Is that kind of change ok with you ?

usernamedt: We can set it to `1` by default

LeGEC: happy to oblige :wink:

LeGEC:
I tried to update the workflow files (by mimicking places where `USE_LZO` or `USE_LIBSODIUM` was set), feel free to correct that part.

Also : I tried updating docs/README.md , do review if the places where `USE_BROTLI=1` is mentioned are relevant.

LeGEC:
Well, apparently, I missed the mongodb tests :

https://github.com/wal-g/wal-g/actions/runs/3481421063/jobs/5823020774#step:8:4982

LeGEC:
I pushed a commit where I *changed the tests for redis and mongodb* : the tests are run using zstd instead of brotli.

I would much rather not change those tests in this specific PR and confirm that the main change (e.g: set USE_BROTLI if you want to use brotli) still allow to run the tests that depend on brotli unmodified, but I fail to understand where I missed passing the `USE_BROTLI=1` environment variable.

Can someone have a second look at the modifications I committed, and point me to what step requires adding `USE_BROTLI=1` ?

serprex:
It's in the docker compose files you'd want to look at

serprex (approved):

LeGEC:
setting the env value in the relevant Docker files seems to do the job.

PR #1394, opened 2022-11-29
Implement retry download backup

bularond:
Add retry logic for downloading backup files (MySQL)

  Previously, a single network error while downloading a backup file would abort the whole restore. This PR adds the ability to retry downloads transparently.

  What's new
  - ReaderWithRetry (internal/ioextensions) — an io.ReadCloser wrapper that reopens the underlying reader and resumes on read errors. It tracks how many bytes were already read and skips them after reconnecting, so retries don't corrupt the
  stream.
  - New setting WALG_BACKUP_DOWNLOAD_MAX_RETRY (default 1) controls the max number of download attempts. The retry wrapper is only engaged when the value is > 1.
  - Wired retries into DownloadAndDecompressSplittedStream / downloadAndDecompressFile.

  Refactor
  - Moved WithSizeReader from internal to utility and added a WithSizeReadCloser variant for reuse.

  Tests
  - Added testutils.NetworkErrorFolder + UnexpectedEOFLimitReader to simulate connection drops after N bytes.
  - New tests TestBackup_Retry_NetworkError and TestSplitBackup_Retry_NetworkError verify that restores recover from mid-download network errors.


docs/MySQL.md
```diff
@@ -43,6 +43,23 @@ If base and binlogs backups are taken from the same host, this variable should b
...
+* `WALG_STREAM_SPLITTER_BLOCK_SIZE`
+
+To configure block size into which backup stream split. Block length size data is put into each partition in turn.
+
+* `WALG_STREAM_SPLITTER_MAX_FILE_SIZE`
+
+To configure max file size before compressing. If partition size become more than max file size, it split on several files.
+Backup file names have a suffix `_0000_0000.bz`.
+
+* `WALG_BACKUP_DOWNLOAD_MAX_RETRY`
```
usernamedt, line 60: s/WALG_BACKUP_DOWNLOAD_MAX_RETRY/WALG_BACKUP_DOWNLOAD_MAX_RETRIES/g

internal/reader_with_retrys.go
```diff
@@ -0,0 +1,77 @@
+package internal
```
usernamedt, line 1: we have the [ioextensions](https://github.com/wal-g/wal-g/tree/master/internal/ioextensions) package for io-related stuff, I guess this should be there

internal/reader_with_retrys.go
```diff
@@ -0,0 +1,77 @@
...
+
+		read, readErr := r.reader.Read(p[n:])
+		n += read
+		r.alreadyRead += read
+		err := readErr
+
+		if err == io.EOF {
+			return n, err
+		} else if err != nil {
+			tracelog.ErrorLogger.Printf("Error while download file: %v. Attempt: %d\n", err, attempt)
```
usernamedt, line 60: Well, this reader doesn't know anything about downloads, so I guess we shouldn't mention it.

internal/reader_with_retrys.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (r *ReaderWithRetry) Read(p []byte) (int, error) {
+	n := 0
+	for attempt := 0; attempt < r.retryAttempts; attempt++ {
+		if r.reader == nil {
+			err := r.setupNewReader()
+			if err == io.EOF {
+				return n, err
+			} else if err != nil {
+				tracelog.ErrorLogger.PrintOnError(r.reader.Close())
```
usernamedt, line 46: utility.LoggedClose

internal/reader_with_retrys.go
```diff
@@ -0,0 +1,77 @@
...
+				r.reader = nil
+				continue
+			}
+		}
+
+		read, readErr := r.reader.Read(p[n:])
```
usernamedt, line 52: why not assign to err instead of readErr?

utility/network_error_folder.go
```diff
@@ -0,0 +1,70 @@
+package utility
+
+import (
+	"io"
+	"sync"
+
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+)
+
+func NewUnexpectedEOFLimitReader(readCloser io.ReadCloser, maxRead int64) io.ReadCloser {
```
usernamedt, line 10: move to the ioextensions?

utility/network_error_folder.go
```diff
@@ -0,0 +1,70 @@
...
+
+func NewNetworkErrorFolder(sourceFolder storage.Folder, networkErrorAfterByteSize int) storage.Folder {
+	return &TestFolder{
+		Folder:       sourceFolder,
+		maxReadSize:  int64(networkErrorAfterByteSize),
+		readFromFile: make(map[string]*int64),
+	}
+}
+
+type TestFolder struct {
```
usernamedt, line 51: this should belong to the `_test.go` file

utility/with_size_reader.go
```diff
@@ -0,0 +1,40 @@
+package utility
+
+import (
+	"io"
+	"sync/atomic"
+)
+
+func NewWithSizeReader(underlying io.Reader, readSize *int64) *WithSizeReader {
```
usernamedt, line 8: same thing about ioextensions

usernamedt (changes requested):

internal/reader_with_retrys.go
bularond: LoggedClose not suitable, because r.reader may change while executing function or it should be stay "open"

internal/ioextensions/network_error_folder.go
```diff
@@ -0,0 +1,71 @@
...
+		err = io.ErrUnexpectedEOF
+	}
+	return
+}
+
+func (u *UnexpectedEOFLimitReader) Close() error {
+	return u.source.Close()
+}
+
+func NewNetworkErrorFolder(sourceFolder storage.Folder, networkErrorAfterByteSize int) storage.Folder {
```
mialinx, line 44: Is it used for tests only ? 
It's better to put it in the corresponding `_test.go` to avoid using in non-test code in the future.

utility/with_size_reader.go
mialinx, line 8: In most cases, modifying value by pointer is not a go way.
It would be better to create a separate method like `GetBytesRead() int` to return counter.

But, as far as I see, this code is used in tests and passing `*int` is handy there.
May be just move this code to `_test.go` files ?

internal/ioextensions/reader_with_retrys.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (r *ReaderWithRetry) Read(p []byte) (int, error) {
+	n := 0
+	for attempt := 0; attempt < r.retryAttempts; attempt++ {
+		if r.reader == nil {
+			err := r.setupNewReader()
+			if err == io.EOF {
+				return n, err
+			} else if err != nil {
+				tracelog.ErrorLogger.PrintOnError(r.reader.Close())
```
mialinx, line 46: 1) use Printf and add some context info
2) `Close()` error is useless, We would like to see `setupNewReader()` error instead

internal/ioextensions/reader_with_retrys.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+	_, err = io.CopyN(io.Discard, reader, int64(r.alreadyRead))
+	r.reader = reader
+	return err
+}
+
+func (r *ReaderWithRetry) Read(p []byte) (int, error) {
+	n := 0
+	for attempt := 0; attempt < r.retryAttempts; attempt++ {
```
mialinx, line 40: `Read` method will be called many times with small buffers.
We should make N attempts to download the whole file, not to perform single `Read` call.
It would be better to store `attempt` counter as  `ReaderWithRetry` field.

internal/ioextensions/reader_with_retrys.go
```diff
@@ -0,0 +1,76 @@
...
+			tracelog.ErrorLogger.Printf("error while read file: %v. Attempt: %d\n", err, attempt)
+			tracelog.ErrorLogger.PrintOnError(r.reader.Close())
+			r.reader = nil
+			continue
+		} else if n == len(p) {
+			break
+		}
+	}
+
+	return n, nil
```
mialinx, line 68: It looks like all attempts are exhausted and we should return an error here !
(Or should check last iteration on L62)

mialinx (changes requested):

mialinx (approved):
LGTM

reshke (approved):

usernamedt (approved):

PR #1400, opened 2022-12-14
Fixed remote backup for postgresql < 9.5

thepax:
PostgreSQL < 9.5 does't support the TABLESPACE_MAP option for BASE_BACKUP.

### DIFF
--- a/internal/databases/postgres/backup_push_handler.go
+++ b/internal/databases/postgres/backup_push_handler.go
@@ -50,6 +50,7 @@ func (err backupFromOtherBD) Error() string {
 type BackupArguments struct {
 	isPermanent           bool
 	verifyPageChecksums   bool
+	skipTablespaceMap     bool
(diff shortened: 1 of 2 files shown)


serprex:
Do we want to support postgres versions this old?

https://www.postgresql.org/support/versioning/

It's been EOL for awhile. In theory if someone wants to have reliable archival/restoration of EOL versions they should be using software packaged while the software wasn't EOL

In particular without testing older versions there's no reliability that we won't regress these older versions again in the future

x4m:
I do not know, but here are some thoughts.
When we tried to deprecate 9.6, @CyberDem0n was actively against this :)
pgBackReset supports 8.4+. It actually makes sense, because folks can have ancient backups.
But creating new backups from versions that are unsupported for 3 years with very novel functionality... It's kind of too much. However I don't see big problem in accepting a fix. But without fixes this fix might not last long.

PR #1403, opened 2022-12-21
Add restore-point-ts flag for Greenplum backup-fetch

usernamedt:
I propose to add the `--restore-point-ts` flag to the WAL-G backup-fetch command for Greenplum:
```bash
wal-g backup-fetch [OPTIONAL_BACKUP_NAME] --restore-point-ts "2022-07-05T01:01:50Z" --restore-config=/path/to/restore_config.json --config=/path/to/config.yaml
```
If specified, WAL-G will fetch the backup onto the restore point closest to the provided time


reshke (approved):

internal/databases/greenplum/restore_point_list_handler_test.go
```diff
@@ -1,243 +0,0 @@
-package greenplum_test
```
usernamedt, line 1: These tests don't actually test anything greenplum-related so I removed them

Issue #1108, opened 2021-10-05
Sort backups by the creation time in generic backup-list command


usernamedt:
Thanks to [this PR](https://github.com/wal-g/wal-g/issues/694), WAL-G for Postgres sorts the backups in the output of the `backup-list` command by the creation time. However, it does that only when being called with the `--detail` flag.

It happens because WAL-G uses two different  backup-list handlers:
1. Generic (same for all databases) when it is launched without the `--detailed` flag ([source code](https://github.com/wal-g/wal-g/blob/3c09d9d467dfcd4466f3654c76d8af68f0d2eacb/internal/backup_list_handler.go#L28))
2. Postgres-specific when being launched with the `--detail` flag ([source code](https://github.com/wal-g/wal-g/blob/3c09d9d467dfcd4466f3654c76d8af68f0d2eacb/internal/databases/postgres/backup_list_handler.go#L16))

I think that WAL-G generic backup list handler should utilize the [GenericMetadata](https://github.com/wal-g/wal-g/blob/3c09d9d467dfcd4466f3654c76d8af68f0d2eacb/internal/generic_metadata.go#L16) to sort the backups by the creation time. For some databases, the appropriate adapters are already written: [MySQL](https://github.com/wal-g/wal-g/blob/3c09d9d467dfcd4466f3654c76d8af68f0d2eacb/internal/databases/mysql/generic_meta_interactor.go#L23), [Postgres](https://github.com/wal-g/wal-g/blob/3c09d9d467dfcd4466f3654c76d8af68f0d2eacb/internal/databases/postgres/generic_meta_interactor.go#L23), [Greenplum](https://github.com/wal-g/wal-g/blob/3c09d9d467dfcd4466f3654c76d8af68f0d2eacb/internal/databases/greenplum/generic_meta_interactor.go#L23). As for FDB, Redis, and SQLServer which also use the generic backup-list command handler currently there is no GenericMetadata adapter implemented.

So, I can see the following plan to implement this feature:
1. Implement the GenericMetaFetcher for FDB, Redis, and SQLServer.
2. Fetch the `GenericMetadata` in the backup-list command handler to sort the backup-list output by the creation time.

OlleLarsson:
Hello @usernamedt, is this issue prioritized anywhere on your roadmap :smile: ?

x4m:
There's a big chance that we need to recruit some other principal author for the feature.
Maybe @debebantur knows anyone, but anyway PR would be welcomed.

[...]

PR #1405, opened 2022-12-26
Issue/1108

tewboat:
# Pull request description
[Issue] (https://github.com/wal-g/wal-g/issues/1108)
Added GenericMetaInteractors for:
-  Mongo
- Redis
- Fdb 
- SqlServers

Interactors are used for fetching backup StartTime to sort backups by creation time. 
Added sorting by creation time in DefaultHandleBackupList when `--detailed` flag is not specified.

### DIFF
--- a/internal/backup_util.go
+++ b/internal/backup_util.go
@@ -87,6 +87,26 @@ func GetBackups(folder storage.Folder) (backups []BackupTime, err error) {
 	return
 }
 
+// GetBackupsWithMetadata receives backup descriptions with meta information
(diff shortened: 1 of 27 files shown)


internal/backup_util.go
```diff
@@ -113,6 +140,27 @@ func SortBackupTimeSlices(backupTimes []BackupTime) {
 	})
 }
 
+func SortBackupTimeWithMetadataSlices(backupTimes []BackupTimeWithMetadata) {
+	order := ByCreationTime
```
usernamedt, line 144: Well, if we always sort by the creation time why separate if down below? I suggest to remove it

usernamedt (changes requested):
Overall looks good, will merge after fixing the issues.

PR #1411, opened 2023-01-08
Add tests to mysql backup list handler

deffer1337:
Add tests and fix format at writeBackupListDetails

### DIFF
--- a/internal/databases/mysql/backup_list_handler_test.go
+++ b/internal/databases/mysql/backup_list_handler_test.go
@@ -1,0 +1,119 @@
+package mysql
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var emptyColumnsBackups = []BackupDetail{
+	{
+		BackupName: "backupName",
+	},
+	{
+		Hostname: "HostName",
+	},
+}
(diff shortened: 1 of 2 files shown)


internal/databases/mysql/backup_list_handler_test.go
```diff
@@ -0,0 +1,119 @@
+package mysql
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var emptyColumnsBackups = []BackupDetail{
```
usernamedt, line 10: I suggest to also add the creation dates.

usernamedt (changes requested):

internal/databases/mysql/backup_list_handler_test.go
deffer1337: You mean that need to add the ModifyTime to emptyColumnsBackups, shortValuesBackups and longValuesBackups?

usernamedt: > You mean that need to add the ModifyTime to emptyColumnsBackups, shortValuesBackups and longValuesBackups?

Yes

PR #1417, opened 2023-01-16
[MongoDB] add opportunity restore mongodb backup with + 1 version

Apelsin234:


internal/databases/mongo/binary/utilities.go
```diff
@@ -19,11 +20,43 @@ func MajorMinorVersion(version string) string {
...
+}
+
+// NextMajorVersion return next major version in format '<MajorVersion> + 1' (without minor, patch, prerelease, build, ...)
+func NextMajorVersion(version string) string {
+	if len(version) == 0 {
+		return version
+	}
+
+	minorVersion := MajorMinorVersion(version)
+	if minorVersion == "v4.2" {
```
KhurtinDN, line 43: it is dirty hack. so you should leave comment about it =)

KhurtinDN (approved):

PR #1421, opened 2023-01-20
fix bug with mixed up arguments 

Apelsin234:


VGoshev (changes requested):
As we discussed, looks like issue not in your code, but in code, which uses this function

VGoshev (approved):

Issue #1422, opened 2023-01-26
Metrics are not pushed on upload failures because of Fatal logging


francoispqt:
### Database name
All of them

# Issue description
The metrics feature implemented [here](https://github.com/wal-g/wal-g/pull/1240) does not work properly on upload failures because the `pushMetrics` happens in the cobra post run hook which is not called because we `ErrorLogger.Fatal` when we encounter an error.

PR #1423, opened 2023-01-26
adding push metrics as a defer in the upload method to make sure metrics are pushed even on failure

francoispqt:
### Database name
All of them

# Pull request description
The metrics feature works fine to record uploads when they are successful but metrics are never pushed when an upload fails because Fatal logging exits directly and prevents the `PostRun` cobra hook from running.

To solve that I propose to call `PushMetrics` directly from the `uploader.Upload` method.

### Describe what this PR fix
It fixes https://github.com/wal-g/wal-g/issues/1422

### DIFF
--- a/internal/uploader.go
+++ b/internal/uploader.go
@@ -166,6 +166,8 @@ func (uploader *Uploader) Compression() compression.Compressor {
 
 // TODO : unit tests
 func (uploader *Uploader) Upload(path string, content io.Reader) error {
+	defer PushMetrics()
+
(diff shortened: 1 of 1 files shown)


usernamedt:
Ok, let's merge it as a temporary solution. But we need to solve the root cause eventually - handle fatal errors gracefully.

francoispqt:
Yes I agree, my take would be to stop using log.Fatal and bubble up errors but that's way more work

jurim76:
Any plans to fix it?

Issue #1117, opened 2021-10-11
Unknown flag --after 


nrot:
### Database name
Postgresql 13/14

# Issue description
Unknown flag --after in the delete command

### Describe your problem
Can`t use flag --after

### Please provide steps to reproduce
Use wal-g-pg-ubuntu-20.04-amd64 V1.1 for command:
/usr/local/bin/wal-g delete retain 30 --after \$(date -d '-30 days' '+\\%FT\\%TZ')

### Please add config and wal-g stdout/stderr logs for debug purpose
Error: unknown flag: --after
Usage:
  wal-g delete retain [FULL|FIND_FULL] backup_count [flags]

Examples:
  retain 5                      keep 5 backups
  retain FULL 5                 keep 5 full backups and all deltas of them
  retain FIND_FULL 5            find necessary full for 5th and keep everything after it
  retain 5 --after 2019-12-12T12:12:12   keep 5 most recent backups and backups made after 2019-12-12 12:12:12

Flags:
  -h, --help   help for retain

Global Flags:
      --aws-access-key string                    
      --aws-access-key-id string                 
      --aws-ca-bundle string                     
  ..........................................
      --walg-verify-page-checksums string        
      --yc-cse-kms-key-id string                 
      --yc-service-account-key-file string

unknown flag: --after

Issue #637, opened 2020-04-22
Unable to use flag `--after TIMESTAMP` when using `wal-g delete retain`


ilicmilan:
Version: `0.2.15`

There is an error that `--after` flag is unknown. Error output even shows an example of a command that uses `--after` flag
```
wal-g delete retain 1 --after "2020-04-22T18:00:05"
Error: unknown flag: --after
Usage:
  wal-g delete retain [FULL|FIND_FULL] backup_count [flags]

Examples:
  retain 5                      keep 5 backups
  retain FULL 5                 keep 5 full backups and all deltas of them
  retain FIND_FULL 5            find necessary full for 5th and keep everything after it
  retain 5 --after 2019-12-12T12:12:12   keep 5 most recent backups and backups made after 2019-12-12 12:12:12

Flags:
  -h, --help   help for retain

Global Flags:
      --config string   config file (default is $HOME/.walg.json)
      --confirm         Confirms backup deletion

unknown flag: --after
```

x4m:
@perekalov , do I understand correctly that you implemented this functionality for generic delete command, not only for MongoDB's ?

lord-alfred:
Someone forgot to add processing `DeleteRetainAfterArgsValidator` to the file: https://github.com/wal-g/wal-g/blob/master/cmd/pg/delete.go#L34-L39

This argument is not working at the moment.

PR #1424, opened 2023-01-27
Add --after option support for PostgreSQL delete retain command

victor-tsykanov:
### Database name
PostgreSQL

# Pull request description

This PR fixes **--after** option support for **delete retain** command for PostgreSQL. The option allows to specify timestamp or backup name after which backups should be kept.

There is an example of using **--after** option in **delete retain** command description but trying to apply it results in error.

This problem is mentioned in #1117 and #637.

### Steps to reproduce

Run command `wal-g delete retain 5 --after 2019-12-12T12:12:12`

Error output: `Error: unknown flag: --after`

### DIFF
--- a/cmd/pg/delete.go
+++ b/cmd/pg/delete.go
@@ -16,6 +16,7 @@ const DeleteGarbageExamples = `  garbage           Deletes outdated WAL archives
   garbage ARCHIVES  Deletes only outdated WAL archives from storage
   garbage BACKUPS   Deletes only leftover backups files from storage`
 const DeleteGarbageUse = "garbage [ARCHIVES|BACKUPS]"
+const afterFlag = "after"
 
 var confirmed = false
 var useSentinelTime = false
@@ -38,6 +39,7 @@ var deleteRetainCmd = &cobra.Command{
 	Use:       internal.DeleteRetainUsageExample, // TODO : improve description
 	Example:   internal.DeleteRetainExamples,
 	ValidArgs: internal.StringModifiers,
+	Args:      internal.DeleteRetainArgsValidator,
 	Run:       runDeleteRetain,
 }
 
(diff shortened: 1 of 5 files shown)


usernamedt:
Hi! Can you also add a simple test like this to make sure it won't break in the future? https://github.com/wal-g/wal-g/blob/master/docker/pg_tests/scripts/tests/delete_retain_full_test.sh

InteNs:
would love to see this merged!

cmd/pg/delete.go
```diff
@@ -81,7 +82,12 @@ func runDeleteRetain(cmd *cobra.Command, args []string) {
 	deleteHandler, err := postgres.NewDeleteHandler(folder, permanentBackups, permanentWals, useSentinelTime)
 	tracelog.ErrorLogger.FatalOnError(err)
 
-	deleteHandler.HandleDeleteRetain(args, confirmed)
+	afterValue, _ := cmd.Flags().GetString("after")
```
usernamedt, line 85: extract to const

usernamedt:
> would love to see this merged!

We still would like to see some test like [this one](https://github.com/wal-g/wal-g/blob/master/docker/pg_tests/scripts/tests/delete_retain_full_test.sh). Unfortunately, there has been many cases with breaking of `delete ***` commands so we won't accept any new ones without tests.

cmd/pg/delete.go
victor-tsykanov: Done

victor-tsykanov:
Test for `delete retain --after` added

egorstarikov:
Hi, guys! Any news for this?

x4m:
Ugh, sorry, we missed this.

x4m (approved):

x4m:
Pushed, thanks!

PR #1427, opened 2023-01-30
Print info logger messages to stdout

ostinru:
# Pull request description

bump github.com/wal-g/tracelog version

### Describe what this PR fix

https://github.com/wal-g/wal-g/issues/1425 wal-g normal output written to stderr not stdout

PR #1430, opened 2023-02-03
Use latest wal-g/tracelog module

joewildish:
The latest `wal-g/tracelog` module supports the `ERROR` level which suppresses all logs except for those at `ERROR`.

### Database name
N/A

# Pull request description

### Describe what this PR fix
NORMAL logging level is too verbose. The `wal-g/tracelog` has already been amended to introduce a quieter level, so lets move to using that.

N/A


ostinru:
It looks like a duplicate of PR https://github.com/wal-g/wal-g/pull/1427
It broke some tests... haven't time to figure out what went wrong

joewildish:
It looks like parts of the tests inspect standard error to deduce certain facts about what `wal-g` can see e.g. backup lists.  Clearly changing the logger to use stdout in some circumstances could break the deductions.

Before trying to change those tests, I attempted to run the test suite as per the github runner.  This turned out to be remarkably complicated and I was unable to get much of the test suite running.  I tried on two different computers--- macOS M1 and Linux Fedora x86.  I know that Docker has some issues on macOS arm kit so thought that may be why I was having trouble; it wasn't.

The latest commit on this PR fixes the pg test properly, but, for the other tests, I have done a dumb thing of `... 2>&1 | ...` where ever such a pipe out from `wal-g` happens.  I have no idea if this will work or not as I can't run the tests properly.

I have not attempted to fix the mongo tests. I noticed that there are various wrappers written in go itself that choose to inspect stdout and stderr streams, so there is no simple fix of `2>&1` (assuming such a simple hack would (a) actually make the tests pass and (b) could even be considered a fix).

Do you know anything about the mongo tests? Do you think they are fixable?

usernamedt:
> The latest commit on this PR fixes the pg test properly, but, for the other tests, I have done a dumb thing of `... 2>&1 | ...` where ever such a pipe out from `wal-g` happens. I have no idea if this will work or not as I can't run the tests properly.

Hi! If you have troubles running the docker test on your machine, you can commit directly to your fork's master branch to trigger the CI :) Just my 5 cents.

PR #1440, opened 2023-03-09
Fix data race in uploader tests

usernamedt:
Fix data race introduced in https://github.com/wal-g/wal-g/pull/1423. For some reason, there were no CI tests triggered.


serprex:
Why not revert #1423 until this is fixed?

usernamedt:
> Why not revert #1423 until this is fixed?

I thought it would be an easy fix... :) But I guess you are right. I'll create a PR shortly.

usernamedt:
> Why not revert #1423 until this is fixed?

https://github.com/wal-g/wal-g/pull/1442

usernamedt:
I'll reopen it with the correct branch.

PR #1447, opened 2023-03-20
Database meta

Catmoonlight:
### Database meta

Adds additional field in filesMetadata to store database names. Now databases can be restored during partial backup with names specifying.
Automatically restores system database.

### DIFF
--- a/cmd/pg/backup_fetch.go
+++ b/cmd/pg/backup_fetch.go
@@ -29,7 +29,7 @@ var restoreSpec string
 var reverseDeltaUnpack bool
 var skipRedundantTars bool
 var fetchTargetUserData string
-var onlyDatabases []int
+var onlyDatabases []string
 
 var backupFetchCmd = &cobra.Command{
 	Use:   "backup-fetch destination_directory [backup_name | --target-user-data <data>]",
@@ -99,7 +99,7 @@ func init() {
 		false, skipRedundantTarsDescription)
 	backupFetchCmd.Flags().StringVar(&fetchTargetUserData, "target-user-data",
 		"", targetUserDataDescription)
-	backupFetchCmd.Flags().IntSliceVar(&onlyDatabases, "restore-only",
+	backupFetchCmd.Flags().StringSliceVar(&onlyDatabases, "restore-only",
 		nil, restoreOnlyDescription)
 
 	Cmd.AddCommand(backupFetchCmd)
--- a/internal/databases/postgres/backup_push_handler.go
+++ b/internal/databases/postgres/backup_push_handler.go
@@ -149,7 +149,7 @@ func (bh *BackupHandler) createAndPushBackup() {
 	tracelog.ErrorLogger.FatalOnError(err)
 	bh.handleDeltaBackup(folder)
 	tarFileSets := bh.uploadBackup()
-	sentinelDto, filesMetaDto := bh.setupDTO(tarFileSets)
+	sentinelDto, filesMetaDto, err := bh.setupDTO(tarFileSets)
 	bh.markBackups(folder, sentinelDto)
 	bh.uploadMetadata(sentinelDto, filesMetaDto)
 
@@ -210,7 +210,8 @@ func (bh *BackupHandler) handleDeltaBackup(folder storage.Folder) {
 	}
 }
 
-func (bh *BackupHandler) setupDTO(tarFileSets internal.TarFileSets) (sentinelDto BackupSentinelDto, filesMeta FilesMetadataDto) {
+func (bh *BackupHandler) setupDTO(tarFileSets internal.TarFileSets) (sentinelDto BackupSentinelDto,
+	filesMeta FilesMetadataDto, err error) {
 	var tablespaceSpec *TablespaceSpec
 	if !bh.Workers.Bundle.TablespaceSpec.empty() {
 		tablespaceSpec = &bh.Workers.Bundle.TablespaceSpec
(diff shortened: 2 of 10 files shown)


internal/databases/postgres/names_metadata.go
```diff
@@ -0,0 +1,15 @@
+package postgres
+
+import (
+	"fmt"
+)
+
+type PathsByNamesMetadata map[string][]string
```
usernamedt, line 7: Lets remove the Metadata suffix and leave only the `PathByDbNames`? Since it is not a separate metadata but a part of another one.

usernamedt:
Acutally, I am not sure why we should lookup the database paths by its name, we can just store the `dbname <-> oid` mapping and it should be enough?

internal/databases/postgres/backup_push_handler.go
```diff
@@ -407,6 +411,16 @@ func (bh *BackupHandler) uploadMetadata(sentinelDto BackupSentinelDto, filesMeta
 	}
 }
 
+func (bh *BackupHandler) collectNamesMetadata() (PathsByNamesMetadata, error) {
```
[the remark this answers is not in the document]
Catmoonlight: Going to store every database and table name in future, so ~~ half of filesMetadata size

internal/databases/postgres/query_runner.go
```diff
@@ -380,6 +378,24 @@ func (queryRunner *PgQueryRunner) GetDatabaseInfos() ([]PgDatabaseInfo, error) {
 	return databases, nil
 }
 
+// GetAllowConDatabaseInfos fetches a list of all databases in cluster which are allowed to connect
+func (queryRunner *PgQueryRunner) GetAllowConDatabaseInfos() ([]PgDatabaseInfo, error) {
```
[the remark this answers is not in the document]
Catmoonlight: Rewritten to old GetDatabaseInfos. Later will write new query for collecting all metadata

internal/databases/postgres/backup_push_handler.go
```diff
@@ -320,7 +323,7 @@ func (bh *BackupHandler) handleBackupPushRemote() {
 	}
 	// If no arg is parsed, try to run remote backup using pglogrepl's BASE_BACKUP functionality
 	tracelog.InfoLogger.Println("Running remote backup through Postgres connection.")
-	tracelog.InfoLogger.Println("Features like delta backup are disabled, there might be a performance impact.")
+	tracelog.InfoLogger.Println("Features like delta backup and collecting database names are disabled, there might be a performance impact.")
```
usernamedt, line 326: same as in docs, I think we shouldn't be too specific

internal/databases/postgres/extract_provider_with_spec.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (p ExtractProviderDBSpec) Get(
+	backup Backup,
+	filesToUnwrap map[string]bool,
+	skipRedundantTars bool,
+	dbDataDir string,
+	createNewIncrementalFiles bool,
+) (IncrementalTarInterpreter, []internal.ReaderMaker, string, error) {
+	_, filesMeta, _ := backup.GetSentinelAndFilesMetadata()
```
usernamedt, line 33: do not ignore this error

internal/databases/postgres/extract_provider_with_spec.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return p.ExtractProviderImpl.Get(backup, filesToUnwrap, skipRedundantTars, dbDataDir, createNewIncrementalFiles)
+}
+
+func (p ExtractProviderDBSpec) makeFullRestoreDatabaseMap(databases []string, names DatabasesByNames) map[int]bool {
+	restoredDatabases := p.makeSystemDatabasesMap()
+
+	for _, db := range databases {
+		dbID, err := names.Resolve(db)
+		if err == nil {
```
usernamedt, line 46: What if we failed to resolve the ID? I guess that we should not ignore it silently.

internal/databases/postgres/extract_provider_with_spec.go
```diff
@@ -0,0 +1,90 @@
...
+	for file := range filesToUnwrap {
+		isDB, dbID, _ := p.tryGetOidPair(file)
+
+		if isDB && !databases[dbID] {
+			delete(filesToUnwrap, file)
+		}
+	}
+}
+
+func (p ExtractProviderDBSpec) tryGetOidPair(file string) (bool, int, int) {
```
usernamedt, line 72: This needs some unit tests.

usernamedt (changes requested):

cmd/pg/backup_fetch.go
```diff
@@ -20,16 +20,16 @@ For information about pattern syntax view: https://golang.org/pkg/path/filepath/
 	reverseDeltaUnpackDescription = "Unpack delta backups in reverse order (beta feature)"
 	skipRedundantTarsDescription  = "Skip tars with no useful data (requires reverse delta unpack)"
 	targetUserDataDescription     = "Fetch storage backup which has the specified user data"
-	restoreOnlyDescription        = `[Experimental] Downloads only databases specified by passed db ids from default tablespace.
```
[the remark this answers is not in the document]
Catmoonlight: Spaces are supported by Cobra. --restore-only="my database" works fine
 
This option supports digit names too. If digit name is not found in metadata, tries to interpret number as oid

serprex:
> Now databases can be restored during partial backup with names specifying.

Do you mean without specifying names?

Catmoonlight:
> 

With these changes you need to pass database name as a parameter of partial restore. Previously oids were used. So, no, names should be specified.

usernamedt (approved):

PR #1453, opened 2023-04-06
Tables partial restore for Postgresql

Catmoonlight:
Implements tables partial restore: with flag `--restore-only=database/my_table` only system tables and `my_table` will be completely restored. Useful for skipping big tables (>1GB).

### DIFF
--- a/internal/databases/postgres/query_runner.go
+++ b/internal/databases/postgres/query_runner.go
@@ -478,30 +478,41 @@ func (queryRunner *PgQueryRunner) ForEachDatabase(
 	}
 
 	for _, db := range databases {
-		dbName := db.Name
-		databaseOption := func(c *pgx.ConnConfig) error {
-			c.Database = dbName
-			return nil
-		}
-		dbConn, err := Connect(databaseOption)
-
-		if err != nil {
-			tracelog.WarningLogger.Printf("Failed to connect to database: %s\n'%v'\n", db.Name, err)
-			continue
-		}
-		runner, err := NewPgQueryRunner(dbConn)
-		if err != nil {
-			return errors.Wrap(err, "Failed to build query runner.")
-		}
-
-		err = function(runner, db)
+		err := queryRunner.executeForDatabase(function, db)
 		if err != nil {
 			return err
 		}
-
-		err = dbConn.Close()
-		tracelog.WarningLogger.PrintOnError(err)
 	}
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


internal/databases/postgres/query_runner.go
```diff
@@ -469,3 +469,84 @@ func (queryRunner *PgQueryRunner) Ping() error {
...
+		if err != nil {
+			return errors.Wrap(err, "Failed to build query runner.")
+		}
+
+		err = function(runner, db)
+		if err != nil {
+			return err
+		}
+
+		err = dbConn.Close()
```
usernamedt, line 502: use defer instead

usernamedt (changes requested):

internal/databases/postgres/query_runner.go
```diff
@@ -478,30 +478,41 @@ func (queryRunner *PgQueryRunner) ForEachDatabase(
...
+		return nil
+	}
+	dbConn, err := Connect(databaseOption)
+	if err != nil {
+		tracelog.WarningLogger.Printf("Failed to connect to database: %s\n'%v'\n", db.Name, err)
+		return nil
+	}
+	defer func() {
+		err := dbConn.Close()
 		tracelog.WarningLogger.PrintOnError(err)
```
usernamedt, line 503: defer utility.LoggedClose(dbConn)

internal/databases/postgres/query_runner.go
```diff
@@ -478,30 +478,41 @@ func (queryRunner *PgQueryRunner) ForEachDatabase(
...
+	}()
+
+	runner, err := NewPgQueryRunner(dbConn)
+	if err != nil {
+		return errors.Wrap(err, "Failed to build query runner")
 	}
+
+	err = function(runner, db)
```
usernamedt, line 511: return function(runner, db)

internal/databases/postgres/names_metadata_test.go
```diff
@@ -0,0 +1,81 @@
...
+	dbID, tableID, err := meta.Resolve("my_folder/my_database/my_table")
+	assert.Equal(t, uint32(0), dbID)
+	assert.Equal(t, uint32(0), tableID)
+	assert.Error(t, err)
+}
+
+func TestDatabasesByNames_ResolveDots(t *testing.T) {
+	meta := genDatabasesByNames()
+
+	dbID, tableID, err := meta.Resolve("my_database/public.my_table.my_column")
```
usernamedt, line 77: What about `public."my_table.my_column"`? Lets add a note to the documentation that some cases are not supported (with examples).

usernamedt (changes requested):

PR #1463, opened 2023-04-26
support flag latest for backup and oplog reply

Apelsin234:
support flag LATEST for
1) fetch_backup - then restore last backup
2) reply_oplog - then reply oplog from last backup to last oplog


VGoshev (approved):
Looks good but need to fix couple linter issues

VGoshev (approved):

PR #1464, opened 2023-04-26
Update golangci-lint in .github/workflows to 1.52.2

serprex:
Removed gofmt/goimports because
1. running these found files which should've been updated, but which haven't been flagged by existing CI
2. after fixing these files CI continued to fail


usernamedt (approved):
LGTM

teem0n (approved):

PR #1483, opened 2023-05-31
Show remaining upload time to user

### LINKED ISSUE #475: Show user info about how much more time do wal-g have left on the upload backup
We can calculate the average rate of uploading as uploaded_size/total_size and use it for calculating the remaining time for complete upload and print this info for user.

amagoosebitch:
### Database name
all databases, supporting uploading backups

# Pull request description
resolves #475


internal/uploader.go
```diff
@@ -216,3 +222,40 @@ func (uploader *SplitStreamUploader) Clone() Uploader {
...
+		}
+		uploadedSizeInt, err := uploader.UploadedDataSize()
+		if err != nil {
+			return
+		}
+
+		totalSize := float64(totalSizeInt) / 1000000
+		uploadedSize := float64(uploadedSizeInt) / 1000000
+
+		remainingSize := totalSize - uploadedSize
```
ostinru, line 244: It looks like you are trying to compare different sizes:
AFAIK `RawDataSize()` is uncompressed size
`UploadedDataSize()` is compressed & encrypted size that was uploaded

So, when we have compression enabled it should never meet each other and Remaining time should only grow.

Isn't it?

internal/uploader.go
```diff
@@ -32,6 +34,7 @@ type Uploader interface {
 	Clone() Uploader
```
usernamedt, line 34: What happens if we Clone the uploader? AFAIK tarSize and dataSize are shared between different uploaders. I guess that this statistics routine should also be the same for different uploaders?

internal/uploader.go
```diff
@@ -216,3 +222,40 @@ func (uploader *SplitStreamUploader) Clone() Uploader {
...
+		totalSize := float64(totalSizeInt) / 1000000
+		uploadedSize := float64(uploadedSizeInt) / 1000000
+
+		remainingSize := totalSize - uploadedSize
+
+		timeElapsed := time.Since(startTime).Seconds()
+		uploadSpeed := uploadedSize / timeElapsed
+		e.Add(uploadSpeed)
+
+		remainingTime := time.Duration(remainingSize/uploadSpeed) * time.Second
```
usernamedt, line 250: Why we can't utilize more accurate ewma that we use to calculate the remaining time?

usernamedt (changes requested):

internal/uploader.go
amagoosebitch: tried to solve this problem

ostinru: I don't think that it will work: `compressedFile` is `io.Reader`... You are trying to read-and-copy data from `io.Reader` to `bytes.Buffer{}`... and then upload.
However you can read from `io.Reader` only once... so, I think that this code will corrupt data you are trying to upload.

amagoosebitch: I'm not really comfortable about current sollution, but it is the best my brain can afford now...

serprex (approved):

usernamedt (changes requested):

internal/uploader.go
```diff
@@ -173,7 +194,20 @@ func (uploader *RegularUploader) Upload(path string, content io.Reader) error {
 	if uploader.tarSize != nil {
 		content = utility.NewWithSizeReader(content, uploader.tarSize)
 	}
+	compressedFileBytes, ioErr := io.ReadAll(content)
```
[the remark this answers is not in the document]
amagoosebitch: I can't really understand why not.
Seems like there is no other places where we read from content(io.Reader), except when we call PutObject func.

at L:207 I copy contents, and get new io.Reader 
`content = bytes.NewReader(compressedFileBytes)`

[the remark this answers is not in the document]
ostinru: It looks like you are trying to read the whole compressed file into RAM... That is likely will cause "out of memory" error. Also, I see additional memory copy from one place to another, which is bad for performance reasons.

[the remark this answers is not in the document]
ostinru: In general, it is desired to keep as little data in RAM as possible (with some buffering), that is what `wal-g` doing.
As a result, `UploadedSize - CompressedSize` will fluctuate around some value (several megabytes - the size of buffers in compressions-encryption-upload pipeline).

Keeping this in mind, I will agree with @usernamedt , that we can estimate time by estimating backup raw size - `uploader.RawDataSize`. Probably, it won't be easy to guess for all databases that wal-g supports.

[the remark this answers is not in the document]
amagoosebitch: I agree with point about RAM.

But if we want to log `Uploaded raw data size (uncompressed)` stat we also need to know what compressed data we already uploaded, isn't it?

Or at least we want to know mapping compressedSize -> decompressedSize(which also seems to use a lot of RAM).

Am I getting something wrong?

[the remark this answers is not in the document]
amagoosebitch: Ok. I Fixed previous sollution.

Current solution seems beautiful to me, this helps me get compressed size without any copying and using additional memory or time.

I hope you will also see this solution as working

internal/uploader.go
```diff
@@ -216,3 +255,41 @@ func (uploader *SplitStreamUploader) Clone() Uploader {
 		blockSize:  uploader.blockSize,
 	}
 }
+
+func (uploader *RegularUploader) ShowRemainingTime() {
+	defer uploader.waitGroup.Done()
+	startTime := time.Now()
+	for {
+		totalSizeInt, err := uploader.CompressedDataSize()
```
usernamedt, line 263: Total backup size is unknown to the uploader. New data is being constantly feed to the uploader, see the rest of the code regarding backup-push.

This is not going to show real total backup size.

usernamedt (changes requested):

internal/uploader.go
```diff
@@ -216,3 +227,32 @@ func (uploader *SplitStreamUploader) Clone() Uploader {
...
+		}
+
+		compressedSize := float64(compressedSizeInt) / 1000000
+		uncompressedSize := float64(uncompressedSizeInt) / 1000000
+
+		timeElapsed := time.Since(startTime).Seconds()
+		uploader.compressedUploadSpeeds.Add(compressedSize / timeElapsed)
+		uploader.uncompressedUploadSpeeds.Add(uncompressedSize / timeElapsed)
+
+		tracelog.InfoLogger.Printf("Uploaded compressed: %v Mb\n", compressedSize)
```
usernamedt, line 251: These 4 messages might bloat the logs a little, lets pack them into a single upload statistics message, e.g.:
`Uploaded: 124MB, average 10MB/s. Raw data read: 339MB, average 34MB/s`.

internal/uploader.go
```diff
@@ -216,3 +227,32 @@ func (uploader *SplitStreamUploader) Clone() Uploader {
...
+		compressedSizeInt, err := uploader.UploadedDataSize()
+		if err != nil {
+			return
+		}
+		uncompressedSizeInt, err := uploader.RawDataSize()
+		if err != nil {
+			return
+		}
+
+		compressedSize := float64(compressedSizeInt) / 1000000
```
usernamedt, line 244: Let's adopt some code to get a nice output for units and place it in the `util` package:
```
func ByteCountSI(b int64) string {
	const unit = 1000
	if b < unit {
		return fmt.Sprintf("%d B", b)
	}
	div, exp := int64(unit), 0
	for n := b / unit; n >= unit; n /= unit {
		div *= unit
		exp++
	}
	return fmt.Sprintf("%.1f %cB",
		float64(b)/float64(div), "kMGTPE"[exp])
}
```

usernamedt (changes requested):

PR #1489, opened 2023-06-11
Add partial restore for Greenplum

nikifkon:
### Database name
Greenplum

# Pull request description
Add `--restore-only` flag to `backup-fetch` command which behaves the same as similar in PostgreSQL

### DIFF
--- a/docker/gp_tests/scripts/configs/partial_restore_test_config.json
+++ b/docker/gp_tests/scripts/configs/partial_restore_test_config.json
@@ -1,0 +1,2 @@
+"WALE_S3_PREFIX": "s3://gppartialbucket",
+"WALG_DELTA_MAX_STEPS": "2"
--- a/docker/gp_tests/scripts/configs/partial_restore_test_config.json
+++ b/docker/gp_tests/scripts/configs/partial_restore_test_config.json
@@ -1,2 +1,0 @@
-"WALE_S3_PREFIX": "s3://gppartialbucket",
-"WALG_DELTA_MAX_STEPS": "2"
--- a/internal/databases/greenplum/extract_provider_with_spec.go
+++ b/internal/databases/greenplum/extract_provider_with_spec.go
@@ -40,10 +40,11 @@ func (m RestoreDescMaker) FromAoSegNamespace(tableName string) bool {
 
 type ExtractProviderDBSpec struct {
 	restoreParameters []string
+	restoreDescMaker  RestoreDescMaker
 }
 
 func NewExtractProviderDBSpec(restoreParameters []string) *ExtractProviderDBSpec {
-	return &ExtractProviderDBSpec{restoreParameters}
+	return &ExtractProviderDBSpec{restoreParameters, RestoreDescMaker{}}
 }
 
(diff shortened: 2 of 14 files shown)


docker/gp_tests/scripts/configs/partial_restore_test_config.json
```diff
@@ -0,0 +1,2 @@
+"WALE_S3_PREFIX": "s3://gppartialbucket",
```
usernamedt, line 1: I would add `database` so it would be easier to distinguish these two tests. Like this: `docker/gp_tests/scripts/configs/partial_database_restore_test_config.json`

usernamedt (changes requested):

internal/databases/greenplum/extract_provider_with_spec.go
```diff
@@ -0,0 +1,59 @@
...
+	return strings.HasPrefix(tableName, aoSegNamespace)
+}
+
+type ExtractProviderDBSpec struct {
+	*postgres.ExtractProviderDBSpec
+}
+
+func NewExtractProviderDBSpec(restoreParameters []string) *ExtractProviderDBSpec {
+	extractor := postgres.NewExtractProviderDBSpec(restoreParameters)
+	extractor.ExtractProviderImpl = ExtractProviderImpl{}
```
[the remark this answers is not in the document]
nikifkon: The `RestoreDescMaker` field is database-specific and only used in the `Get` method, should I also remove it from struct `ExtractProviderDBSpec`?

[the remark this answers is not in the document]
usernamedt: > The `RestoreDescMaker` field is database-specific and only used in the `Get` method, should I also remove it from struct `ExtractProviderDBSpec` like i did in [a4c2dd8](https://github.com/wal-g/wal-g/pull/1489/commits/a4c2dd8eb71baba6b09a955def7e4a06ed81ed01)?

You can make it a private field

internal/databases/greenplum/extract_provider_with_spec.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return restoredDatabases, nil
+}
+
+func (m RestoreDescMaker) FromAoSegNamespace(tableName string) bool {
+	return strings.HasPrefix(tableName, aoSegNamespace)
+}
+
+type ExtractProviderDBSpec struct {
+	*postgres.ExtractProviderDBSpec
```
usernamedt, line 42: What is the goal of including this implementation? It only stores restoreParameters.

usernamedt (changes requested):

internal/databases/greenplum/extract_provider_with_spec.go
nikifkon: It also hold `FilterFilesToUnwrap()` method that I call in overridden `Get()`. Maybe I should move `FilterFilesToUnwrap()` into RestoreDesc struct. I don't know.

usernamedt (approved):
LGTM

PR #1502, opened 2023-06-23
Support failover storages in `backup-push` and backups in `st transfer`

rdjjke:
Support failover storages in `wal-g backup-push` command. If the default storage isn't alive, the first alive failover storage is used to push the backup.

It was also necessary to support transferring backups between storages, because now they may exist not only in a single storage.
So `st transfer` command became a parent command for the following subcommands:
- `st transfer files <path-prefix>` - behaves exactly the same as `st transfer` does before.
- `st transfer pg-wals` - is a convinient alias that allows users to transfer PG WAL-files without knowing the path where they are stored.
- `st transfer backups` - moves backups in a special way that makes it impossible to restore from an inconsistent copy of a backup while it's being moved between storages. It certainly works with PostgreSQL backups, and possibly with other DBs also (but it hasn't been tested). The most changes are related to this feature.


cmd/common/st/remove.go
```diff
@@ -8,21 +8,21 @@ import (
...
 
-const deleteObjectShortDescription = "Delete the specified storage object"
+const removeShortDescription = "Removes objects by the prefix from the specified storage"
 
-// deleteObjectCmd represents the deleteObject command
-var deleteObjectCmd = &cobra.Command{
-	Use:   "rm relative_object_path",
-	Short: deleteObjectShortDescription,
+// removeCmd represents the deleteObject command
+var removeCmd = &cobra.Command{
```
rdjjke, line 14: When we support failover storages in `backup-fetch` command, we can get rid of the most part of copy-paste in `pg_full_backup_failover_storage_test` and `pg_transfer_backup_test` Docker tests.
However, in these tests, we also need a way for clearing all the storages. Before, the `delete` command was used for that in tests, and that command doesn't support failover storages. So I just supported removing multiple files by a prefix in this `st rm` command, because it already supports failover storages and it required really a few changes.

cmd/common/st/remove.go
rdjjke, line 1: I've renamed it from `delete_object` to `remove` because:
1. It deletes not only a single object anymore
2. command with name `delete` (instead of `delete_object`) already exists
3. it's executed using `st rm`, where `rm` is a short form for "remove"

cmd/common/st/transfer_backups.go
```diff
@@ -0,0 +1,46 @@
...
+
+		err = handler.Handle()
+		tracelog.ErrorLogger.FatalOnError(err)
+	},
+}
+
+var transferMaxBackups int
+
+func init() {
+	backupsCmd.Flags().IntVar(&transferMaxBackups, "max-backups", -1,
```
usernamedt, line 42: Actually, regarding the `-1` logic, is there any reason to introduce it? We can check if flag is set via `.IsSet()` method.

docker/pg_tests/scripts/tests/partial_restore_test.sh
```diff
@@ -66,6 +66,7 @@ else
 fi
 
 if psql -t -c "SELECT COUNT(*) FROM tbl;" -d third -A 2>&1 | grep -q "is not a valid data directory"; then
+  /tmp/scripts/drop_pg.sh
```
usernamedt, line 69: Let's instead do a clean-up in the common tests run pipeline? So it would guarantee that nobody will forget to add this cleanup later. And (maybe) in the separate PR?

cmd/common/st/transfer_backups.go
rdjjke: Hmm, didn't know about that. Yes, I can do it so. `--max-files` flag is currently behaving the same way

rdjjke: I just realised, that it's impossible to hide the `(default <N>)` note in the command help description, even if we call `.IsSet()` on our own. Given that we need the infinite number here (actually, max int value), it will look either weird (very long integer number) or incorrect (for example, the default value in help is `0`, but in fact, it is max int).

docker/pg_tests/scripts/tests/partial_restore_test.sh
rdjjke: I don't mind doing it in a separate PR

cmd/common/st/transfer_backups.go
rdjjke: Just made the default value = `math.MaxInt`

usernamedt (approved):

PR #1505, opened 2023-06-27
[PG] wal-g daemon \ client: ArchiveNonExistence

sergey-arefev:
# Pull request description
different approach to handle ArchiveNonExistence error
+ more detailed log in daemon mode

### DIFF
--- a/internal/databases/postgres/daemon_handler.go
+++ b/internal/databases/postgres/daemon_handler.go
@@ -25,7 +25,7 @@ type SocketWriteFailedError struct {
 }
 
 func newSocketWriteFailedError(socketError error) SocketWriteFailedError {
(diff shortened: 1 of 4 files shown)


internal/databases/postgres/daemon_handler.go
```diff
@@ -19,6 +19,10 @@ const (
 	SdNotifyWatchdog = "WATCHDOG=1"
 )
 
+var (
+	SocketWriteFailed = "socket write failed: %w"
```
usernamedt, line 23: Let's convert it to a separate error like this one https://github.com/wal-g/wal-g/blob/61e2466e2651604f5280f6742e4b29a0425aa11b/internal/fetch_helper.go#L19

usernamedt (approved):

PR #1508, opened 2023-06-28
fix SdNotify: send in goroutine

sergey-arefev:
fix issue with running wal-g daemon mode as systemd service

### DIFF
--- a/internal/databases/postgres/daemon_handler.go
+++ b/internal/databases/postgres/daemon_handler.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"path"
+	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


usernamedt (approved):

internal/databases/postgres/daemon_handler.go
```diff
@@ -172,9 +173,11 @@ func HandleDaemon(options DaemonOptions, pathToSocket string) {
 	if err != nil {
 		tracelog.ErrorLogger.Fatal("Error on listening socket:", err)
 	}
+
+	sdNotifyTicker := time.NewTicker(30 * time.Second)
+	go SendSdNotify(sdNotifyTicker.C)
```
usernamedt, line 178: defer sdNotifyTicker.Stop()

PR #1511, opened 2023-07-03
Support/improve wal g

wangyelei:
1. filter noop oplog
2. reset the Latest opTs if the file is uncompleted


x4m:
Hey @wangyelei , thanks for the PR! We would love to see your contributions. In backup systems every detail, every pair of eyes matter.
However, it woud be much easier to review your changes, if they would be split into small, logically consistent and integral parts. Can you plz solve one problem at a time? We need to discuss every change separately.
Many thanks for your interest in WAL-G.

PR #1514, opened 2023-07-13
Support failover storages in `backup-fetch`

rdjjke:
### Database name
Most of the written code is database-independent, though only PostgreSQL currently supports failover storages.

# Pull request description
It expands the `multistorage` package:
- adds such a new abstraction, as "mutli-storage folder" (`multistorage.Folder`), that implements the common `storage.Folder` interface, but also provides more specific ways to operate files in multiple storages.
- changes the implementation of the cache for the last alive storage. The previous implementation supported single storage only, and the new one supports all configured storages.

It applies these new multi-storage features in the Postgres `backup-fetch` command to support failover storages.


internal/multistorage/cache/file.go
```diff
@@ -0,0 +1,95 @@
...
+	file, err := os.OpenFile(StatusFile, os.O_RDONLY, 0666)
+	if os.IsNotExist(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("open cache file: %w", err)
+	}
+	defer func() { _ = file.Close() }()
+
+	err = unix.Flock(int(file.Fd()), unix.LOCK_SH)
```
rdjjke, line 50: In the previous implementation, the cache file wasn't locked. In some improbable coincidences, it might lead to a problem with content corruption, because there may be multiple wal-g processes writing to this file at the same time. That's why I added locking here.

To be honest, it's not too critical, because the problem mustn't exist if just a few storages are used and the whole file fits in 512 bytes or something (disk block size or the minimum writable data size declared by OS)

internal/backup_util.go
```diff
@@ -39,14 +40,15 @@ func (err NoBackupsFoundError) Error() string {
 	return fmt.Sprintf(tracelog.GetErrorFormatter(), err.error)
 }
 
-func GetLatestBackupName(folder storage.Folder) (string, error) {
+func GetLatestBackupName(folder storage.Folder) (backup string, storage string, err error) {
```
usernamedt, line 43: We can return a `Backup` struct here instead.

internal/backup.go
```diff
@@ -107,48 +108,49 @@ func UploadDto(folder storage.Folder, dto interface{}, path string) error {
 	return folder.PutObject(path, r)
 }
 
-func (backup *Backup) CheckExistence() (bool, error) {
-	exists, err := backup.SentinelExists()
+func (backup *Backup) CheckExistence() (exists bool, storage string, err error) {
+	exists, storage, err = backup.SentinelExists()
```
usernamedt, line 112: Backup is logically defined by name + location (storage). Instead of passing the backup storage name from all these functions, let's make it a field of the `Backup` struct itself.

internal/backup_selector.go
```diff
@@ -17,10 +18,13 @@ const LatestString = "LATEST"
 // BackupSelector returns the name of storage backup chosen according to the internal rules.
 // Returns NoBackupsFoundError in case there are no backups matching the criteria.
 type BackupSelector interface {
-	Select(folder storage.Folder) (string, error)
+	// TODO (rodrijjke): Fix other usages
```
usernamedt, line 21: we can switch to the `Select(folder storage.Folder) (Backup, error)` since the backup is no longer defined only by its name.

internal/multistorage/cache/alive_check.go
```diff
@@ -0,0 +1,74 @@
+package cache
```
usernamedt, line 1: Please use `git mv` next time you'll move the files to produce actual file diffs.

internal/multistorage/cache/alive_check.go
```diff
@@ -0,0 +1,74 @@
...
+			}
+			resCh <- checkRes{
+				name: s.Name,
+				err:  nil,
+			}
+		}(storage)
+	}
+
+	aliveCount := 0
+	results := make(map[string]bool, len(storages))
```
usernamedt, line 44: We need to retain the initial order of the storages, see the [current implementation](https://github.com/wal-g/wal-g/blob/e0d0d90da2b0c44b4d20bd16b30925a46665fff5/internal/multistorage/alive_checker.go#L16) for details.

internal/multistorage/folder.go
```diff
@@ -0,0 +1,458 @@
...
+	ExistsInAll(objectRelativePath string) (bool, string, error)
+
+	ReadObjectFromFirst(objectRelativePath string) (io.ReadCloser, string, error)
+	ReadObjectFoundFirst(objectRelativePath string) (io.ReadCloser, string, error)
+
+	ListFolderInFirst() (objects []storage.Object, subFolders []storage.Folder, err error)
+	ListFolderWhereFoundFirst() (objects []storage.Object, subFolders []storage.Folder, err error)
+	ListFolderAll() (objects []storage.Object, subFolders []storage.Folder, err error)
+
+	PutObjectToFirst(name string, content io.Reader) error
```
usernamedt, line 88: Instead of duplicating these functions, let's provide the desired behavior by argument. If these functions are meant to be internal, remove them from the interface.

internal/multistorage/cache/status_cache.go
```diff
@@ -0,0 +1,160 @@
...
+	err = writeFile(newFile)
+	if err != nil {
+		return nil, fmt.Errorf("write cache file: %w", err)
+	}
+
+	memCache = newFile
+	return memCache.getAllAlive(c.storages), nil
+}
+
+func (c *StatusCache) FirstAliveStorage() (*NamedFolder, error) {
```
usernamedt, line 55: We still need to retain the logic where "default" storage is chosen if it is available. See the [current implementation](https://github.com/wal-g/wal-g/blob/master/internal/multistorage/alive_checker.go#L16) for details.

internal/multistorage/cache/status_cache.go
```diff
@@ -0,0 +1,160 @@
...
+	memCacheMu.Lock()
+	defer memCacheMu.Unlock()
+
+	if memCache.isRelevant(c.ttl, c.storages...) {
+		return memCache.getAllAlive(c.storages), nil
+	}
+
+	oldFile, err := readFile()
+	if err != nil {
+		return nil, fmt.Errorf("read cache file: %w", err)
```
usernamedt, line 32: I guess that reading the cache file should not produce an error of the entire operation. Just ignore and overwrite it?

usernamedt (changes requested):

internal/multistorage/cache/alive_check.go
rdjjke: This is actually retained. In the cache (file or memory), storages are really not ordered. But they're being ordered in requesting them from the cache. For example https://github.com/wal-g/wal-g/pull/1514/files#diff-dd9521bc8256b4d4c075015b7ae04c65bd7fd7d7003b2f30a64dc89e52cdeabfR52

I did so, because the file is shared between different processes and commands, and these commands might have a different view on what storages are more priority. So now it's up to the command to decide which storages to use and how they should be ordered.

internal/multistorage/folder.go
rdjjke: I decided to remove the interface at all - it's not directly used now anyway. If someone needs these specific functions to call, they can use the struct instead.

internal/multistorage/cache/status_cache.go
rdjjke: It already works so. For the same reason as in [the comment](https://github.com/wal-g/wal-g/pull/1514#discussion_r1273831782)

I'll rename something in my code and/or add some comments to make it clearer.

internal/multistorage/cache/alive_check.go
rdjjke: As for retaining the order of failover storages (not the default one), I'll add this

internal/backup.go
```diff
@@ -108,49 +124,31 @@ func UploadDto(folder storage.Folder, dto interface{}, path string) error {
...
+		return err
 	}
 	if !exists {
-		return "", NewBackupNonExistenceError(backup.Name)
+		return NewBackupNonExistenceError(backup.Name)
 	}
-	return storageName, nil
-}
-
-func GetBackupByName(backupName, subfolder string, folder storage.Folder) (backup Backup, storageName string, err error) {
```
rdjjke: I moved this to `backup_util.go` because it's similar to other functions `GetLatestBackup` and `GetSpecificBackup` that are stored there

internal/multistorage/cache/file.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // StatusFile is stored on a disk and therefore shared between all WAL-G processes and commands.
-const StatusFile = ".walg_storage_status_cache"
+const StatusFile = "/tmp/.walg_storage_status_cache"
```
rdjjke, line 14: I see three options where to store this cache file:
1. In a user folder, as previously. @usernamedt can you please explain why you decided to put it there
2. In a working directory. I made it so initially, but in our Docker tests the working directory is "/" so the user doesn't have access to it
3. In "/tmp" directory, every user should have access to it so that the cache will be shared

We can also add a setting to the config file, because there may be a problem if two different WAL-G configs are used on the same machine but they use the same name for different failover storages

internal/multistorage/cache/file.go
```diff
@@ -0,0 +1,113 @@
...
+		return fmt.Errorf("open cache file: %w", err)
+	}
+	defer func() { _ = file.Close() }()
+
+	err = lockFile(file, true)
+	if err != nil {
+		return fmt.Errorf("acquire exclusive lock for the cache file: %w", err)
+	}
+
+	err = file.Truncate(int64(len(bytes)))
```
rdjjke, line 86: Truncate the file here, because otherwise, it will be a problem when a bigger file is overwritten by a smaller one. It usually can be truncated by `os.O_TRUNC` flag on opening, but it's not the case here - we shouldn't change the file before locking it.

TBH, this "O_TRUNC" flag may be implemented "lazy" (doesn't change the file size before actual writing), and it should be so if the file size is stored in a file descriptor, but I didn't find any proofs after a quick glance. It may also be implementation-dependent. A separate `Truncate` call must work in any case

internal/multistorage/cache/file.go
usernamedt: We've decided to stick with the per-user-folder approach. It would automatically avoid any permission errors, also it would stay consistent with the [decompressor cache file](https://github.com/wal-g/wal-g/blob/ab22a2303e5ddedeabc3706b641ec97be9e2974d/internal/fetch_helper.go#L112) logic.

usernamedt (changes requested):

internal/multistorage/cache/status_cache_test.go
```diff
@@ -0,0 +1,280 @@
...
+		require.Equal(t, "failover_2", alive[1].Name)
+	})
+
+	t.Run("update existing cache file", func(_ *testing.T) {
+		statuses, err := readFile()
+		require.NoError(t, err)
+		require.True(t, statuses.isRelevant(time.Hour, cache.storagesInOrder[2]))
+	})
+
+	invalidateInMem("default")
```
usernamedt, line 79: What happens if all storages in cache are marked as failed?

rdjjke: It would be an error. I added such test case

cmd/common/st/multistorage/failover_folder.go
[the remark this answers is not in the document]
rdjjke: Moved to a new subpackage in "internal/multistorage/"

usernamedt (approved):

teem0n (approved):

PR #1528, opened 2023-08-11
S3 priority flag

debebantur:
### Database name
postgres + s3

# Pull request description

### Describe what this PR fix
Want to add custom headers when sending request to s3

### DIFF
--- a/pkg/storages/swift/folder.go
+++ b/pkg/storages/swift/folder.go
@@ -32,7 +32,7 @@ func NewFolder(connection *swift.Connection, container swift.Container, path str
 	return &Folder{connection, container, path}
 }
 
-func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder, error) {
+func ConfigureFolder(prefix string, settings map[string]string, method ...string) (storage.Folder, error) {
 	connection := new(swift.Connection)
 	//Set settings as env variables
 	for prop, value := range settings {
--- a/tests_func/config.go
+++ b/tests_func/config.go
@@ -20,7 +20,7 @@ var Env = map[string]string{
 	"S3_BUCKET":                     "dbaas",
 	"S3_ACCESS_KEY":                 "S3_ACCESS_KEY",
 	"S3_SECRET_KEY":                 "S3_SECRET_KEY",
-	"S3_REQUEST_ADDITIONAL_HEADERS": `{"X-Yandex-Prioritypass":"ok"}`,
+	"S3_REQUEST_ADDITIONAL_HEADERS": `"wal_push":{"X-Yandex-Prioritypass":"ok"}`,
 
 	"EXPOSE_SSH_PORT": "22",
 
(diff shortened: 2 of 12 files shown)


tests_func/config.go
```diff
@@ -15,11 +15,12 @@ var Env = map[string]string{
...
-	"S3_PORT":       "9000",
-	"S3_BUCKET":     "dbaas",
-	"S3_ACCESS_KEY": "S3_ACCESS_KEY",
-	"S3_SECRET_KEY": "S3_SECRET_KEY",
+	"S3_HOST":                       "minio01",
+	"S3_PORT":                       "9000",
+	"S3_BUCKET":                     "dbaas",
+	"S3_ACCESS_KEY":                 "S3_ACCESS_KEY",
+	"S3_SECRET_KEY":                 "S3_SECRET_KEY",
+	"S3_REQUEST_ADDITIONAL_HEADERS": `{"X-Yandex-Prioritypass":"ok"}`,
```
usernamedt, line 23: We support a nested configuration in WAL-G.

So you can define these headers as YAML to avoid parsing.

```
S3_REQUEST_ADDITIONAL_HEADERS:
    - X-Yandex-Prioritypass: "ok"
    - header2: val2
```

and so on...

usernamedt (changes requested):

tests_func/config.go
usernamedt: But from the other hand, it might be difficult to set these headers from environment, maybe the current approach is also fine.

debebantur: We thought about it with @x4m  and decided that json will be ok

x4m: Well, at some point we can consider env vars old fashined way, and config file as a future.
Is it possible to make it interchangable?
so that both ways would work?
`S3_REQUEST_ADDITIONAL_HEADERS:
    - X-Yandex-Prioritypass: "ok"
    - header2: val2`
 and `S3_REQUEST_ADDITIONAL_HEADERS='{'X-Yandex-Prioritypass':'ok'}"`
and
`"S3_REQUEST_ADDITIONAL_HEADERS": '{"X-Yandex-Prioritypass":"ok"}'`

x4m (approved):

pkg/storages/swift/folder.go
```diff
@@ -32,7 +32,7 @@ func NewFolder(connection *swift.Connection, container swift.Container, path str
 	return &Folder{connection, container, path}
 }
 
-func ConfigureFolder(prefix string, settings map[string]string) (storage.Folder, error) {
+func ConfigureFolder(prefix string, settings map[string]string, method ...string) (storage.Folder, error) {
```
usernamedt, line 35: Why folder should know anything about method? We need to pass the custom behaviour in the wal_push command handler instead.

usernamedt (changes requested):

PR #1530, opened 2023-08-14
Dead lock fix

noname0443:
### Database name
MySQL, MongoDB

### Describe what this PR fix
Splitreader now doesn't stuck when get error from remote folder.

Just download internal/testutils/s3_error_folder.go, internal/stream_push_helper_test.go and launch go test

Default

### DIFF
--- a/internal/stream_push_helper_test.go
+++ b/internal/stream_push_helper_test.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/wal-g/tracelog"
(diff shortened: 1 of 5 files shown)


ostinru:
Hi, @noname0443 ! Thank you for reporting an issue and providing the patch.
I will test it and return back in couple days.

ostinru (commented):
Looks really cool!

I have left couple notes and ideas... 
And... let's roll back changes that are not related to the current change (e.g. `var` -> `:=`) - they are just noise.

internal/stream_push_helper_test.go
```diff
@@ -167,3 +168,183 @@ func TestBackup_Retry_NetworkError(t *testing.T) {
...
+				return os.RemoveAll(tmpDir)
+			}, nil
+	} else {
+		return folder, func() error {
+			return os.RemoveAll(tmpDir)
+		}, nil
+	}
+}
+
+// We have to repeat testtools code to evade import cycle
```
ostinru, line 203: I don't like copy-pasting code...
I have seen following approach to solve circular dependencies - rename the package to `internal_test` (without moving file to new directory).
Not sure that it is best practice... however it works!

ostinru (approved):
Awesome!

PR #1531, opened 2023-08-16
WAL-G storage tool copy command

reshke:
Storage tool that provides cross-storage objects copying. Also, additional argument `--prefix` allow to filter only objects with names having specific prefix in them. This comes in handy when only some folders need to be copied.

Example:
```
./main/gp/wal-g --config ./conf.yaml st copy -f ./conf.yaml -t ./conf2.yaml -p segments_005/seg1/basebackups_005/yezzey
```
Another example:
```
 ./main/gp/wal-g --config ./conf1.yaml st copy --from ./conf1.yaml --to ./conf2.yaml -d -e
```

### Database name

# Pull request description

### Describe what this PR fix
// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>


internal/splitmerge/splitmerge_test.go
```diff
@@ -34,9 +34,11 @@ func (b *BufferCloser) Close() error {
 	return nil
 }
 
-//               ┌─> copy data per 1 byte    ─>┐
+//	┌─> copy data per 1 byte    ─>┐
+//
 // data ─> split ├─> copy data per ... bytes ─>├─> merge
-//               └─> copy data per 42 bytes  ─>┘
+//
```
reshke, line 40: this is `make fmt` artefact

ostinru: Lol. I found a workaround - add emoji:
```
// 🙈               ┌─> copy data per 1 byte    ─>┐
// 🙉 data ─> split ├─> copy data per ... bytes ─>├─> merge
// 🙊               └─> copy data per 42 bytes  ─>┘
```

PS: or cdots =)
```
// ·               ┌─> copy data per 1 byte    ─>┐
// · data ─> split ├─> copy data per ... bytes ─>├─> merge
// ·               └─> copy data per 42 bytes  ─>┘
```

reshke:
Hi, i have updated this pr, this now contains logic to decrypt files from source storage and encrypt files before putting them into target folder, because source and target config may contain different encryption settings. 

Note, than compress/uncompress logic is not needed here, because data copy can be done without this!

reshke:
There is problem with `--config` file actually. You can-not run `wal-g` without specifying it, while it is not needed for this exact command.

internal/storagetools/copy_object_handler.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/wal-g/tracelog"
+	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/internal/copy"
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+
+	"github.com/wal-g/wal-g/internal/crypto"
+	"github.com/wal-g/wal-g/utility"
+)
+
+// CompressAndEncrypt compresses input to a pipe reader. Output must be used or
```
reshke, line 16: comment need to be updated. tbd

usernamedt (changes requested):

cmd/common/st/copy_object.go
```diff
@@ -0,0 +1,64 @@
...
+	fromConfigFile string
+	toConfigFile   string
+	prefix string
+
+	decryptSource bool
+	encryptTarget bool
+)
+
+// catObjectCmd represents the catObject command
+var copyObjectCmd = &cobra.Command{
```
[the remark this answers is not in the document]
reshke: yeah, done

docker/st_tests/scripts/tests/storage_tool_tests.sh
```diff
@@ -46,9 +46,48 @@ wal-g st rm testfolder/testfile.br
...
+AWS_ENDPOINT: "$AWS_ENDPOINT"
+WALE_S3_PREFIX: "s3://storagetoolsbucket_target"
+AWS_S3_FORCE_PATH_STYLE: $AWS_S3_FORCE_PATH_STYLE
+AWS_ACCESS_KEY_ID: "$AWS_ACCESS_KEY_ID"
+AWS_SECRET_ACCESS_KEY: "$AWS_SECRET_ACCESS_KEY"
+EOH
+
+wal-g --config ./conf1.yaml st copy --from ./conf1.yaml --to ./conf2.yaml
+
+# Should get empty storage after file copy
```
usernamedt, line 92:
```suggestion
# Should not have an empty storage after the file copy
```

reshke: Thanks!

usernamedt (approved):

teem0n (approved):

PR #1535, opened 2023-08-24
Now supported: additional headers for s3 requests

debebantur:
### Database name
postgres

# Pull request description
Added headers to s3 requests

### DIFF
--- a/pkg/storages/s3/session.go
+++ b/pkg/storages/s3/session.go
@@ -18,7 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/pkg/errors"
 	"github.com/wal-g/tracelog"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 )
 
 const DefaultPort = "443"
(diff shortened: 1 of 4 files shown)


pkg/storages/s3/session.go
```diff
@@ -231,9 +232,53 @@ func createSession(bucket string, settings map[string]string) (*session.Session,
...
+			for k, v := range headers {
+				request.HTTPRequest.Header.Add(k.(string), v.(string))
+			}
+		})
+	}
+
 	return s, err
 }
 
+func getHeaders(encodedHeaders string) (map[interface{}]interface{}, error) {
```
x4m, line 252: HTTP header keys and values known to be strings. I do not see a point for interface{} thing.

pkg/storages/s3/session.go
```diff
@@ -231,9 +232,53 @@ func createSession(bucket string, settings map[string]string) (*session.Session,
...
+		if !ok {
+			return nil, NewFolderError(err, "Invalid %s setting", RequestAdditionalHeaders)
+		}
+		headers = refomHeaderListToMap(headerList)
+	}
+
+	return headers, nil
+}
+
+func refomHeaderListToMap(headerList []interface{}) map[interface{}]interface{} {
```
x4m, line 271: refom?

x4m (changes requested):

pkg/storages/s3/session.go
serprex, line 271:
```suggestion
func refomHeaderListToMap(headerList []interface{}) map[string]string {
```
Might as well put the types as strings if only use is going to be casting all to string

x4m (approved):

PR #1538, opened 2023-08-25
Print storage names in `backup-list`

rdjjke:
PostgreSQL now supports failover storages which are used when the main one is dead.

In this PR I supported failover storages in `backup-list` command: storage names are now printed along with the rest of the info.

Storage names are also printed for all other databases besides PG, but only in the "basic" mode when `--detail` flag isn't set.


pkg/storages/memory/storage.go
```diff
@@ -21,8 +21,10 @@ type TimeStampedData struct {
 	Size      int
 }
 
+var NewObjectTime = time.Now
```
usernamedt, line 24: I guess we want to store the time the object has been put into memory storage? Anyway, I don't quite get what has been changed here.

pkg/storages/memory/folder.go
```diff
@@ -19,6 +19,7 @@ type Folder struct {
 }
 
 func NewFolder(path string, storage *Storage) *Folder {
+	path = strings.TrimPrefix(path, "/")
```
usernamedt, line 22: Why do we trim it?

internal/multistorage/folder.go
```diff
@@ -137,9 +137,13 @@ func (mf Folder) GetPath() string {
 
 // GetSubFolder provides a multi-storage subfolder, which includes subfolders of all used storages.
 func (mf Folder) GetSubFolder(subFolderRelativePath string) storage.Folder {
+	newPath := path.Join(mf.path, subFolderRelativePath) + "/"
```
usernamedt, line 140: We have `storage.AddDelimiterToPath / storage.JoinPath` functions, let's re-use them?

internal/databases/mongo/models/backup.go
```diff
@@ -33,6 +38,83 @@ func (b *Backup) IsPermanent() bool {
...
+		},
+		{
+			Name:       "hostname",
+			PrettyName: "Hostname",
+			Value:      b.Hostname,
+		},
+		{
+			Name:       "start_ts",
+			PrettyName: "Start Ts",
+			Value:      b.MongoMeta.Before.LastMajTS.String(),
```
usernamedt, line 80: `b.MongoMeta.Before.LastMajTS.ToBsonTS()`

usernamedt (changes requested):

pkg/storages/memory/storage.go
rdjjke: This is for tests. Before, objects stored in this memory received the `time.Now()` timestamp. But now it's possible to override this `NewObjectTime` func to make objects' timestamps static and be able to test it.

pkg/storages/memory/folder.go
rdjjke: This memory folder in its implementation of `ListFolder` func decides that a file is stored inside a folder if their prefixes are the same. However, file paths never start with "/", but the folder one do. So it doesn't assign any files to such folder. The same problem may be with other funcs too. 

I suppose it would be great to remove the ambigutiy between "/a/b/c" and "a/b/c" paths (as well as between "/" and "") in all our `storage.Folder` implimentations, because there may be quite a few bugs related to that. But this is too much to do it right here. So I fixed only the problem I found with my tests.

internal/multistorage/folder.go
rdjjke: OK, we can use `storage.AddDelimiterToPath(storage.JoinPath(mf.path, subFolderRelativePath)))` in this specific line and it will work correctly.

However, we can't drop the 3 lines below, because the functions you mentioned return "/" instead of "" in case of the root folder (other folder paths don't start with "/", though). This can be kind of unsafe to create a multistorage folder with this path: in the other comment I mentioned a problem with it in `memory.Folder` and it can be not alone. 

I was afraid to change logic of the functions you mentioned because it can potentially break something that isn't covered by tests now.

internal/databases/mongo/models/backup.go
rdjjke: I was just confused how it should be printed if `bson/primitive/Timestamp` stuct doesn't implement `String()`. Just with the standard Go "%v" struct formatter?

internal/multistorage/folder.go
rdjjke: Did you mean to replace this line only, or to drop the lines below also?

rdjjke: Made it so. Please, say, if I shouldn't have done this

internal/databases/mongo/models/backup.go
rdjjke: Did it

pkg/storages/memory/storage.go
usernamedt: Then it would be better to pass it through the storage constructor `NewStorage()`. Because now this logic is not straightforward.

rdjjke: I though it's a common practice to mock time this way, but I can pass it through the contructor, OK

usernamedt (approved):

teem0n (approved):

KhurtinDN (approved):

PR #1539, opened 2023-08-28
Enhancement: Read file content from STDIN for wal-g st put.

reshke:
### Database name
All, wal-g st patch.

# Pull request description

Add Feature to wal-g st put to read file content from STDIN if an additional flag was specified. Number of possible args for this command was changes, so additional validation code added.

### DIFF
--- a/cmd/common/st/put_object.go
+++ b/cmd/common/st/put_object.go
@@ -3,7 +3,6 @@ package st
 import (
 	"io"
 	"os"
-	"fmt"
 	
 	"github.com/spf13/cobra"
 	"github.com/wal-g/tracelog"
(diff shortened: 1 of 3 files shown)


cmd/common/st/put_object.go
```diff
@@ -16,23 +20,66 @@ const (
 	noCompressFlag     = "no-compress"
 	overwriteFlag      = "force"
 	overwriteShorthand = "f"
+	readStdinFlag 	   = "read-stdin"
+	readStdinShorthand = "s"
 )
 
+
+func openLocalFile(localPath string) (io.ReadCloser, error) {
```
usernamedt, line 28: We should do only command args handling here. Let's keep the actual logic inside the handler?

usernamedt (changes requested):

cmd/common/st/put_object.go
reshke: Sure, no problem, fixed

usernamedt (approved):

PR #1546, opened 2023-09-04
Pass Context around Folder upload functions, add PG WAL push timeout

ne2pit:
### Database name

Timeout is added to Postgres WAL push. Changes in Upload is common for all DBs.

# Pull request description

- Folder interface is extended with PutObjectWithContext method for gradual migration.
- Folder implementations supports cancelation by context for new method.
- Context passing added to some Upload logic.
- Timeout is added to PG WAL Push handle.

# Side changes

- github.com/ncw/swift bumped to v2 major.
- Swift storage integration docker test
- SSH storage integration docker test


pkg/storages/sh/folder.go
```diff
@@ -268,6 +270,11 @@ func (folder *Folder) PutObject(name string, content io.Reader) error {
 	return nil
 }
 
+func (folder *Folder) PutObjectWithContext(ctx context.Context, name string, content io.Reader) error {
+	ctxReader := contextio.NewReader(ctx, content)
+	return folder.PutObject(name, ctxReader)
```
serprex, line 275: Seems like contextio could be skipped in favor of adding a context parameter to folder.PutObject

internal/databases/postgres/bguploader.go
```diff
@@ -189,7 +189,8 @@ func (b *BgUploader) processFiles(fileChan <-chan string) {
 		b.started[name] = struct{}{}
 		if err := b.workerCountSem.Acquire(b.ctx, 1); err == nil {
 			go func() {
-				uploadedFile := b.upload(name)
+				// todo Should it use b.ctx? It will interrupt last file upload and change behavior.
```
usernamedt, line 192: I guess we should not use b.ctx here. The default behavior is to wait for all uploads to finish.

pkg/storages/swift/folder.go
```diff
@@ -151,6 +153,11 @@ func (folder *Folder) PutObject(name string, content io.Reader) error {
 	return nil
 }
 
+func (folder *Folder) PutObjectWithContext(ctx context.Context, name string, content io.Reader) error {
```
usernamedt, line 156: `ObjectPut` actually [supports](https://github.com/ncw/swift/blob/b37a86bc3491c732a2b5ea198bc7cb307e239992/swift.go#L1597C4-L1597C4) passing the context, maybe just bump the swift library version and use it?

pkg/storages/memory/mock/folder.go
```diff
@@ -75,6 +76,13 @@ func (f *Folder) PutObject(name string, content io.Reader) error {
 	return f.MemFolder.PutObject(name, content)
 }
 
+func (f *Folder) PutObjectWithContext(ctx context.Context, name string, content io.Reader) error {
+	if err := ctx.Err(); err != nil {
```
usernamedt, line 80: Why not reuse the `contextio.NewReader` here?

usernamedt (changes requested):

usernamedt:
I also suggest to add the new setting to the WAL-G daemon [docs](https://github.com/wal-g/wal-g/blob/master/docs/PostgreSQL.md#daemon).

pkg/storages/memory/mock/folder.go
ne2pit: It is a performance micro optimisation. Memory folder works fast. So I decided to skip context checking on every Read call. Checking context on start should be enough to detect most of cancelling cases.

usernamedt: > It is a performance micro optimisation. Memory folder works fast. So I decided to skip context checking on every Read call. Checking context on start should be enough to detect most of cancelling cases.

Since it is being used only for tests, I suggest sticking with the approach that is being used on other storages to catch possible bugs in the future.

internal/databases/postgres/daemon_handler.go
```diff
@@ -60,15 +61,21 @@ type ArchiveMessageHandler struct {
...
+func (h *ArchiveMessageHandler) Handle(ctx context.Context, messageBody []byte) error {
 	tracelog.DebugLogger.Printf("wal file name: %s\n", string(messageBody))
 
 	fullPath, err := getFullPath(path.Join("pg_wal", string(messageBody)))
 	if err != nil {
 		return err
 	}
 	tracelog.DebugLogger.Printf("starting wal-push: %s\n", fullPath)
-	err = HandleWALPush(h.uploader, fullPath)
+	pushTimeout, err := internal.GetDurationSetting(internal.PgDaemonWALUploadTimeout)
```
usernamedt, line 72: I suggest moving this to the ArchiveMessageHandler constructor.

pkg/storages/gcs/folder.go
```diff
@@ -282,12 +282,16 @@ func (folder *Folder) ReadObject(objectRelativePath string) (io.ReadCloser, erro
...
-	object := folder.BuildObjectHandle(folder.joinPath(folder.path, name))
-
 	ctx, cancel := folder.createTimeoutContext()
 	defer cancel()
 
+	return folder.PutObjectWithContext(ctx, name, content)
+}
+
+func (folder *Folder) PutObjectWithContext(ctx context.Context, name string, content io.Reader) error {
+	tracelog.DebugLogger.Printf("Put %v into %v\n", name, folder.path)
```
usernamedt, line 292: Actually, I found a regression here: now we won't use the timeout context as we used to do. I suggest passing the ctx to the `folder.createTimeoutContext()`.

ne2pit:
todo: Add Swift and SSH docker tests to `make pg_integration_test`. Something like:
```bash
	@if [ "$(TEST)" = "pg_tests" ]; then\
		docker-compose build ssh &&\
		docker-compose up --exit-code-from pg_storage_swift_test pg_storage_swift_test &&\
		docker-compose up --exit-code-from pg_storage_ssh_test pg_storage_ssh_test ;\
	fi
```
Similar changes are made in #1551 . Doing so in parallel will lead to merge conflicts. It is possible to add it to one of two PRs that will be merged second.

usernamedt (approved):

teem0n (approved):

KhurtinDN (approved):

PR #1547, opened 2023-09-05
Feature/envelope pgp

Alviner:
### Database name
Enveloped gpg via yckms was added to WAL mondo/pg/sqlserver

# Pull request description
Use Envelope encryption on gpg keys to stop storing the original gpg (dek key) in the system and make possible a painless gpg key rotation.

TBD:

- [x] clean configure to prevent setting both gpg and encrypted gpg

- [x] make ttl cache on yckms enveloper
- [x] make stale cache on yckms envelope
- [x] docs
- [x] ~~try to setup yckms mock) it use~~ aws local kms use http(


internal/crypto/envelope/enveloper.go
```diff
@@ -0,0 +1,11 @@
+package envelope
+
+import "io"
+
+//go:generate mockery --name Enveloper --with-expecter=true
```
Alviner, line 5: There is no standart mock gen in project.
Should we make an issue in backlog for that?

internal/configure.go
```diff
@@ -314,19 +358,27 @@ func ConfigureCrypterForSpecificConfig(config *viper.Viper) crypto.Crypter {
...
+func configureEncryptedPgpCrypter(config *viper.Viper) crypto.Crypter {
+	if !config.IsSet(YcKmsKeyIDSetting) {
+		tracelog.ErrorLogger.Fatalf("Yandex Cloud KMS key for client-side encryption and decryption must be configured")
 	}
 
-	if config.IsSet(YcKmsKeyIDSetting) {
-		return yckms.YcCrypterFromKeyIDAndCredential(config.GetString(YcKmsKeyIDSetting), config.GetString(YcSaKeyFileSetting))
-	}
+	yckmsEnveloper := yckmsenvlpr.EnveloperFromKeyIDAndCredential(
+		config.GetString(YcKmsKeyIDSetting), config.GetString(YcSaKeyFileSetting),
```
Alviner, line 370: maybe it's worth having separate yckms config for encrypted gpg to avoid misconfiguration..

internal/crypto/envelope/enveloper.go
usernamedt: Yes, current zoo of mock generators is confusing and needs to be fixed eventually. Would appreciate an issue addressing this.

internal/crypto/envelope/enveloper.go
```diff
@@ -0,0 +1,11 @@
...
+type Enveloper interface {
+	GetName() string
```
usernamedt, line 7:
```
	Name() string
	ReadEncryptedKey(r io.Reader) ([]byte, error)
	DecryptKey([]byte) ([]byte, error)
	SerializeEncryptedKey(encryptedKey []byte) []byte
```

internal/crypto/envelope/enveloper/yckms/enveloper.go
```diff
@@ -0,0 +1,119 @@
...
+	_, err = r.Read(encryptedKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return encryptedKey, nil
+}
+
+func EnveloperFromKeyIDAndCredential(keyID string, saFilePath string) envelope.Enveloper {
+	authorizedKey, err := iamkey.ReadFromJSONFile(saFilePath)
```
usernamedt, line 106: We need to return error here, and throw fatal only on the high-level.

usernamedt (changes requested):

internal/crypto/envelope/enveloper/cached/enveloper.go
```diff
@@ -0,0 +1,86 @@
...
+	"sync"
+	"time"
+
+	"github.com/wal-g/tracelog"
+	"github.com/wal-g/wal-g/internal/crypto/envelope"
+)
+
+type Item struct {
+	Object    []byte
+	ExpiredAt int64
```
usernamedt, line 16: Why not store it as a `time.Time`?

internal/crypto/envelope/enveloper/cached/enveloper.go
```diff
@@ -0,0 +1,86 @@
...
+			return item.Object, nil
+		}
+		return nil, err
+	}
+
+	enveloper.locker.Lock()
+	defer enveloper.locker.Unlock()
+
+	var expiredAt int64
+	if enveloper.expiration > 0 {
```
usernamedt, line 66: Why do we need this if? If expiration is zero it would work anyway.

internal/crypto/envelope/enveloper/yckms/enveloper.go
```diff
@@ -0,0 +1,124 @@
...
+	keyID string
+	sdk   *ycsdk.SDK
+}
+
+func (enveloper *Enveloper) Name() string {
+	return "yckms"
+}
+
+func (enveloper *Enveloper) ReadEncryptedKey(r io.Reader) ([]byte, error) {
+	encryptedKey, err := readEncryptedKey(r)
```
usernamedt, line 31: `return readEncryptedKey(r)`

usernamedt (changes requested):

internal/crypto/envelope/enveloper/cached/enveloper.go
Alviner: i want to permanent store in cache with zero expiration, if there is more clearly way to do it?)

usernamedt (approved):

PR #1549, opened 2023-09-05
Improve storage fail detection by putting the object

ne2pit:
Sometimes storage might be alive but work very slow or go to read-only mode. Current fail detector uses ListObject method to check it's liveness. Such check is unreliable.

This PR changes check method to PutObject. Data is pseudo random generated. The object size and the timeout is configurable. Default values are 1MiB and 30s. The object has predefined name always. So there is no additional storage space waste on sequential checks.


usernamedt (approved):
LGTM, let's wait for the context PR to be merged. And the last thing - add a note to the docs about the new setting.

internal/multistorage/cache/alive_check.go
```diff
@@ -52,11 +56,13 @@ func checkForAlive(timeout time.Duration, storages ...NamedFolder) map[key]bool
...
-func checkStorage(ctx context.Context, folder NamedFolder) error {
-	// Currently, we use the simple ListFolder() call to check if the storage is up and reachable
+func checkStorage(ctx context.Context, size uint, folder NamedFolder) error {
 	errCh := make(chan error, 1)
+	// todo simplify code using PutObjectWithContext after merge https://github.com/wal-g/wal-g/pull/1546
 	go func() {
-		_, _, err := folder.ListFolder()
+		r := rand.New(rand.NewSource(1))
+		lr := io.LimitReader(r, int64(size))
+		err := folder.PutObject(checkObjectName, lr)
```
rdjjke, line 65: Don't we want to check storages by both putting an object and listing the folder?

ne2pit:
Returned ListFolder check as we discussed. And now our assumption to mix it with context PR become different. Adding ListFolderWithContext is trivial for most of storages except SSH. It is impossible to make it with current SFTP library on the first sight. What do you think? Should we still wait for context PR merge?

usernamedt:
> Returned ListFolder check as we discussed. And now our assumption to mix it with context PR become different. Adding ListFolderWithContext is trivial for most of storages except SSH. It is impossible to make it with current SFTP library on the first sight. What do you think? Should we still wait for context PR merge?

We still have the `WALG_FAILOVER_STORAGES_CHECK_TIMEOUT` so I guess it is ok to simply call the ListObjects().

PR #1550, opened 2023-09-06
Allow to specify storage names in PG commands

rdjjke:
This is one of the changes directed to support failover storages in all PostgreSQL commands. Such PG commands as `backup-push`, `backup-fetch`, `wal-push` and some others already support failover storages that are used in case the primary one is dead. Currently, all storages specified in the config file can be used.

In this PR I'm introducing an additional `--target-storage` flag to all commands that already support failover storages. This allows us to specify an exact storage to execute the command against, instead of using all configured storages. This feature may be useful for debugging purposes or in case of repairing something broken.

### DIFF
--- a/internal/databases/postgres/configure.go
+++ b/internal/databases/postgres/configure.go
@@ -1,6 +1,7 @@
 package postgres
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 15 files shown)


cmd/daemonclient/main.go
```diff
@@ -54,11 +54,11 @@ var (
 	commands = map[string]*commandOpts{
 		"wal-push": {
```
usernamedt, line 55: I suggest adding the config setting for controlling this behavior. And for daemon mode, it might be the only option since it breaks backward compatibility if we change the protocol like this. I don't see the use case where we need to decide it for each WAL file anyway.

usernamedt (changes requested):

cmd/daemonclient/main.go
rdjjke: If this setting is set via env variable, will it be merged with the rest of the config? If so, it could be a nice option

internal/databases/postgres/configure.go
```diff
@@ -1,15 +1,49 @@
...
 	"github.com/wal-g/wal-g/internal/fsutil"
+	"github.com/wal-g/wal-g/internal/multistorage"
+	"github.com/wal-g/wal-g/internal/multistorage/cache"
+	"github.com/wal-g/wal-g/pkg/storages/storage"
 )
 
+// ConfigureMultiStorageFolder configures the primary storage folder and all the failover ones, if any, and builds a
+// multi-storage folder that aggregates them. It also sets up a cache to keep storage alive check results there.
+// This function doesn't set any specific multi-storage folder policies, so policies.Default are used initially.
+func ConfigureMultiStorageFolder() (storage.Folder, error) {
```
rdjjke, line 20: I moved this function as well as [`PrepareMultiStorageWalUploader`](https://github.com/wal-g/wal-g/pull/1550/files#diff-b9064099fa73c81cd39e3b8987ce68bad01730259e32d36c9100ae43e81bd5a9R75) and [`PrepareMultiStorageFolderReader`](https://github.com/wal-g/wal-g/pull/1550/files#diff-8bbbe6ade21b38b228c1d7d55c1f661da67bb808ff3fa119763e0a291ef038b6R28) from `cmd` package to here to allow the daemon to use them in its handlers (it would be bad if the daemon handlers depended on `cmd`).

But after the changes requested by @usernamedt, these functions aren't used in daemon handlers anymore.

However, I can leave them here anyway, because we may need it in the daemon again in the future, maybe for some commands other than `wal-fetch` and `wal-push`.

usernamedt (approved):

internal/databases/postgres/daemon_handler.go
```diff
@@ -99,13 +101,17 @@ func (h *WalFetchMessageHandler) Handle(_ context.Context, messageBody []byte) e
 	if len(args) != 2 {
 		return fmt.Errorf("wal-fetch incorrect arguments count")
 	}
-	fullPath, err := getFullPath(args[1])
+	var (
```
usernamedt, line 104:
```
walFileName := args[0]
location := args[1]
```

why not using the short vars declaration?

PR #1551, opened 2023-09-07
PG Integration tests and linter local dev improvements

ne2pit:
PG Integration tests improvements:
- Add compiled binary to .dockerignore. Content of the binary is different always because of timestamp imprint. This fixes docker build cache.
- Fixed typo in `pg_remote_backup_without_files_metadata_test`. Add to CI.
- Improved cleanup. Stop docker compose after test run. Delete more docker artefacts on `make pg_clean`.
- Removed one more redundant container image. All tests can be run in image build from `Dockerfile_prefix`. Continue optimisation from #1510.
- Tests `pg_pgbackrest_backup_fetch_test` and `pg_ssh_backup_test` are excluded from running in one container because of special requirements. Makefile runs them in another docker-compose calls.
- Add cleanup of PGDATA_ALPHA and PGDATA_BETA for a couple of tests.
- Decreased time of pgbench working in `pg_pgbackrest_backup_fetch_test` from 1000 to 60 seconds. I don't know much about test logic. But anyway running a data creation in test for 16 minutes looks overkill.

After improvements all tests successfully runs on notebook. It took 30 minutes with Colima as a Docker runtime. Configuration is 2 CPU / 8 GB Memory.
```
make pg_integration_test  8.46s user 8.79s system 0% cpu 30:00.78 total
```

Linter improvements:
- Added notice about golangci-lint is not recommended to build from source.
- Fixed version of golangci-lint container. Now it is matched with GH Actions version.
- Added cache docker volume. Second run is less than minute against 6-7 minutes without cache.


ne2pit:
Moving golangci-lint setting `modules-download-mode: vendor` to correct place brake linter on CI. There is no vendor in repo actually. Change setting value to `readonly` as recommended for CI.

usernamedt (approved):

PR #1553, opened 2023-09-08
fix: use yc endpoint and optional file creds with envelope gpg

Alviner:
* use metadata server if no sa file was passed
* pass yc endpoint to use in different environments

### DIFF
--- a/internal/crypto/envelope/enveloper/yckms/enveloper.go
+++ b/internal/crypto/envelope/enveloper/yckms/enveloper.go
@@ -114,7 +114,7 @@ func getCredentials(saFilePath string) (ycsdk.Credentials, error) {
 	return credentials, nil
 }
 
-func EnveloperFromKeyIDAndCredential(keyID string, saFilePath string, endpoint string) (envelope.Enveloper, error) {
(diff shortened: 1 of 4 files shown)


internal/crypto/envelope/enveloper/yckms/enveloper.go
```diff
@@ -97,18 +97,32 @@ func readEncryptedKey(r io.Reader) ([]byte, error) {
...
+		if credErr != nil {
+			return nil, errors.Wrap(credErr, "Can't initialize yc sdk")
+		}
+		credentials = accountCredentials
 	}
-	var credentials, credErr = ycsdk.ServiceAccountKey(authorizedKey)
+	return credentials, nil
+}
+
+func EnveloperFromKeyIDAndCredential(keyID string, saFilePath string, endpoint string) (envelope.Enveloper, error) {
```
usernamedt, line 117:
```suggestion
func EnveloperFromKeyIDAndCredential(keyID, saFilePath, endpoint string) (envelope.Enveloper, error) {
```

usernamedt (changes requested):

usernamedt (approved):

PR #1556, opened 2023-09-15
fix: b64 encoded envelope key length

Alviner:
DecodedLen returns the maximum length in bytes of the decoded data, not the real


usernamedt:
Ok, but let's add this comment to the code.

usernamedt (approved):

PR #1559, opened 2023-09-18
feat: add in header key id information

Alviner:
feat key id in encrypted file headers to clearly identify which key was used.
also add debug info with found key id

### DIFF
--- a/internal/crypto/envelope/enveloper/yckms/enveloper.go
+++ b/internal/crypto/envelope/enveloper/yckms/enveloper.go
@@ -10,6 +10,7 @@ import (
 	ycsdk "github.com/yandex-cloud/go-sdk"
 	"github.com/yandex-cloud/go-sdk/iamkey"
 
+	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal/crypto/envelope"
 )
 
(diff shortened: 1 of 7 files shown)


internal/crypto/envelope/openpgp/crypter.go
```diff
@@ -47,10 +49,25 @@ func (crypter *Crypter) Encrypt(writer io.Writer) (io.WriteCloser, error) {
 	if err != nil {
 		return nil, err
 	}
+	var key []byte
```
usernamedt, line 52: you can use the short variable declaration here:
```
	key, err := crypter.enveloper.DecryptKey(crypter.encryptedKey)
```

internal/crypto/envelope/enveloper/yckms/enveloper.go
```diff
@@ -46,23 +47,30 @@ func (enveloper *Enveloper) DecryptKey(encryptedKey []byte) ([]byte, error) {
...
 		scheme version (current version is 1)
+		uint32 - keyID len
+		keyID ...
 		uint32 - encrypted key len
 		encrypted key ...
 	*/
 
+	result := append([]byte(magic), schemeVersion)
+
+	keyIDLen := make([]byte, 4)
```
usernamedt, line 66: Let's also extract this size to const like it is done [in other places](https://github.com/wal-g/wal-g/blob/953036f7482b0f33c7602aebacd9d91d0476ebf9/internal/databases/postgres/pagefile.go#L34) to get rid of the magic constants

usernamedt (changes requested):

usernamedt (approved):

PR #1560, opened 2023-09-18
Support failover storages in backup deletion

rdjjke:
This PR introduces support of failover storages in all variations of `wal-g delete` command for PostgreSQL.


internal/delete_handler.go
```diff
@@ -448,10 +453,40 @@ func (h *DeleteHandler) findDependantBackups(target BackupObject) []BackupObject
...
 		curr, queue = queue[0], queue[1:]
-		relatedBackups = append(relatedBackups, curr)
+		dependantBackups = append(dependantBackups, curr)
 		queue = append(queue, incrementsByBackup[curr.GetBackupName()]...)
 	}
-	return relatedBackups
+	return dependantBackups
+}
+
+func DeleteObjectsWhere(
```
rdjjke, line 462: Moved this function here from `storage/folder.go` to make it possible to log storage names inside. It wasn't possible to reference `multistorage` package from `storage` one because of a cyclic dependency.

internal/delete_handler_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/wal-g/wal-g/pkg/storages/memory"
+	"github.com/wal-g/wal-g/pkg/storages/storage"
+)
+
+func CreateMockStorageFolder() storage.Folder {
```
rdjjke, line 13: Just moved corresponding tests along with the tested function `DeleteObjectsWhere`

internal/delete_handler.go
```diff
@@ -486,28 +521,28 @@ func getBeforeChoiceFunc(name string, modifier int) func(object BackupObject) bo
 }
 
 func getRetainChoiceFunc(retentionCount, modifier int) func(object BackupObject) bool {
-	count := 0
+	uniqueNames := map[string]bool{}
```
rdjjke, line 524: In multi-storage world, backup name doesn't reference to a single backup anymore as multiple backups with the same name may exist in different storages. 

When we call `delete retain N` command, we expect to retain N unique backups without duplicates, so now we count them using a map.

internal/multistorage/folder.go
```diff
@@ -98,9 +98,13 @@ func EnsureSingleStorageIsUsed(folder storage.Folder) error {
 func changeDirectory(path string, storages ...cache.NamedFolder) []cache.NamedFolder {
 	newStorages := make([]cache.NamedFolder, len(storages))
 	for i := range storages {
+		folder := storages[i].Folder
```
rdjjke, line 101: There is a bug in some `storage.Folder` implementations (at least the s3 one): they work bad when calling `GetSubFolder` with the empty path "". 

In such case, the underlying path "" becomes "/", and the subsequent subfolders are appended to this "/" at the beginning making the underlying folder's path look like "/root/aaa/bbb". This breaks file listing and maybe something else. The problem is that listing is performed by matching the common suffixes between file and folder. And if folder path starts with "/", no files are matched to this folder.

I encountered this problem multiple times already with different storage implementations, because in this `multistorage.Folder` implementation it's very common to call `GetSubFolder("")`. So now I decided just keep `folder` variable the same if `path` is empty.

internal/multistorage/object.go
```diff
@@ -5,16 +5,38 @@ import (
...
+func (mo multiObject) GetStorage() string {
+	return mo.StorageName
+}
+
+// GetStorage provides the name of the storage where the object is stored. If the object can't tell the storage name on
+// its own, provides "default".
 func GetStorage(obj storage.Object) string {
-	if mo, ok := obj.(multiObject); ok {
-		return mo.storageName
+	// TODO: Don't rely on different storage.Object implementations here (issue #TODO). Possible solutions:
```
rdjjke, line 30: @usernamedt  Is github issues a suitable place to put such refactoring task?

pkg/storages/storage/folder.go
```diff
@@ -99,11 +84,14 @@ func ListFolderRecursivelyWithFilter(folder Folder, folderSelector func(path str
 	return relativePathObjects, nil
 }
 
-func addPrefixToNames(objects []Object, folderPrefix string) []Object {
+func makePathsRelative(objects []Object, folderPrefix string) []Object {
```
rdjjke, line 87: This function is called from different `ListFolderRecursively*` methods. And this functions replaces names of the received objects to the relative paths, which may be longer than just names.

This doesn't work with multistorage folders, because when we create a new instance of `storage.LocalObject` struct, we loose the information about the storage where this object is stored.

So here I introduced a special `RelativePathObject` struct, that implements both `storage.Folder` and `multistorage.MultiObject` interfaces and just wraps any other `storage.Folder` calculating the relative path on flight. 

I also don't think that storing paths in `storage.Object.GetName()` is a good idea because it's always not clear from the code what is stored in some specific object - just the name or the path. Maybe we need to rework it somehow. Now I introduced at least a specific implementation for that.

internal/backup_selector.go
```diff
@@ -93,19 +93,24 @@ func (s UserDataBackupSelector) findBackupByUserData(userData interface{}, folde
...
+	StorageName string
+}
+
 // Search backups in storage using specified criteria
 func searchInMetadata(
 	criteria func(GenericMetadata) bool,
 	folder storage.Folder,
 	metaFetcher GenericMetaFetcher,
-) (metaByStorage map[string]GenericMetadata, err error) {
+) ([]GenericMetadataInStorage, error) {
```
rdjjke, line 106: I found a bug that I made in some previous PR. This function could provide only a single backup metadata per storage, because I used `map` here.

This led to that any backup selection by metadata worked incorrectly - it couldn't find the desired backup even if it existed. And found oldest backups could not be actually the oldest ones.

cmd/mysql/delete.go
```diff
@@ -54,7 +54,7 @@ var deleteTargetCmd = &cobra.Command{
 
 type DeleteHandler struct {
 	*internal.DeleteHandler
-	permanentObjects map[string]bool
+	permanentObjects []string
```
teem0n, line 57: Why are you changing the map here for slice?
Map was used for search, didn't it?

rdjjke: The map wasn't used for search here. It's only provided to `internal.DeleteHandler.HandleDeleteEverything`. And that function uses this only to log backup names and nothing else. The reason why it now accepts slice instead of map is that for DBs that support multiple storages, backup name is not a unique identifier anymore, because backups with the same name can be stored in different storages. I could actually make the key type of this map a special struct like `{backupName, storageName}`, but I just noticed that map isn't really necessary here and slice is also OK, so I made it slice.

docker/pg_tests/scripts/tests/delete_before_time_find_full_test.sh
```diff
@@ -21,34 +21,42 @@ echo "archive_timeout = 600" >> /var/lib/postgresql/10/main/postgresql.conf
 
 wal-g --config=${TMP_CONFIG} delete everything FORCE --confirm
 
-for i in 1 2 3 4
```
usernamedt, line 24: I think there is no need to alter the existing tests. We should not break the current `wal-g delete` behavior for users, and these tests should prove it. Create a separate test(s) for failover scenarios.

usernamedt (changes requested):

internal/multistorage/object.go
usernamedt: What if we just switch to the `multistorage.Object` in our delete handler? To avoid all this logic with guessing object types

usernamedt (changes requested):

docker/pg_tests/scripts/tests/delete_before_time_find_full_test.sh
rdjjke: Took all tests back, and made only some of them check the multi-storage functionality.

internal/multistorage/recursive_list.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, fmt.Errorf("can't list folder %q: %w", prefix, err)
+	}
+	return prependPaths(objects, prefix), nil
+}
+
+func prependPaths(objects []storage.Object, folderPrefix string) []storage.Object {
+	relativePathObjects := make([]storage.Object, len(objects))
+	for i, object := range objects {
+		relativePathObjects[i] = multiObject{
+			Object: storage.NewDeepLocalObject(
```
usernamedt, line 74: Do we actually need a `DeepLocalObject`? We don't rely on it anywhere. And we can use the `LocalObject` with the same logic.

usernamedt (changes requested):

internal/multistorage/recursive_list.go
rdjjke: Just introduced it to have a way to distinguish between objects, which are created in storage implementations and contain file names only, and this derivative ones containing paths

usernamedt (approved):

teem0n (approved):

PR #1565, opened 2023-09-22
Don't require user $HOME dir to exist

rdjjke:
Previously, user $HOME dir was required to present and be accessible for failover storages functionallity. Only PostgreSQL actually used failover storages, but because of the fact that $HOME dir was checked in `init()` func of `multistorage` package, every DB executed this.

In this PR, I took out this code from the `init()` func.

### DIFF
--- a/internal/multistorage/cache/file.go
+++ b/internal/multistorage/cache/file.go
@@ -13,15 +13,13 @@ import (
 )
 
 // StatusFile is stored on a disk and therefore shared between all WAL-G processes and commands.
-var StatusFile string
-
-func init() {
+var StatusFile = func() (string, error) {
 	homeDir, err := os.UserHomeDir()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


usernamedt (approved):

internal/multistorage/cache/file.go
```diff
@@ -13,15 +13,13 @@ import (
 )
 
 // StatusFile is stored on a disk and therefore shared between all WAL-G processes and commands.
-var StatusFile string
-
-func init() {
+var StatusFile = func() (string, error) {
 	homeDir, err := os.UserHomeDir()
```
usernamedt, line 17: Let's just move this code [here](https://github.com/wal-g/wal-g/blob/cc850987d1e46cdddb088577744b448e82bb6f5d/internal/multistorage/cache/status_cache.go#L31)?

usernamedt (changes requested):

PR #1570, opened 2023-09-29
Log storage names

rdjjke:
This PR just improves observability a little in case failover storages are used.
Added printing storage names that are used in commands, and also storage status cache hits/misses.


internal/multistorage/cache/status_cache.go
rdjjke, line 1: Here in StatusCache, I added logging of:
- what storages are taken
- are they alive or not
- were they taken from the cache, or we performed a new alive check (One exception here: we don't log if the requested storage(s) is alive and taken from **memory** cache. Just not to print too many logs when the same process calls these functions multiple times)

KhurtinDN (approved):

PR #1579, opened 2023-10-18
Introduce a setting to turn off storage alive checks

rdjjke:
In this PR I added the `WALG_FAILOVER_STORAGES_CHECK` flag to the config, that allows to turn off checking storage statuses before using them.

That feature with checking storages and caching results can be harmfull if only a single storage is used.
 
Despite you can configure the behaviour using custom cache TTL, it has some drawbacks anyway. If you set large cache TTL, you loose some flexibility: after storage got alive you need to wait for some time and nothing works during this period. If you set cache TTL to zero, you need to perform the storage check at least once per command execution.

So I made this configurable. If single storage is used, checking storages and caching their statuses is turned off by default. If many, it's turned on by default. If this setting is specified in the config, the value from there is used.


usernamedt (approved):
LGTM, let's also make a PR to skip writing the cache if there are no failover folders found to avoid stuck uploads.

PR #1582, opened 2023-10-20
Recheck storages and don't wait cache to expire if they all are dead

rdjjke:
In several previous PRs I introduced the failover storages feature, which allows to use multiple storages with different priorities so if the first storage is dead, the second can be used.

In that code, repetative checks and caching is used: if the requested storage status is outdated in the cache, a special requests would be performed to check if the storage is alive.

However, there was a problem: if all configured storages are dead, all these negative statuses are cached. And nothing works until the cache gets expired. The cache TTL is configurable, but the default value is 15min, and it may take less time for some storage to become alive.

In this PR I changed the logic a little. Any time the requested storages are dead, they are checked again. So caching is actually used only if there're some alive storages.

I also rewrote tests for `multistorage.cache.StatusCache` because the previous version was hard to read and append.

### DIFF
--- a/internal/multistorage/cache/status_cache.go
+++ b/internal/multistorage/cache/status_cache.go
@@ -28,9 +28,9 @@ type statusCache struct {
 	sharedMem   *storageStatuses
 	sharedMemMu *sync.Mutex
 
-	// sharedFile keeps the inter-process cache that's shared between different command executions.
-	sharedFile    string
-	sharedFileUse bool
+	// sharedFilePath is the path to the file that keeps the inter-process cache that's shared between different command executions.
+	sharedFilePath string
+	sharedFileUse  bool
 }
 
 type StatusCacheOpt func(c *statusCache)
@@ -42,9 +42,9 @@ func WithSharedMemory(mem *storageStatuses, memMu *sync.Mutex) StatusCacheOpt {
 	}
 }
 
-func WithSharedFile(file string) StatusCacheOpt {
+func WithSharedFile(filePath string) StatusCacheOpt {
 	return func(c *statusCache) {
-		c.sharedFile = file
+		c.sharedFilePath = filePath
 		c.sharedFileUse = true
 	}
 }
@@ -68,7 +68,7 @@ func NewStatusCache(
 	}
 
 	homeFileUse := true
-	homeFile, err := HomeStatusFile()
+	homeFilePath, err := HomeStatusFile()
 	if err != nil {
 		tracelog.WarningLogger.Printf("Can't use storage status cache file from $HOME: %v", err)
 		homeFileUse = false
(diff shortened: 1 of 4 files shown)


internal/multistorage/cache/che
rdjjke, line 1: Some garbage file, deleted it.

internal/multistorage/cache/status_cache.go
```diff
@@ -21,169 +23,281 @@ type statusCache struct {
 	storagesInOrder []NamedFolder
 	ttl             time.Duration
 	checker         AliveChecker
+
+	// sharedMem keeps the intra-process cache that's shared between different threads and subsequent storages requests.
+	sharedMem   *storageStatuses
+	sharedMemMu *sync.Mutex
+
+	// sharedFile keeps the inter-process cache that's shared between different command executions.
+	sharedFile    string
```
usernamedt, line 32: cacheFilePath?

internal/multistorage/cache/status_cache.go
```diff
@@ -21,169 +23,281 @@ type statusCache struct {
...
+	defer func() {
+		if c.sharedFileUse {
+			err := writeFile(c.sharedFile, newFile)
+			if err != nil {
+				tracelog.WarningLogger.Printf(
+					"Failed to write cache file, each subsequent command will check the storages again: %v",
+					err,
+				)
+			}
+		}
```
usernamedt, line 213: let's wrap this logic into some separate func, it will be easier to read and remove the code duplication

internal/multistorage/cache/status_cache.go
```diff
@@ -21,169 +23,281 @@ type statusCache struct {
...
-		tracelog.WarningLogger.Printf("Failed to read cache file, it will be overwritten: %v", err)
+	var oldFile storageStatuses
+	if c.sharedFileUse {
+		oldFile, err = readFile(c.sharedFile)
+		if err != nil {
+			tracelog.WarningLogger.Printf("Failed to read cache file, it will be overwritten: %v", err)
+		}
 	}
-	_, outdated := oldFile.splitByRelevance(c.ttl, c.storagesInOrder)
+	relevant, outdated := oldFile.splitByRelevance(c.ttl, c.storagesInOrder)
```
usernamedt, line 113: Will it result in nil dereference exception here if there is no shared file?

usernamedt (changes requested):

internal/multistorage/cache/status_cache.go
rdjjke: Do you want to emphasise that this is a path and not a file itself? If so, can we name it `sharedFilePath`? I just added this "shared" prefix to show the similarity between the `sharedMem` and `sharedFile`. And the difference is described in the code comments.

It also should be obvious that this file is used for "cache" because it's a field of the `statusCache` struct.

internal/multistorage/cache/status_cache.go
rdjjke: No, and this situation (and all the similar ones) is covered by tests. All methods of `storageStatuses` struct can work if this is nil - because both memory or file could be nil.
In this specific case all storages will be considered "outdated".

internal/multistorage/cache/status_cache.go
rdjjke: Did it, please take a look

internal/multistorage/cache/status_cache.go
rdjjke: Done

usernamedt (approved):
LGTM

PR #1583, opened 2023-10-20
Add MongoDB 7.0 support. Remove MongoDB 4.2 support. Fix dropIndexes operation replay.

dapqa:
### Database name
MongoDB

# Pull request description

## Added MongoDB 7.0 support and removed MongoDB 4.2 support

* bumped Ubuntu version to `focal` in base image for functional tests (`focal` is supported by all the Mongo versions from 4.4 to 7.0)
* added separate test configs for Mongo 4.4-6.0 and 6.1-7.0 (since starting from version 6.1, the `storage.journal.enabled` option is always `true` and no longer valid in config)
* removed `mongodb-database-tools` from `Dockerfile.mongodb` (since it was needed only for Mongo 4.2 and older)
* added test config file for WAL-G & Mongo 7.0
* added Mongo 7.0.2 test to a GitHub Actions pipeline

## Fixed `dropIndexes` operations replay

To replay a `dropIndexes` operation using the `applyOps` command, MongoDB needs the collection UUID (the `ui` field in an oplog document).
However, WAL-G erases all UUIDs before pushing an oplog.
This is a correct strategy (since operations are meant to be replayable without UUIDs), but it clashes with the current Mongo behavior when dropping indexes.

In Mongo versions prior to 7.0, such a command (without UUID) has no effect: it completes successfully, but no indexes are dropped.
In Mongo 7.0.1-7.0.2, there is a [bug](https://jira.mongodb.org/browse/SERVER-82349) that leads to server crash on such commands.

So, this PR solves two issues:
* it fixes `dropIndexes` operation replay on all the Mongo versions (by handling them separately, without using `applyOps`)
* it helps support Mongo 7.0, since without this fix an oplog replay may crash the server thread and drop the connection

Before this PR, index drop and corresponding collection drop were performed in a single scenario step, so we did not catch this bug earlier.
In this commit, new steps are added to verify that an index is dropped correctly.

Since index drop works now, we can face a `BackgroundOperationInProgressForNamespace` on Mongo versions prior to 5.2.
Ignored this error and documented this behavior.


VGoshev (approved):

usernamedt (approved):
GP tests are expected to fail, LGTM

PR #1586, opened 2023-10-25
Reuse SSH connections opened once when daemon started

rdjjke:
SSH folder implementation doesn't close opened connections. It leads to a problem when the WAL-G daemon is used, because it opens new connections on each command execution. 

In this PR I change the behaviour of the daemon: now it opens new connections only once on start and reuses them in every command executions.

The connections still aren't excplicitly closed, but they're closed automatically when the process dies.


cmd/pg/daemon.go
```diff
@@ -13,8 +14,12 @@ var daemonCmd = &cobra.Command{
 	Short: DaemonShortDescription,
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
+		storageStatusCache, err := postgres.InitStatusCache(true)
```
rdjjke, line 17: I moved opening new conenctions to the place that called only once instead of closing connections for two reasons:
1. I tried to close them, but it's not so easy: we can't just close `storage.Folder`s. because multiple folders can share the same connection (e.g. all folders got by listing the original folder share the same connection). So for proper closing, we need to return something (in addition to the folder itself) from `internal.ConfigureFolder` and close it later. We still can do that, but in another issue.
2. Even with proper closing, it's not really necessary to reopen the connections on each command exectution. It's better to open the minimum number of them once and reuse them later.

usernamedt: Does it mean that if the connection is dead, daemon will be unable to detect it until restart?

On top of that, it seems unclear why do we use something called  'statuscache' to provide some connection/folder. I think that we need to alter its name to represent the actual purpose, like `MultiStorageProvider` or something like that.

usernamedt (changes requested):

cmd/pg/daemon.go
rdjjke: > Does it mean that if the connection is dead, daemon will be unable to detect it until restart?

Hm, yeah, you're right. So, unfortunately, it means we need to explicitly close it somehow.

pkg/storages/sh/storage.go
```diff
@@ -59,7 +59,11 @@ func NewStorage(config *Config) (*Storage, error) {
 	client := NewSFTPLazy(address, sshConfig)
 
 	path := storage.AddDelimiterToPath(config.RootPath)
-	folder := NewFolder(client, path)
+	var folder storage.Folder = NewFolder(client, path)
```
usernamedt, line 62: why not to use compact var initialization here?

pkg/storages/s3/storage.go
```diff
@@ -56,7 +56,11 @@ func NewStorage(config *Config) (*Storage, error) {
 		return nil, fmt.Errorf("create new S3 uploader: %w", err)
 	}
 
-	folder := NewFolder(s3Client, uploader, config.RootPath, config)
+	var folder storage.Folder = NewFolder(s3Client, uploader, config.RootPath, config)
```
usernamedt, line 59: same for here

pkg/storages/sh/storage.go
rdjjke: Because `NewFolder` returns `*sh.Folder` struct, and `storage.Folder` is an interface.
It's then passed to `storage.WrapRootFolder` that works with `storage.Folder`.

I can change the declaration to `folder := storage.Folder(NewFolder(client, path))`, but it looks worse IMHO.

usernamedt (approved):

Issue #1385, opened 2022-11-23
backup-push does not work on the Postgresql 15


IgorOhrimenko:
### Database name
Postgresql 

# Issue description

### Describe your problem
Command backup-push does not work on the Postgresql 15 with error:
```
ERROR: 2022/11/23 17:36:15.308547 unexpected response: &{%!t(string=ERROR) %!t(string=ERROR) %!t(string=42601) %!t(string=syntax error) %!t(string=) %!t(string=) %!t(int32=0) %!t(int32=0) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=repl_scanner.l) %!t(int32=237) %!t(string=replication_yyerror) map[]}:
    github.com/jackc/pglogrepl.getBaseBackupInfo
        /home/user/go/src/github.com/wal-g/wal-g/vendor/github.com/jackc/pglogrepl/pglogrepl.go:408
```

I try last v2.0.1 and compiled master 3d22cab3.

### Please provide steps to reproduce

```bash
sudo -u postgres wal-g backup-push
```

### Please add config and wal-g stdout/stderr logs for debug purpose

<details><summary>/var/lib/postgresql/.walg.json</summary>
<p>

```yaml
{
    "WALG_LOG_LEVEL": "DEVEL",
    "WALE_S3_PREFIX": "s3://backup_postgresql_wal-g/cluster-psql-test/15",
    "AWS_ACCESS_KEY_ID": "backup_postgresql_wal-g",
    "AWS_ENDPOINT": "https://s3",
    "AWS_S3_FORCE_PATH_STYLE": "true",
    "AWS_SECRET_ACCESS_KEY": "pass",
    "AWS_REGION": "region",
    "PGDATA": "/var/lib/postgresql/15/cluster-psql-test",
    "PGHOST": "/var/run/postgresql",
    "PGPORT": "5432",
    "WALG_DELTA_MAX_STEPS": "7",
    "WALG_PREFETCH_DIR": "/var/lib/postgresql/",
    "WALG_TAR_SIZE_THRESHOLD": "4294967296",
    "WALG_UPLOAD_CONCURRENCY": "10",
    "WALG_COMPRESSION_METHOD": "brotli"
}
```
</p>
</details>

<details><summary>Full logs with "WALG_LOG_LEVEL": "DEVEL"</summary>
<p>

```bash
[...]

PR #1595, opened 2023-11-16
Supports remote backup push from PostgreSQL 15 or later (#1385)

tangtx3:
### Database name
PostgreSQL

# Pull request description
Supports remote backup push from PostgreSQL 15 or later (#1385)

### Describe what this PR fix
* In the streamFromPostgres method, CopyData case adds data receiving processing from pgsql version 15 or later.
* In the streamFromPostgres method, add NoticeResponse case.
* Troubleshoot EOF problems caused by the last zeroBlock message sent from pgsql15 or later.
* Add test for Supports remote backup push from PostgreSQL 15 or later

`wal-g backup-push`
`ERROR: 2023/11/16 10:37:36.274535 unexpected response: &{%!t(string=ERROR) %!t(string=ERROR) %!t(string=42601) %!t(string=syntax error) %!t(string=) %!t(string=) %!t(int32=0) %!t(int32=0) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=) %!t(string=repl_scanner.l) %!t(int32=247) %!t(string=replication_yyerror) map[]}`

test case！

AWS_ACCESS_KEY_ID=minio;
AWS_ENDPOINT=http://127.0.0.1:32009;
AWS_S3_FORCE_PATH_STYLE=true;
AWS_SECRET_ACCESS_KEY=123456;
PGHOST=127.0.0.1;
PGPASSWORD=123456;
PGPORT=5432;
PGUSER=postgres;
WALE_S3_PREFIX=s3://test-pg/pg15-1116

image


tangtx3:
Can the code owner help me review the code? THX！

Aevin1387:
What needs to be done to get this reviewed and merged? We would like to use the remote backup feature for our PG15+ clusters, and would prefer not to maintain/run a fork.

x4m:
@debebantur Can you please take a look? Architecturally the feature seem important.

tangtx3:
I agree with @x4m . The latest wal-g is no longer applicable to the remote backup capability of pgsql above version 15. Please take a look！ @Aevin1387

internal/databases/postgres/backup_push_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	deltaBaseSelector, err := internal.NewDeltaBaseSelector(
+		deltaFromName, deltaFromUserData, NewGenericMetaFetcher())
+	tracelog.ErrorLogger.FatalOnError(err)
+
+	userData, err := internal.UnmarshalSentinelUserData(userDataRaw)
+	tracelog.ErrorLogger.FatalfOnError("Failed to unmarshal the provided UserData: %s", err)
+
+	folder, err := internal.ConfigureFolder()
```
debebantur, line 106: Seems, that this err value is unused

internal/databases/postgres/backup_push_handler_test.go
```diff
@@ -0,0 +1,117 @@
...
+	uploader, err := internal.ConfigureUploaderToFolder(folder)
+
+	arguments := NewBackupArguments(uploader, dataDirectory, utility.BaseBackupPath,
+		permanent, verifyPageChecksums || viper.GetBool(internal.VerifyPageChecksumsSetting),
+		fullBackup, storeAllCorruptBlocks || viper.GetBool(internal.StoreAllCorruptBlocksSetting),
+		tarBallComposerType, NewRegularDeltaBackupConfigurator(deltaBaseSelector),
+		userData, withoutFilesMetadata)
+
```
debebantur, line 114: And here err is not checked too

x4m:
@sebasmannem Hi Sebas!
Just in case if you have some spare time... here's a good contribution to remote backups, but it need a rebase. Maybe you could move it forward?

Aevin1387:
@x4m @tangtx3 @sebasmannem I started looking into bringing this up to date and fixing the test. Unfortunately, this particular test requires PG actually be running, so it's better to cover with [this](https://github.com/wal-g/wal-g/blob/master/docker/pg_tests/scripts/tests/test_functions/remote_backup_and_restore_test.sh) integration test, but that requires either updating the docker image to PG15+ and fixing all the tests for the changes between PG10 and PG15 (or later), or splitting the tests, either of which is quite complicated.

x4m:
ugh... we definitely should upgrade from PG10... maybe gradually, PG11, PG12, PG13 etc.
Staying with PG10 would be a bad idea.
Though I cannot imagine what could go wrong with upgrade to PG14. I know in PG15 there are breaking changes. But in all other aspectes - it should be just fine.

edoronin:
Is Wal-g not supporting PG15+? Can't manage to run it successfully on Ubuntu 24 even when building from source

chipitsine:
this looks pretty interesting. seems we need to add modern postgres to build/test matrix

chipitsine:
> Is Wal-g not supporting PG15+? Can't manage to run it successfully on Ubuntu 24 even when building from source

the answer is "it depends".
personally, I use wal-g on various Postgres versions, including 15. on the other hand I do not use remote backup push

ieugen:
Remote backup would be very nice in some scenarios. 
Is deciding what version to use for testing the only think keeping this moving forward? 
From what I could tell , PostgreSQL 12 is EOL and 13 is getting EOL this year https://www.postgresql.org/support/versioning/

EamonZhang:
any news?

serprex:
replaced by #1909

PR #1598, opened 2023-11-20
Add tests for partitionStrings

ReshetnikovPavel:
### Database name
s3

# Pull request description
Added tests for partitionStrings in pkg/storages/s3/uploader.go


pkg/storages/s3/uploader_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+)
+
+func TestPartitionStrings(t *testing.T) {
+	testCases := []struct {
+		strings   []string
+		blockSize int
+		expected  [][]string
+	}{
+		{[]string{"1", "2", "3", "4", "5"}, 2, [][]string{{"1", "2"}, {"3", "4"}, {"5"}}},
```
usernamedt, line 15: I would also add some example string consisting of more than 1 char

usernamedt (changes requested):

pkg/storages/s3/uploader.go
```diff
@@ -96,9 +96,11 @@ func CreateUploaderAPI(svc s3iface.S3API, partsize, concurrency int) s3managerif
 	return uploaderAPI
 }
 
-// TODO : unit tests
 func partitionStrings(strings []string, blockSize int) [][]string {
 	// I've unsuccessfully tried this with interface{} but there was too much of casting
+	if blockSize <= 0 {
+		return [][]string{}
```
usernamedt, line 102: I guess in that just like in the case when blockSize exceeds the `len(strings)` we should return the `[]string{strings}`

usernamedt (changes requested):

usernamedt (approved):

PR #1600, opened 2023-11-21
Cherry-pick updates for MongoDB archiver

ArnobKumarSaha:


internal/databases/mysql/mysql_binlog.go
```diff
@@ -141,8 +141,12 @@ func GetBinlogStartTimestamp(filename string, flavor string) (time.Time, error)
 	parser.SetFlavor(flavor)
 	parser.SetVerifyChecksum(false) // the faster, the better
 	parser.SetRawMode(true)         // choose events to parse manually
+
 	err := parser.ParseFile(filename, 0, func(event *replication.BinlogEvent) error {
 		ts = event.Header.Timestamp
+		if ts == 0 {
+			return nil // keep reading until first event with timestamp is found
+		}
```
ostinru, line 149: Is there a real case when binlog event timestamp is 0?
If so, let's move to separate PR.

PR #1608, opened 2023-12-13
Retry fetch file 

debebantur:
Now file retries are not controlled by user.  Seems it is better to set it in config


docs/Greenplum.md
```diff
@@ -25,6 +25,7 @@ AWS_ACCESS_KEY_ID: "aws_access_key_id"
 WALG_UPLOAD_CONCURRENCY: 5
 WALG_PGP_KEY_PATH: "/path/to/PGP_KEY"
 WALG_DOWNLOAD_CONCURRENCY: 5
+WALG_DOWNLOAD_FILE_RETRIES: 10
```
EinKrebs, line 28:
```suggestion
WALG_DOWNLOAD_FILE_RETRIES: 15
```

internal/extract_test.go
```diff
@@ -89,6 +129,29 @@ func TestExtractAll_simpleTar(t *testing.T) {
...
+
+	err := internal.ExtractAllWithSleeper(buf, files, Sleeper05{})
+
+	duration := time.Since(start)
+
+	// file corrupted so we expect error
+	assert.Error(t, err)
+	// function must retry 7 times so overall time should be +- 3.5 seconds
+	assert.Greater(t, duration, time.Second*3)
+	assert.Less(t, duration, time.Second*4)
```
EinKrebs, line 152: Better use something different here

EinKrebs (changes requested):

internal/extract_test.go
debebantur: Changed it to checking logger

EinKrebs (approved):

PR #1609, opened 2023-12-18
Fix delete garbage in gp

debebantur:
### Database name
Greenplum

# Pull request description
When we call delete garbage for greenplum, it sorts wals by timestamp. But for permanent backups restore point could be in later wal. So I changed time check to segment lsn check from restore point metadata
Also changed gpstop  -M flag from "immediate" to "fast" to prevent data loss

### DIFF
--- a/internal/databases/greenplum/restore_point.go
+++ b/internal/databases/greenplum/restore_point.go
@@ -248,6 +248,55 @@ func FindRestorePointBeforeTS(timestampStr string, folder storage.Folder) (strin
 	return targetPoint.Name, nil
 }
 
+func FindRestorePointAfterTS(timestampStr string, folder storage.Folder) (string, error) {
+	ts, err := time.Parse(time.RFC3339, timestampStr)
+	if err != nil {
+		return "", fmt.Errorf("timestamp parse error: %v", err)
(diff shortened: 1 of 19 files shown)


internal/databases/greenplum/restore_point.go
```diff
@@ -248,6 +248,53 @@ func FindRestorePointBeforeTS(timestampStr string, folder storage.Folder) (strin
...
+		}
+
+		restorePointMetas = append(restorePointMetas, meta)
+	}
+
+	var targetPoint *RestorePointMetadata
+	for i := range restorePointMetas {
+		meta := restorePointMetas[i]
+		// target restore point should be started before or right at the provided ts
+		if meta.FinishTime.Before(ts) {
```
EinKrebs, line 279: If restore point should be started before, don't we need to use `meta.StartTime` here?

EinKrebs: If we indeed need to use `meta.FinishTime`, maybe the comment needs to be changed

EinKrebs (changes requested):

EinKrebs (approved):

PR #1612, opened 2023-12-24
propagate error from ParseLSN function

sivukhin:
### Context
I ran [govanish](https://github.com/sivukhin/govanish) linter (it still in the WIP phase) against `wal-g` repo and it found suspicious issue: compiler eliminates a lot of err checks after `ParseLSN` function like this:
```go
lsn, err := ParseLSN(lsnStr)
if err != nil {
	return "", nil, 0, errors.Wrap(err, "UploadLabelFiles: failed to parse finish LSN")
}
```

```
2023/12/24 15:38:26 it seems like your code vanished from compiled binary: func=[StartBackup], file=[/home/sivukhin/code/wal-g/internal/databases/postgres/bundle.go], line=[156], snippet:
	return "", 0, err
```

Compiler is right here - because `ParseLSN` function always return `nil` for error. There are quite a lot of code which handle potential error from `ParseLSN` function and it can really fail for incorrect log sequence numbers - so this looks like a bug. 

### Changes

This PR propagate error from `pgx.ParseLSN`.

Also, I removed several useless errchecks detected by `govanish`.

### DIFF
--- a/internal/databases/sqlserver/sqlserver.go
+++ b/internal/databases/sqlserver/sqlserver.go
@@ -508,7 +508,12 @@ func GetBackupProperties(db *sql.DB,
 	if err != nil {
 		return res, err
 	}
-	defer rows.Close()
+	defer func() {
+		err := rows.Close()
+		if err != nil {
+			tracelog.ErrorLogger.Printf("failed to close connection: %v", err)
+		}
+	}()
 	for rows.Next() {
 		var dbf BackupProperties
 		err = utility.ScanToMap(rows, map[string]interface{}{
--- a/internal/databases/sqlserver/sqlserver.go
+++ b/internal/databases/sqlserver/sqlserver.go
@@ -19,6 +19,7 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/wal-g/tracelog"
+
 	"github.com/wal-g/wal-g/internal"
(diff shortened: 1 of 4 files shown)


reshke (approved):

sivukhin:
What is the reason for tests failures? Should I do something about this?

internal/databases/sqlserver/sqlserver.go
```diff
@@ -508,9 +509,6 @@ func GetBackupProperties(db *sql.DB,
 		return res, err
 	}
 	defer rows.Close()
```
debebantur, line 511: rows.Close() could return error, better to change it to smth like

```suggestion
	defer func() {
	err := rows.Close()
		if err != nil {
			tracelog.ErrorLogger.Printf("failed to close connection %v", err)
		}
	}()
```

debebantur:
> What is the reason for tests failures? Should I do something about this?

Seems it is not connected to your pr, so don`t worry

internal/databases/sqlserver/sqlserver.go
sivukhin: Sure, added error handling in this defer

debebantur (approved):

mialinx (approved):
LGTM

PR #1622, opened 2024-01-17
MongoDB binary restore: Add resetup of shardIdentity field for mongod…

ElectricR:
…, add restore of valid shard list for mongocfg

### Database name
MongoDB

# Pull request description

This PR adds ability to restore valid fields for admin.system.version.shardIdentity and config.shards for mongod and mongocfg respectively


VGoshev:
This PR doesn't touch any MySQL code. Also looks like, that MySQL tests are broken for some time for now, so skipping it

PR #1624, opened 2024-01-18
Add namespace to greenplum partial-restore

debebantur:
Allowed to restore db/schema/table


docker/gp_tests/scripts/tests/partial_namespace_restoe.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
EinKrebs, line 1: typo in file name

docs/Greenplum.md
```diff
@@ -151,7 +151,7 @@ wal-g backup-fetch LATEST --mode=unpack --restore-config=restore-config.json --c
 ```
 
 #### Restore only specific databases
-During partial restore wal-g restores only specified databases' files. Use 'database' or 'database/namespace.table' as a parameter ('public' namespace can be omitted).
+During partial restore wal-g restores only specified databases' files. Use 'database', 'database/namespace.table' or 'database/namespace/table' as a parameter ('public' namespace can be omitted).
```
EinKrebs, line 154: Consider adding `'database/namespace.table'` case to gp_tests.

EinKrebs (changes requested):

docs/Greenplum.md
debebantur: added this case to test

EinKrebs (approved):

PR #1626, opened 2024-01-18
STORAGES doc using wrong Azure variable name

fgendc:
### Database name
This is just a documentation fix after #873 

# Pull request description
Originally the docs said to use `AZURE_STORAGE_SAS_TOKEN` to select a different Azure Storage Endpoint when using non-default regions, such as `AzureChinaCloud` or `AzureGermanCloud`. 

The right setting to use is `AZURE_ENVIRONMENT_NAME`, [as seen in the code](https://github.com/wal-g/wal-g/blob/4c82cc73519c721299bbb3ecab36657814df990e/pkg/storages/azure/configure.go#L18).

### Describe what this PR fix
Just a documentation update to help people with this use case.


usernamedt (approved):
LGTM. Thanks for your contribution!

PR #1634, opened 2024-02-01
GP, prevent multiple backups in the same time

debebantur:
Greenplum
Now to prevent multiple backups we check pg_is_in_backup(), but if wal-g process dies unexpectedly, we do not call pg_stop_backup. Instead of that it is better to use advisory locks, so wal-g can check that other wal-g backup process is now running.

### DIFF
--- a/internal/databases/greenplum/backup_push_handler.go
+++ b/internal/databases/greenplum/backup_push_handler.go
@@ -204,6 +204,8 @@ func (bh *BackupHandler) HandleBackupPush() {
 		bh.abortBackup()
 	}
 
+	//TODO prevent disconnecting wile long backup
+
 	// wait for segments to complete their backups
 	waitBackupsErr := bh.waitSegmentBackups()
 	if waitBackupsErr != nil {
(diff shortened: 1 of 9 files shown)


internal/databases/greenplum/backup_push_handler.go
```diff
@@ -204,16 +204,12 @@ func (bh *BackupHandler) HandleBackupPush() {
 		bh.abortBackup()
 	}
 
-	// WAL-G will reconnect later
```
usernamedt, line 207: Previously there were problems with keeping the master connection alive while taking long backups.

debebantur: Checked it on 5 hour and 12 hour backups, seems okay, will continue to monitor it

internal/databases/greenplum/query_runner.go
```diff
@@ -168,6 +168,21 @@ SELECT pg_is_in_backup(), -1;
 `
 }
 
+// TryGetLock tries to take advisory lock
+func (queryRunner *GpQueryRunner) TryGetLock() (err error) {
+	conn := queryRunner.Connection
+	var lockFree bool
+	err = conn.QueryRow("SELECT pg_try_advisory_lock('2131121')").Scan(&lockFree)
```
x4m, line 175: Maybe something more semantically loaded?

debebantur: changed it to hashtext('gp_backup')

x4m (approved):

PR #1649, opened 2024-02-28
Allow extra tags to be added to statsd metrics

nick-kentik:
### Database name

Common / generic / all. Enhancement to statsd metrics implementation

# Pull request description

### Describe what this PR fix

We're rolling out WAL-G into an environment where we have several different PostgreSQL databases running on the same host. Metrics are sent to a per-host Telegraf instance (which applies a `host` tag to the received metrics), and from there into a cluster-wide metrics store. At present, we have no way to distinguish WAL-G metrics relating to postgresql A from those relating to postgresql B.

This PR adds a config flag - `WALG_STATSD_EXTRA_TAGS` - that allows a set of static tags to be applied to every metric emitted during the wal-g process lifetime. We can configure that differently for wal-g when running against A, compared to B, and get the differentiation we need that way, e.g.:

```yaml
walg_statsd_extra_tags:
  database: "a"
```

We could also consider using this to allow us to differentiate between the `backup-push` and `wal-push` invocations of `wal-g`, so if uploads are failing in one but not the other, we can more easily tell which is at fault.

Just run `wal-g backup-push` against two different database instances. The metrics emitted by it have nothing that can distinguish between the two. The only way to do so with the current setup is to have a different `WAL_G_STATSD_ADDRESS` per database, and have *that* apply extra tags, which comes with a great deal of maintenance overhead.

cc @igorwwwwwwwwwwwwwwwwwwww (hi!) who put together the original metrics implementation

### DIFF
--- a/internal/metrics.go
+++ b/internal/metrics.go
@@ -53,13 +53,15 @@ func PushMetrics() {
 		return
 	}
 
-	err := pushMetrics(address)
+	extraTags := viper.GetStringMapString(StatsdExtraTagsSetting)
+
+	err := pushMetrics(address, extraTags)
 	if err != nil {
 		tracelog.WarningLogger.Printf("Pushing metrics failed: %v", err)
 	}
 }
 
-func pushMetrics(address string) error {
+func pushMetrics(address string, extraTags map[string]string) error {
 	config := &statsd.ClientConfig{
 		Address:       address,
 		UseBuffered:   true,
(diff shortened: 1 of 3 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Looks like a reasonable addition. 👍

internal/metrics.go
```diff
@@ -53,13 +53,15 @@ func PushMetrics() {
 		return
 	}
 
-	err := pushMetrics(address)
+	extraTags := viper.GetStringMapString(StatsdExtraTagsSetting)
```
nick-kentik, line 56: Just digging a bit into what viper does with this - I don't *think* we can easily populate a `StringMapString` from env or cli, so perhaps this would be better as `GetString`, the breaking down something like `foo=bar,baz=qux`? I don't see any other users of `GetStringMapString` in wal-g, anyway. Happy with whatever works here.

debebantur (approved):
Seems like a great addition! Thank you for this feature

PR #1653, opened 2024-03-06
Added s3 response code metrics

debebantur:
### Database name
All databases

# Pull request description

1) Added s3 response code metrics (200, 400, 404, 429, 500)
2) Moved config to separate package to prevent cyclic dependencies. Many packages import whole "internal", when they actually need only some settings.

### DIFF
--- a/internal/statistics/metrics.go
+++ b/internal/statistics/metrics.go
@@ -16,6 +16,10 @@ import (
 type metrics struct {
 	UploadedFilesTotal       prometheus.Counter
 	UploadedFilesFailedTotal prometheus.Counter
+
+	S3Code200 prometheus.Counter
+	S3Code400 prometheus.Counter
+	S3Code500 prometheus.Counter
 }
 
 var (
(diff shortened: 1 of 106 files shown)


EinKrebs (approved):

x4m (approved):

internal/statistics/metrics.go
```diff
@@ -43,24 +79,46 @@ func init() {
...
+func WriteStatusCodeMetric(code int) {
+	if code >= 500 {
+		WalgMetrics.S3Code500.Inc()
+	} else if code >= 400 {
+		if code == 404 {
+			WalgMetrics.S3Code404.Inc()
+		} else if code == 429 {
+			WalgMetrics.S3Code429.Inc()
+		} else {
+			WalgMetrics.S3Code400.Inc()
```
mialinx, line 115: It's a bit confusing to collapse all 4xx codes (but 429) to one counter.

May be it's better to use `prometheus.GaugeVec`  https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#GaugeVec
for fair http code statistics ?

```
WalgMetrics.S3Codes.WithLabelValues(strconv.Itoa(code)).Inc()
```

mialinx (changes requested):

PR #1661, opened 2024-03-16
Add unit tests for internal.databases.greenplum.buildFetchCommand

denisSurkov:
# Pull request description

### Describe what this PR fix
Adding unit tests for internal.databases.greenplum.buildFetchCommand

### DIFF
--- a/internal/databases/greenplum/backup_fetch_handler_test.go
+++ b/internal/databases/greenplum/backup_fetch_handler_test.go
@@ -1,10 +1,11 @@
 package greenplum
 
 import (
-	"testing"
-
 	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"github.com/stretchr/testify/assert"
+	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/internal/config"
+	"testing"
 )
 
 func TestPrepareContentIDsToFetch(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/databases/greenplum/backup_fetch_handler_test.go
```diff
@@ -50,3 +51,131 @@ func TestPrepareContentIDsToFetch(t *testing.T) {
...
+	}()
+	config.CfgFile = "testConfig"
+
+	testcases := []struct {
+		handler   *FetchHandler
+		contentID int
+		cmdLine   string
+	}{
+		{
+			&FetchHandler{
```
denisSurkov, line 68: I'm not sure if I should use `NewFetchHandler` or construct an object like this. At first I tried to use `NewFetchHandler` but it quicky became very complicated.

But I guess I should use `NewFetchHandler`. It's easer to refactor. I think no one will be happy if they change `cluster.SeqConfig` object and this test will fail. 

But I need some tips how to call `NewFetchHandler` correctly without a headache.

internal/databases/greenplum/backup_fetch_handler_test.go
```diff
@@ -50,3 +51,131 @@ func TestPrepareContentIDsToFetch(t *testing.T) {
 		assert.Equal(t, tc.contentIDsToFetch, contentIDsToFetch)
 	}
 }
+
+func TestBuildFetchCommand(t *testing.T) {
+	beforeValue := config.CfgFile
+	defer func() {
+		config.CfgFile = beforeValue
+	}()
+	config.CfgFile = "testConfig"
```
denisSurkov, line 60: I agree this is tricky. But `buildFetchCommand` relies on this global variable. May be there is some other way to do it. 

Without this setting `cmdLine` would be something like this `... --config= ...`. I think it's not good.

denisSurkov:
@usernamedt , could you help with review?

denisSurkov:
@x4m , sorry to bother you, but may be you could help with review too?

internal/databases/greenplum/backup_fetch_handler_test.go
debebantur: Well, if you want to use `NewFetchHandler `, the most difficult things seem to be `BackupSentinelDto` and `SegConfigMaker`. From the first one you need only segments, so you can define it like
```
dto := BackupSentinelDto{
		Segments: []SegmentMetadata{
			{
			DatabaseID: 1,
			ContentID: 1,
			Role: Primary,
			Port: 1234,
			Hostname: "name",
			DataDir: "datadir",		
			BackupID: "backup",
			},
		},
	}
```
And `SegConfigMaker` you can make from string:
```
seg, err := NewRestoreCfgSegMaker(strings.NewReader(`{"segments":{"1":{"hostname":"name", "port":1234, "data_dir": "dir"}}}`))
```

denisSurkov: @debebantur , thank you

> if you want to use

Do you think I should do it the way you suggested or it's okey to leave it like this?

And with machinery all this variables is it possible to declare it in `testcases` array?

debebantur: I believe that it is ok to leave it, anyway we need to test only `buildFetchCommand`. Also it seems difficult to declare all this structures in testcases array.

internal/databases/greenplum/backup_fetch_handler_test.go
```diff
@@ -50,3 +51,131 @@ func TestPrepareContentIDsToFetch(t *testing.T) {
...
+				"--target-user-data=\"{\\\"id\\\":\\\"other-value-from-testing\\\"}\" " +
+				"--config=testConfig " +
+				"--restore-only=test1,test2 " +
+				">> /wal-g-log-seg1.log 2>&1",
+		},
+	}
+
+	for _, tc := range testcases {
+		cmdLine := tc.handler.buildFetchCommand(tc.contentID)
+		assert.Equal(t, tc.cmdLine, cmdLine)
```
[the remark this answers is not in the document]
debebantur: Sorry, i forgot that `Fatalf` does not throw panic but instead exits process. Still we can test that we do not have backup id for content id. 
We can write something like 
```
func TestBuildFetchCommandCrushes(t *testing.T) {
	handler := &FetchHandler{
		cluster: &cluster.Cluster{
			ByContent: map[int][]*cluster.SegConfig{
				1: {
					{
						DbID:      1,
						ContentID: 2,
						Role:      "controlled",
						Port:      1234,
						Hostname:  "test.com",
						DataDir:   "/etc/test/",
					},
				},
			},
		},
		backupIDByContentID: map[int]string{},
		backup:              internal.Backup{},
		contentIDsToFetch:   map[int]bool{1: true},
	}

	if os.Getenv("TEST_BUILD_FETCH_COMMAND") == "1" {
		handler.buildFetchCommand(1)
		return
	}

	cmd := exec.Command(os.Args[0], "-test.run=TestBuildFetchCommandCrushes")
	cmd.Env = append(os.Environ(), "TEST_BUILD_FETCH_COMMAND=1")
	err := cmd.Run()
	if e, ok := err.(*exec.ExitError); ok && !e.Success() {
		return
	}
	t.Fatalf("process ran with err %v, want exit status 1", err)
}
```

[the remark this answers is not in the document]
denisSurkov: That's actually pretty clever, thank you!

denisSurkov:
@debebantur , so, I guess that's it

Thank you for your time and patience

Of course, If you have something else to mention I will fix it

But if everything looks good than what should I do next? 

Wait until someone merges it? May be you know somebody to ask to merge?

internal/databases/greenplum/backup_fetch_handler_test.go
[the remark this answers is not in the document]
debebantur: Yep, that`s correct. If you want more cool ideas, there is a presentation by one of go developers https://go.dev/talks/2014/testing.slide#1

debebantur:
> @debebantur , so, I guess that's it
> 
> Thank you for your time and patience
> 
> Of course, If you have something else to mention I will fix it
> 
> But if everything looks good than what should I do next?
> 
> Wait until someone merges it? May be you know somebody to ask to merge?

LGTM, we are merging it

debebantur (approved):

PR #1681, opened 2024-04-04
Clarified dev installing instruction on Ubuntu for go 1.16+

### LINKED ISSUE #1632: "go get" Build instructions in README are no longer correct
### Database name
any

# Issue description

Ubuntu Build instructions in README are wrong for recent versions of go.

### Describe your problem
The README mentions that to build wal-g under Ubuntu,
```go get github.com/wal-g/wal-g```

I believe go changed the way this works a few versions ago and now gives:
```
% go get github.com/wal-g/wal-g
go: go.mod file not found in current directory or any parent directory.
	'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.

```
(This was with go 1.21.5 on darwin and Linux but I'd expect similar from others)

I got it working by `git clone`ing the repo and building it that way, but trying to follow the instructions in the README had me lost for a little while.

I'm not sure what the correct command would be, `go install` didn't work (for me) either, I'm not familiar with go.

Start with a reasonably clean Ubuntu or MacOSX install and follow the build instructions in README.md

```
# Install latest Go compiler
sudo add-apt-repository ppa:longsleep/golang-backports 
sudo apt update
sudo apt install golang-go

# Install lib dependencies
sudo apt install libbrotli-dev liblzo2-dev libsodium-dev curl cmake

# Fetch project and build
go get github.com/wal-g/wal-g
```
^ error happens here

n/a

### PR DESCRIPTION
**Pull request by homper**
Fixes #1632
Maybe `git clone` should be with `--single-branch --depth=1` but this looks uglier and probably developers need whole repo

Installing like this will be more convenient
```
go install github.com/wal-g/wal-g/main/pg@latest
```
but wrong in this case, for developing, espesially when additional lib support needed

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -304,6 +304,7 @@ sudo apt install libbrotli-dev liblzo2-dev libsodium-dev curl cmake
 # Go 1.15 and below
 go get github.com/wal-g/wal-g
 # Go 1.16+ - just clone repository to $GOPATH
+# if you want to save space add --depth=1 or --single-branch
 git clone https://github.com/wal-g/wal-g $(go env GOPATH)/src/github.com/wal-g/wal-g
 
 cd $(go env GOPATH)/src/github.com/wal-g/wal-g
@@ -339,6 +340,7 @@ brew install cmake
 # Go 1.15 and below
 go get github.com/wal-g/wal-g
 # Go 1.16+ - just clone repository to $GOPATH
+# if you want to save space add --depth=1 or --single-branch
 git clone https://github.com/wal-g/wal-g $(go env GOPATH)/src/github.com/wal-g/wal-g
 
 cd $(go env GOPATH)/src/github.com/wal-g/wal-g
--- a/docs/README.md
+++ b/docs/README.md
@@ -334,11 +334,23 @@ GOBIN=/usr/local/bin make pg_install
 ```sh
 # brew command is Homebrew for Mac OS
 brew install cmake
+
+# Fetch project and build
+# Go 1.15 and below
+go get github.com/wal-g/wal-g
+# Go 1.16+ - just clone repository to $GOPATH
+git clone https://github.com/wal-g/wal-g $(go env GOPATH)/src/github.com/wal-g/wal-g
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by homper** [comment]
Also added same instructions to mac os so it is in one place for each os
Also looks like call of ./link_brotli.sh and ./link_libsodium.sh not needed if USE_BROTLI and USE_LIBSODIUM provided, as I understanded, but I left it for now

File: docs/README.md [hunk 1]
```diff
@@ -293,16 +293,26 @@ Optional:
...
 
 # Install lib dependencies
 sudo apt install libbrotli-dev liblzo2-dev libsodium-dev curl cmake
 
 # Fetch project and build
+# Go 1.15 and below
 go get github.com/wal-g/wal-g
-cd ~/go/src/github.com/wal-g/wal-g
+# Go 1.16+ - just clone repository to $GOPATH
+git clone https://github.com/wal-g/wal-g $(go env GOPATH)/src/github.com/wal-g/wal-g
```
debebantur, line 307: Cloning all repo seems reasonable, since we need it for development. However i believe it would be better to mention `--single-branch --depth=1` options. Something like `# if you want to save space add --depth=1 or --single-branch`

homper: Updated

homper:
Also, seems that in Makefile install_and_build_pg are just deps pg_build now, so maybe it's better somehow merge ubuntu and macos installation instructions, or add reference to macos instructions to follow ubuntu installation except apt-get-s?

debebantur (approved):
Thank you, for keeping docs up to date!

debebantur:
> Also, seems that in Makefile `install_and_build_pg` are just `deps pg_build` now, so maybe it's better somehow merge ubuntu and macos installation instructions, or add reference to macos instructions to follow ubuntu installation except apt-get-s?

Merging instructions for different systems looks questionable. I am afraid it can cause confusion.
If you believe that this will be beneficial, you can open new issue for discussion.

PR #1684, opened 2024-04-10
Add glob pattern support in storagetools

ReshetnikovPavel:
### Database name

# Pull request description
Add glob pattern support for storagetools ls, rm and cat commands

### Describe what this PR fix
// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>


internal/storagetools/remove_handler_test.go
```diff
@@ -72,4 +72,38 @@ func TestHandleRemove(t *testing.T) {
 			assert.False(t, exists)
 		}
 	})
+
+	t.Run("remove files that match glob pattern", func(t *testing.T) {
```
debebantur, line 76: It would be great to add unit tests also for cat handler and list handler

debebantur:
Also check failing tests, seems like there is a bug
`panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/wal-g/wal-g/cmd/common/st.glob..func6.1({0x3fb1aa0, 0xc0005b3dc0})
	/go/src/github.com/wal-g/wal-g/cmd/common/st/folder_list.go:25 +0x16b`

ReshetnikovPavel:
> Also check failing tests, seems like there is a bug
> 
> ```
> panic: runtime error: index out of range [0] with length 0
> 
> goroutine 1 [running]:
> github.com/wal-g/wal-g/cmd/common/st.glob..func6.1({0x3fb1aa0, 0xc0005b3dc0})
> 	/go/src/github.com/wal-g/wal-g/cmd/common/st/folder_list.go:25 +0x16b
> ```

Oh, it was already there. Seems to be quite easy to fix though. Okay, I'll do it

internal/storagetools/remove_handler_test.go
debebantur: Actually, it seems like another task

debebantur (approved):

Issue #1683, opened 2024-04-10
mysql backup-restore should check architecture


ostinru:
### Database name
xtrabackup cannot restore backup on other architecture (CPU | windows/linux). So, we can also check architecture before restoring backup.

Let's add WARNING for now...

ostinru:
they said, that clone-plugin refuses to copy database on different architecture... but it is [rumoured](https://mysqlcommunity.slack.com/archives/C8Y94HRMX/p1730723490601109) that it should be fine.

PR #1693, opened 2024-04-21
WALG-1683. Save CPU arch / OS to mysql sentinel.

ostinru:
xtrabackup cannot restore backup on other architecture (CPU | windows/linux). So, we can also check architecture before restoring backup.


ostinru:
Hm... probably, we should use GOOS / GOARCH - because during `backup-fetch` we don't have MySQL to get @@version_compile_os =(

teem0n (approved):

ostinru:
#1683

PR #1696, opened 2024-04-25
Fix Mongodb commitIndexBuild panic in olpog-replay

sayedppqq:
### Database name
Mongodb

### Describe what this PR fix
For `oplog-replay`, any oplog with commitIndexBuild command, in indexSpecFromCommitIndexBuilds func, its getting panic due to for loop's variables misposition in [indexSpecs slice](https://github.com/wal-g/wal-g/blob/master/internal/databases/mongo/oplog/applier.go#L259C1-L265C7)

multiple index commit in oplog-replay

### DIFF
--- a/internal/databases/mongo/oplog/applier.go
+++ b/internal/databases/mongo/oplog/applier.go
@@ -254,8 +254,8 @@ func indexSpecFromCommitIndexBuilds(op db.Oplog) (string, []client.IndexDocument
 				if !ok {
 					return "", nil, NewTypeAssertionError("bson.D", fmt.Sprintf("indexes[%d]", i), elemE.Value)
 				}
-				for i := range elements {
-					elemE = elements[i]
+				for j := range elements {
+					elemE = elements[j]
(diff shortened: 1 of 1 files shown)


internal/databases/mongo/oplog/applier.go
```diff
@@ -254,8 +254,8 @@ func indexSpecFromCommitIndexBuilds(op db.Oplog) (string, []client.IndexDocument
 				if !ok {
 					return "", nil, NewTypeAssertionError("bson.D", fmt.Sprintf("indexes[%d]", i), elemE.Value)
 				}
-				for i := range elements {
-					elemE = elements[i]
+				for j := range elements {
```
Apelsin234, line 257: Maybe instead of this line 
'for _, elemE := range elements {'

sayedppqq: updated

Apelsin234:
bug was fixed in another commit

PR #1698, opened 2024-04-26
Orioledb basic support

homper:
I added some basic support of [orioledb](https://github.com/orioledb/orioledb) by wal-g
I tested that with these small modifications these features work:
- backup-push
- backup-fetch
- wal-push
- wal-fetch

Adding Dockerfile, scripts, and log of running this test to prove that it works
Sorry, that I'm not added it to pg_tests yet. Was not sure.

For now I added code to postgres wal-g version with option to enable it

Orioledb implemented as postgres extension with own table access method, and so it basically works alongside regular postgres, but uses own file format to store table pages
Without modification wal-g performs a full backup (with WALG_DELTA_MAX_STEPS=6)

wal-push and wal-fetch works because we implemented custom wal records, so no additional code required for this

I duplicated some functions with Orioledb* prefixed versions just to not affect any non-postgres related code for now

But I wan't to discuss before going any further what will be the better way to implement this
I see three variants:
1. Implement orioledb as a separate wal-g version, but it will duplicate many of postgres-es code
2. Adding some kind of generic interface for processing data of postgres extensions
3. Leave just orioledb as option but implement it in a different, better way

What do you think?

Files:
[ubuntu_check.log](https://github.com/wal-g/wal-g/files/15126355/ubuntu_check.log)
[wal-g-orioledb-test.zip](https://github.com/wal-g/wal-g/files/15126356/wal-g-orioledb-test.zip)

### DIFF
--- a/cmd/pg/backup_push.go
+++ b/cmd/pg/backup_push.go
@@ -28,7 +28,6 @@ const (
 	deltaFromNameFlag         = "delta-from-name"
 	addUserDataFlag           = "add-user-data"
 	withoutFilesMetadataFlag  = "without-files-metadata"
-	withOrioledb              = "with-orioledb"
 
 	permanentShorthand             = "p"
 	fullBackupShorthand            = "f"
@@ -107,8 +106,7 @@ var (
 			userData, err := internal.UnmarshalSentinelUserData(userDataRaw)
 			tracelog.ErrorLogger.FatalfOnError("Failed to unmarshal the provided UserData: %s", err)
 
-			var arguments postgres.BackupArguments
-			arguments = postgres.NewBackupArguments(uploader, dataDirectory, utility.BaseBackupPath,
+			arguments := postgres.NewBackupArguments(uploader, dataDirectory, utility.BaseBackupPath,
 				permanent, verifyPageChecksums || viper.GetBool(conf.VerifyPageChecksumsSetting),
 				fullBackup, storeAllCorruptBlocks || viper.GetBool(conf.StoreAllCorruptBlocksSetting),
 				tarBallComposerType, postgres.NewRegularDeltaBackupConfigurator(deltaBaseSelector),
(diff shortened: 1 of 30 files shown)


akorotkov:
@homper, thank you.
I'd like to highlight some details.
1. OrioleDB implements row-level WAL instead of block-level WAL. Thus, an OrioleDB data page doesn't contain the LSN.  But instead, an OrioleDB data page contains a checkpoint number. This pull request implements incremental backup by archiving only pages whose checkpoint number is greater than the checkpoint number of the last backup.
2. This is proof-of-concept. Obviously, this needs some cleanup. Also, it doesn't support OrioleDB compressed tables yet. We would like to get your initial feedback on which way to move it further (any of the options proposed by @homper or even something different).

cmd/pg/backup_push.go
```diff
@@ -28,6 +28,7 @@ const (
 	deltaFromNameFlag         = "delta-from-name"
 	addUserDataFlag           = "add-user-data"
 	withoutFilesMetadataFlag  = "without-files-metadata"
+	withOrioledb              = "with-orioledb"
```
x4m, line 31: can we query database for this somehow? Why ask user, if we can ask db?

x4m (approved):
Thanks for the PR! Of course, this is a functionality we want in WAL-G.

> But I wan't to discuss before going any further what will be the better way to implement this

Personally, I'd prefer if OrioleDB would be supported by regular wal-g binary for Postgres. I do not see any reason to make OrioleDB 2nd class citizen, enabled by some kind of option etc.
Also I see no reason to invent some generalization layer for other extensions.
Let's follow simple contract:
1. Plz do not break Postgres stuff :) And other DBs
2. If something bugs regarding OrioleDB will emerge and WAL-G team cannot fix them - we will ask for assistance.
3. Modularity of a code is desired. It's cool if OrioleDB logic is concentrated in some specific place, but not so terrible if it's incorporated into many appropriate places.

Thanks!

akorotkov:
@x4m, thank you! The contract is accepted!

x4m (approved):
Overall looks like a great addition, thanks!
Code is well-writted.
But one thing that worries me is we would like to have some integration tests...

homper:
Finally added integration tests
And also added message to "catchup-push" that full backup performed for orioledb data if it is present
Probably one ore two lines needed in documentation about orioledb support, but I'm not sure where to put it
Also made small non-related fix for `clean_compose` in Makefile

PR #1713, opened 2024-05-21
Added retention period for s3

debebantur:
Added retention period support for s3.
To prevent files from deletion for some time, added support for:
S3_RETENTION_PERIOD - time in seconds to store file version
S3_RETENTION_MODE - mode of retention (GOVERNANCE/COMPLIANCE)
Warning: bucket object lock configuration needs to be set before using these options

### DIFF
--- a/docs/STORAGES.md
+++ b/docs/STORAGES.md
@@ -82,6 +82,14 @@ Overrides the default minimum time between retries when throttled in millisecond
 
 Overrides the default minimum time between retries when throttled in milliseconds. Default is 300000ms.
 
+* `S3_RETENTION_PERIOD`
+
+Sets retention period in seconds to files, uploaded to s3. Default is disabled with value -1.
+
+* `S3_RETENTION_MODE`
+
+Sets mode of retention (GOVERNANCE/COMPLIANCE). Default is GOVERNANCE, which means, that files can still be deleted if user has special permissions.
+
 GCS
 -----------
 To store backups in Google Cloud Storage, WAL-G requires that this variable be set:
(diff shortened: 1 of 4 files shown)


reshke (approved):

docs/STORAGES.md
```diff
@@ -82,6 +82,14 @@ Overrides the default minimum time between retries when throttled in millisecond
 
 Overrides the default minimum time between retries when throttled in milliseconds. Default is 300000ms.
 
+* `S3_RETENTION_PERIOD`
+
+Sets retention period in seconds to files, uploaded to s3. Default is disabled with value -1.
+
+* `S3_RETENTION_MODE`
+
+Sets mode of retention (GOVERNANCE/COMPLIANCE). Default is GOVERNANCE, which means, that files can still be deleted if user has special permissions.
```
ostinru, line 91: What does `COMPLIANCE` mean?

ostinru:
@debebantur, how wal-g behave when we will try to delete backup? How users can understand what's wrong, fix it and try again?
I will guess that this flags are not shown in `wal-g st ls` - and it may be tricky to remove this protection when you don't see it.

debebantur:
> @debebantur, how wal-g behave when we will try to delete backup? How users can understand what's wrong, fix it and try again?
> I will guess that this flags are not shown in `wal-g st ls` - and it may be tricky to remove this protection when you don't see it.

Well, yes, it is a good idea to show locks in st tools, i'll work on it.
Now, it works the same as if we just turn on versioning in bucket. File will be mark deleted, but version remains. Locks only add retention period to this version, so it can't be deleted for some time

docs/STORAGES.md
debebantur: My bad, will fix docks. Compliance mode prohibits to delete object version without any way to bypass it

PR #1718, opened 2024-05-24
Fixes for working with s3

debebantur:
- fix st ls
After [#1684](https://github.com/wal-g/wal-g/pull/1684) when calling st ls without specified dir, output does not show root folder content
- fix validation
Validation used ListObjectPages, so it listed whole dir. Instead changed it to ListObjects with max keys = 1

### DIFF
--- a/pkg/storages/s3/folder.go
+++ b/pkg/storages/s3/folder.go
@@ -2,7 +2,6 @@ package s3
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"path"
 	"strings"
@@ -12,6 +11,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/pkg/errors"
+	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/pkg/storages/storage"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/storages/s3/folder.go
```diff
@@ -156,14 +156,13 @@ func (folder *Folder) ListFolder() (objects []storage.Object, subFolders []stora
...
-	listFunc func(commonPrefixes []*s3.CommonPrefix, contents []*s3.Object)) error {
-	var err error
+	listFunc func(commonPrefixes []*s3.CommonPrefix, contents []*s3.Object)) (err error) {
 	if folder.config.UseListObjectsV1 {
 		err = folder.listObjectsPagesV1(prefix, delimiter, maxKeys, listFunc)
 	} else {
 		err = folder.listObjectsPagesV2(prefix, delimiter, maxKeys, listFunc)
 	}
-	return err
+	return
```
EinKrebs, line 165: Why ignore the error here?

EinKrebs (changes requested):

pkg/storages/s3/folder.go
debebantur: err is declared in return block of function, so it will be automatically returned

ostinru (approved):
Looks great!

EinKrebs (approved):

PR #1734, opened 2024-06-14
Redis: delete backups by name

victor-tsykanov:
### Database name
Redis

# Pull request description

### Describe what this PR fix
This PR make it possible to delete specific Redis backup by providing its name in newly introduced `--target` option for `delete` command.

None

None


internal/backup_util.go
```diff
@@ -199,13 +199,28 @@ func FolderSize(folder storage.Folder, path string) (int64, error) {
...
-	retainAll := retainCount == nil && retainAfter == nil
+	retainAll := retainCount == nil && retainAfter == nil && target == nil
 	retainedCount := 0
+	deleteTarget := target != nil
 	for i := range backups {
 		backup := backups[i]
+
+		if deleteTarget {
+			if backup.Name() == *target {
+				purge[backup.Name()] = true
```
ostinru, line 215: I think that we shouldn't delete `target` backup if it is `permanent`.

How it works in Pg/MySQL:
https://github.com/wal-g/wal-g/blob/v3.0.1/internal/delete_handler.go#L391-L397

How it is documented in README.md:
`
Is used to delete backups and WALs before them. By default, ``delete`` will perform a dry run. If you want to execute deletion, you have to add ``--confirm`` flag at the end of the command. Backups marked as permanent will not be deleted.
`

victor-tsykanov: I've fixed it and added the test.

PR #1742, opened 2024-06-28
Add SSH jump server support

pmatrosov:
### Database name
I'm using PostgreSQL database only.

# Pull request description

### Describe what this PR fix
Fix adds support of SSH jump servers for SSH storage. 
This is useful in some cases where the database servers do not have direct access to the Internet.
We successfully use this fix on our production database.


chipitsine:
this may be configured already in ~/.ssh/config

having that as a feature require you to maintain code in go.

Issue #1028, opened 2021-07-01
backup-push Connection cancelled by remote host


iget-master:
### Database name
Postgres 13

# Issue description

Trying to run backup-push to create my first backup, and this error occurs:

C:\Program Files\PostgreSQL\13\data>c:\\wal-g\\wal-g.exe --walg-file-prefix "D:\\ACS_Backup_Postgres\\WAL-G" --pgport 5432 --pguser postgres --pgpassword "xxxxxxxxxxxxxxxxxxx" backup-push "C:\Program Files\PostgreSQL\13"
INFO: 2021/07/01 10:56:00.400758 Selecting the latest backup as the base for the current delta backup...
ERROR: 2021/07/01 10:56:00.936214 read tcp 127.0.0.1:64006->127.0.0.1:5432: wsarecv: An existing connection was forcibly closed by the remote host.

### Describe your problem
// here is your problem

### Please provide steps to reproduce
// it can really help

### Please add config and wal-g stdout/stderr logs for debug purpose

```
DEBUG: 2021/07/01 10:59:15.216132 --- COMPILED ENVIRONMENT VARS ---
DEBUG: 2021/07/01 10:59:15.216132 =::=::\
DEBUG: 2021/07/01 10:59:15.217220 =C:=C:\Program Files\PostgreSQL\13\data
DEBUG: 2021/07/01 10:59:15.221885 =ExitCode=00000001
DEBUG: 2021/07/01 10:59:15.221885 ALLUSERSPROFILE=C:\ProgramData
DEBUG: 2021/07/01 10:59:15.221885 APPDATA=C:\Users\Jacarey\AppData\Roaming
DEBUG: 2021/07/01 10:59:15.221885 AWS_ACCESS_KEY=
DEBUG: 2021/07/01 10:59:15.221885 AWS_ACCESS_KEY_ID=
DEBUG: 2021/07/01 10:59:15.221885 AWS_CA_BUNDLE=
DEBUG: 2021/07/01 10:59:15.221885 AWS_CONFIG_FILE=
DEBUG: 2021/07/01 10:59:15.221885 AWS_DEFAULT_OUTPUT=
DEBUG: 2021/07/01 10:59:15.221885 AWS_DEFAULT_REGION=
DEBUG: 2021/07/01 10:59:15.221885 AWS_ENDPOINT=
DEBUG: 2021/07/01 10:59:15.221885 AWS_PROFILE=
[...]

PR #1745, opened 2024-07-01
Fix "backup-push Connection cancelled by remote host" error on Windows

rail-rain:
As in https://github.com/wal-g/wal-g/issues/1028#issuecomment-2197864705, we're trying to upstream the patch for #1028.

This PR simply logs the error emitted on closing the connection to Postgres rather than failing on it, allowing the command to proceed. This seems to be more common behaviour across the codebase.

This does mean that on Windows users are likely to see the `wsarecv: An existing connection was forcibly closed by the remote host` error message every time. Integrating the fix from https://github.com/jackc/pgx/issues/637 avoids this cluttering. However, it requires upgrading the dependency `pgx` from v3 to v4.2[^1].

[^1]: The fix commit mentioned in the `pgx` issue thread was for `pgconn` v1.2.0, which `pgx` v4.2.0 depends on. Note that `wal-g`’s dependency on `pgconn` is unrelated since the the error we’re experiencing is coming from `pgx`, and `pgx` v3 doesn’t use `pgconn`.

### DIFF
--- a/internal/databases/postgres/backup_push_handler.go
+++ b/internal/databases/postgres/backup_push_handler.go
@@ -582,10 +582,7 @@ func GetPgServerInfo(keepRunner bool) (pgInfo BackupPgInfo, runner *PgQueryRunne
 	tracelog.DebugLogger.Printf("Timeline: %d", pgInfo.Timeline)
 
 	if !keepRunner {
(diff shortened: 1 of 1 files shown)


x4m (approved):
Looks good to me, thanks!

PR #1760, opened 2024-07-28
Reduce log clutter from storages

x4m:
"INFO: 2024/07/20 18:53:16.817627 Files will be read from storages: [default]" for every single call is just too much


reshke (approved):
OK

PR #1764, opened 2024-07-30
fix LoadCustomCABundleError when setting WALG_S3_CA_CERT_FILE

### LINKED ISSUE #1762: Error when using custom CA bundle with AWS S3 in wal-g v3.0.1
### Database name
PostgreSQL

# Issue description

### Describe your problem
We are trying to upgrade the wal-g version used in Zalando's [Spilo](https://github.palantir.build/oss-forks/zalando-spilo) to v3.0.1 (v3.0.2 has the same issue and v3.0.0 works fine). However, we encountered an error related to the HTTPClient's transport type when using a custom CA bundle with AWS S3. In our setup, we specify the path to a custom CA bundle file by setting the `WALG_S3_CA_CERT_FILE` environment variable. This is necessary for us as we are using a private S3-compatible storage service with a self-signed SSL certificate.

The error seems to be caused by the addition of the following line in `session.go` as part of [PR #1670](https://github.com/wal-g/wal-g/pull/1670) (which is released in v3.0.1):
`awsConfig.HTTPClient.Transport = NewRoundTripperWithLogging(awsConfig.HTTPClient.Transport)`
This line is part of the new logging mechanism for collecting statistics about read/written bytes.

We suspect that the custom transport type (*s3.loggingTransport) is not supported when loading the custom CA bundle (specified by `WALG_S3_CA_CERT_FILE`), causing the `LoadCustomCABundleError`.

Could you please help us investigate this issue? Any guidance on how to resolve it would be greatly appreciated.

1. Set the `WALG_S3_CA_CERT_FILE` environment variable to point to a custom CA bundle file.
2. Upgrade wal-g to version v3.0.1. 
3. Run any wal-g command (for example wal-g backup-list).

```
ERROR: 2024/07/29 18:36:35.390721 configure primary storage: configure storage with prefix "s3://postgres-bucket/spilo/postgres-integration-test-postgres-source-environment/wal/14": create S3 storage: create new AWS session: LoadCustomCABundleError: unable to load custom CA bundle, HTTPClient's transport unsupported type
caused by: unsupported transport, *s3.loggingTransport
```

berler:
### Database name
PostgreSQL

# Pull request description

Passes the custom CA bundle to the AWS sdk before configuring the session to avoid LoadCustomCABundleError.

Also fixes an issue where if you had set `WALG_S3_CA_CERT_FILE`, some other config would be ignored due to the early function return (UseYCSessionToken, EndpointSource, and RequestAdditionalHeaders).

In wal-g 3.0.1 and 3.0.2, using S3 storage and setting the environment variable `WALG_S3_CA_CERT_FILE` would result in
```
LoadCustomCABundleError: unable to load custom CA bundle, HTTPClient's transport unsupported type
caused by: unsupported transport, *s3.loggingTransport
```

Fixes #1762

See description in #1762

### DIFF
--- a/pkg/storages/s3/session.go
+++ b/pkg/storages/s3/session.go
@@ -23,24 +23,24 @@ import (
 )
 
 func createSession(config *Config) (*session.Session, error) {
-	sess, err := session.NewSession()
-	if err != nil {
-		return nil, fmt.Errorf("init new default session: %w", err)
-	}
-
-	err = configureSession(sess, config)
-	if err != nil {
-		return nil, fmt.Errorf("configure session: %w", err)
-	}
-
+	sessOpts := session.Options{}
 	if config.CACertFile != "" {
 		file, err := os.Open(config.CACertFile)
 		if err != nil {
 			return nil, err
 		}
 		defer utility.LoggedClose(file, "S3 CA cert file")
-		sess, err = session.NewSessionWithOptions(session.Options{Config: *sess.Config, CustomCABundle: file})
-		return sess, err
+		sessOpts.CustomCABundle = file
+	}
+
+	sess, err := session.NewSessionWithOptions(sessOpts)
+	if err != nil {
+		return nil, fmt.Errorf("init new session: %w", err)
+	}
+
+	err = configureSession(sess, config)
+	if err != nil {
+		return nil, fmt.Errorf("configure session: %w", err)
 	}
 
 	if config.UseYCSessionToken != "" {


x4m (approved):

x4m:
Thanks for the fix! Let's see how tests will react and merge it.
@annielzy can you please confirm that this solves your problem?

annielzy:
@x4m  I have tested it on my end and it indeed resolves the issue. Thanks for the fix @berler !

x4m:
Thanks @annielzy and @berler !

PR #1769, opened 2024-08-08
Fix docker compose in func tests

zykovd:
Deprecating docker compose v1 from func tests


zykovd:
closed as duplicate #1765

PR #1797, opened 2024-10-02
MongoDB: Support restore of sharded cluster with PiTR

Apelsin234:
### Database name

# Pull request description

### Describe what this PR fix
// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/internal/databases/mongo/oplog/applier.go
+++ b/internal/databases/mongo/oplog/applier.go
@@ -90,7 +90,7 @@ func (ap *DBApplier) Apply(ctx context.Context, opr models.Oplog) error {
 		return fmt.Errorf("can not unmarshal oplog entry: %w", err)
 	}
 
-	if err := ap.shouldSkip(op.Operation, op.Namespace); err != nil {
+	if err := ap.shouldSkip(&op); err != nil {
 		tracelog.DebugLogger.Printf("skipping op %+v due to: %+v", op, err)
 		return nil
 	}
@@ -123,13 +123,12 @@ func (ap *DBApplier) Close(ctx context.Context) error {
 	return nil
 }
 
-func (ap *DBApplier) shouldSkip(op, ns string) error {
-	if op == "n" {
+func (ap *DBApplier) shouldSkip(oplog *db.Oplog) error {
+	if oplog.Namespace == "n" {
 		return fmt.Errorf("noop op")
 	}
 
-	// sharded clusters are not supported yet
-	if strings.HasPrefix(ns, "config.") {
+	if !isOpAllowed(oplog) {
 		return fmt.Errorf("config database op")
 	}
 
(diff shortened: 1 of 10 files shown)


internal/databases/mongo/oplog/applier.go
```diff
@@ -123,13 +123,13 @@ func (ap *DBApplier) Close(ctx context.Context) error {
...
-func (ap *DBApplier) shouldSkip(op, ns string) error {
-	if op == "n" {
+func (ap *DBApplier) shouldSkip(oplog *db.Oplog) error {
+	if oplog.Namespace == "n" {
 		return fmt.Errorf("noop op")
 	}
 
-	// sharded clusters are not supported yet
-	if strings.HasPrefix(ns, "config.") {
+	//sharded clusters are not supported yet
```
VGoshev, line 131: PR name says, that we are starting support of restore of sharded clusters, but comment says, that we don't support it yet. Looks a bit weird

internal/databases/mongo/oplog/applier.go
```diff
@@ -156,6 +156,67 @@ func (ap *DBApplier) shouldIgnore(op string, err error) bool {
...
+	if !ok {
+		return true // OK: not a "config" database. allow any ops
+	}
+
+	if Contains(ConfigCollectionsToKeep, coll) {
+		return true // OK: create/update/delete a doc
+	}
+
+	if coll != "$cmd" || len(oe.Object) == 0 {
+		return false // other collection is not allowed
```
VGoshev, line 205: "collection is" -> "collections are"

internal/databases/mongo/oplog/applier.go
```diff
@@ -123,13 +123,13 @@ func (ap *DBApplier) Close(ctx context.Context) error {
...
+func (ap *DBApplier) shouldSkip(oplog *db.Oplog) error {
+	if oplog.Namespace == "n" {
 		return fmt.Errorf("noop op")
 	}
 
-	// sharded clusters are not supported yet
-	if strings.HasPrefix(ns, "config.") {
+	//sharded clusters are not supported yet
+	if !isOpAllowed(oplog) {
 		return fmt.Errorf("config database op")
```
VGoshev, line 133: Need more specific message, may be isOpAllowed(oplog) should return such message.
We need it to being able to understand, why some operation was skipped during debug

VGoshev (changes requested):
Added some tiny issues. Also need to fix tests.
Also it would be good to add test for new feature, so We'll not miss moment, when someone'll brake it

internal/databases/mongo/oplog/applier.go
Apelsin234: fixed

internal/databases/mongo/oplog/applier.go
Apelsin234: fixed

VGoshev (approved):

internal/databases/mongo/oplog/applier.go
Apelsin234: It seems to me that this is not necessary, since here we briefly limit the `config.` collections according to the white list . And giving a detail message for each check is too detailed information that will never be useful to us

PR #1726, opened 2024-06-02
Add linters

annetannet:
Add some linters, fixes #1722

PR #1800, opened 2024-10-03
fix: fix slice init length

cuishuang:
### Describe what this PR fix

The intention here should be to initialize a slice with a capacity of  len(ce.specificStorageErrs)   rather than initializing the length of this slice. 

The only demo: https://go.dev/play/p/q1BcVCmvidW

### DIFF
--- a/internal/multistorage/storage.go
+++ b/internal/multistorage/storage.go
@@ -109,7 +109,7 @@ func (ce *CloseError) Add(err error) {
 }
 
 func (ce *CloseError) Error() string {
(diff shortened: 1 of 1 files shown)


ostinru:
Nice catch!
Probably, it should be found by linter...

ostinru:
Hm... we had #1726 with `makezero` linter.

haiderRizvii (approved):

cuishuang:
friendly ping~

x4m (approved):

x4m:
thanks!

PR #1806, opened 2024-10-10
Add JournalSize to backup sentinels

noname0443:
### Database name
MySQL

# Pull request description
It can be useful to know journal size from one backup to another to estimate recovery time or bill user.

### Describe what this PR fix
Add JournalSize in sentinels and evaluate the sum of binlog size in the semi interval (time of the previous backup; time of new the backup]. If a backup falls within the specified interval, it will be taken into account. Binlog size sum is written to the previous backup sentinel.

### DIFF
--- a/internal/databases/mysql/backup_push_handler.go
+++ b/internal/databases/mysql/backup_push_handler.go
@@ -32,10 +32,11 @@ func HandleBackupPush(
 		tracelog.WarningLogger.Printf("Failed to obtain the OS hostname")
 	}
 
-	lastSentinel, err := getLastUploadedBackupSentinel(folder)
+	latestSentinelName, latestSentinel, err := getLastUploadedBackupSentinel(folder)
 	if err != nil {
 		tracelog.ErrorLogger.Printf("failed to find the last backup: %v", err)
 	}
+	tracelog.InfoLogger.Printf("latest sentinel is %+v", latestSentinelName)
 
 	db, err := getMySQLConnection()
 	tracelog.ErrorLogger.FatalOnError(err)
@@ -75,27 +76,6 @@ func HandleBackupPush(
 	tracelog.ErrorLogger.FatalfOnError("failed to get last uploaded binlog (after): %v", err)
 	timeStop := utility.TimeNowCrossPlatformLocal()
 
-	if lastSentinel != nil {
-		err = internal.AddJournalSizeToPreviousBackup(
-			folder,
-			BinlogPath,
-			utility.BaseBackupPath,
-			lastSentinel.GetName(),
-			func(sentinel map[string]interface{}) (firstBackupJournal string, lastBackupJournal string) {
-				firstBackupJournal = sentinel["BinLogEnd"].(string)
-				lastBackupJournal = binlogEnd
-				tracelog.InfoLogger.Printf("We take in account binlogs in the semi interval (%s;%s]", firstBackupJournal, lastBackupJournal)
-				return firstBackupJournal, lastBackupJournal
-			},
-			func(a, b string) bool {
-				return a < b
-			},
-		)
-		if err != nil {
-			tracelog.ErrorLogger.Printf("Failed to push journal size to the previous backup: %v", err)
-		}
-	}
-
 	uploadedSize, err := uploader.UploadedDataSize()
 	if err != nil {
 		tracelog.ErrorLogger.Printf("Failed to calc uploaded data size: %v", err)
@@ -125,7 +105,6 @@ func HandleBackupPush(
 		BinLogEnd:         binlogEnd,
 		StartLocalTime:    timeStart,
 		StopLocalTime:     timeStop,
-		JournalSize:       0,
 		CompressedSize:    uploadedSize,
 		UncompressedSize:  rawSize,
 		Hostname:          hostname,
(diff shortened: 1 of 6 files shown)


internal/journal.go
```diff
@@ -0,0 +1,89 @@
...
+		return 0, err
+	}
+	if len(journalFiles) == 0 {
+		return 0, nil
+	}
+
+	sum := int64(0)
+	for i := 0; i < len(journalFiles); i++ {
+		jt := journalFiles[i].GetLastModified()
+		if start.Before(jt) && (end.After(jt) || end.Equal(jt)) {
```
debebantur, line 82: I believe that checking time here can lead to some unpredictable situations (files can be stored on host and then archived all at the same time, or if they were copied from somewhere). Is there any lsn in mysql, perhaps?
Or maybe that is acceptable @teem0n

noname0443: You are right. It seems not reliable, so I think we can use `BinLogEnd` from the last sentinel to determine binlogs we should take into account. In other words, we should sum binlogs in the semi interval (pervious BinLogEnd; current BinLogEnd].

internal/journal.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	sentinel[JournalSize] = journalSize
+	sentinelJSON, err := json.Marshal(sentinel)
+	if err != nil {
+		return err
+	}
+
+	r := bytes.NewReader(sentinelJSON)
+	err = folder.PutObject(sentinelName, r)
```
ostinru, line 60: I suspect that this will update sentinel's `LastModified` object's timestamp... and then it will break all our logic of sorting / selecting / handling backups.

e.g. getLastUploadedBackupSentinel() will always return one backup, isn't it?

noname0443: Yes, I agree with you. I think we can use similar logic to MongoDB: create a file "backup.json" with information about backup sizes, binlog sizes and etc., and just read/update it instead of sentinels.

noname0443:
I think I should clarify my intentions about this PR.
- My goal is to add a unified `JournalSize` field. This field could be used to check the size of the WALs/oplogs/binlogs that should be applied to reach the next backup. Additionally, it could be used to bill users for the space occupied by the journals.
- `JournalSize`s will only be evaluated for files, not for LSN/GTID, as this is a simpler approach. I believe this precision will be sufficient.
- `GetJournalSizeInSemiInterval` utilizes the `journalCmpLess func(a, b string) bool`. This function compares two journal names to identify the journals in the semi-interval. I believe this will suffice, given that MySQL journals increment monotonically, which I presume holds true for other databases as well.
- I have implemented this functionality for MySQL as an example. If my concept is well-received and effective, we could reuse this code to other databases.
- We should not count journals for permanent backups. Journals are not permanent and there's no guarantee that they will be preserved until we need to recover from a permanent backup.
- I added backups.json to store backup metrics, journal sizes, and improve the performance of fetching the last backup's metrics. We should not include permanent backups in our tally, so we should either list and read all sentinels to filter whether they are permanent or not, or we could read just one file, backups.json, which appears to be a more efficient solution.
- I am not processing the use case where a backup is deleted mid-sequence. In such cases, the journals will likely be incorrect and smaller than expected. I plan to submit two PRs: this one addressing the concept of using journals and another handling the remaining logic to delete backup information.

I would appreciate any comments and help to identify details I might have missed in my code.

internal/databases/mysql/backup_push_handler.go
```diff
@@ -119,6 +119,42 @@ func HandleBackupPush(
...
+		JournalEnd:       binlogEnd,
+		JournalSize:      0,
+		CompressedSize:   uploadedSize,
+		UncompressedSize: rawSize,
+		IsPermanent:      isPermanent,
+		StopLocalTime:    timeStop,
+	}
+
+	// permanent backups can live longer than binlogs; they should not take part in binlog counting
+	if !isPermanent && err == nil {
```
ostinru, line 141: Is there a reason not to update JournalSizes after making permanent backup?

internal/journal.go
```diff
@@ -0,0 +1,227 @@
...
+	ok, err := folder.Exists(BackupsInfo)
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		tracelog.InfoLogger.Printf("can not find backups.json, creating it...")
+		err := folder.PutObject(BackupsInfo, bytes.NewBuffer([]byte("{}")))
+		if err != nil {
+			return nil, err
+		}
```
ostinru, line 153: It seems that you can skip empty json creation - just return empty map to caller.

noname0443:
@ostinru

I agree with you. Your approach seems to solve the problem with the race conditions, and I think it should be better than the approach with sentinels we had before. We had these problems:
1. We had to list and read all backups to filter out permanent ones.
2. We could not rely on a `LastModified` field of a S3 object and had to read all backups to find the last one.
And, correct me if I'm wrong, with the separate entity `backup.json`, we can solve these issues:
1. Do not create `backup.json` for permanent backups at all.
2. Update previous `backup.json` and create current in the right order. It will not disrupt the backup logic of sorting, etc.

On first glance, it looks fine for me. I will try your approach. Thank you!

internal/journal.go
noname0443: I think we could add logging to the error section and retries for the S3 functions like PutObject, ReadObject, and ListFolder.

internal/journal.go
```diff
@@ -0,0 +1,170 @@
...
+	newJournalEnd string,
+) error {
+	lastSentinelName, lastSentinel, err := GetLastBackupInfo(folder)
+	if err != nil {
+		return err
+	}
+	lastSentinelName = strings.ReplaceAll(lastSentinelName, JournalPrefix, "")
+
+	if len(lastSentinelName) == 0 {
+		tracelog.WarningLogger.Printf("last sentinel was not found, we can not evaluate journal size")
```
debebantur, line 101: Are we really unable to evaluate? Can`t we just count all existing journal files?

debebantur:
I have a question. What will happen if we create some non-permanent backups (a, b, c) with some journal sizes (a_size, b_size, c_size). And then we delete backup in the middle (b). What should happen with sizes? Do we need to rewrite c_size as c_size + b_size?

noname0443:
If I understand you correctly, "a" is the oldest one, and "c" is the newest one. The journal size represents the size of the binlogs/wals/oplogs that need to be applied to the current backup in order to form the next one. In case we delete backup "b", the size of backup "a" must be altered: `a_size = a_size + b_size`.

internal/journal.go
debebantur: Realised that binlog without previous backup is meaningless

debebantur (approved):
LGTM

noname0443:
The new approach encounters a problem with the `ListFolder` method of S3. This operation can become expensive if we have many instances of wal-g running simultaneously. The operation is used in two steps: when wal-g is trying to find the last non-permanent backup and when wal-g is searching binlogs in a semi-interval. 

An approach with a single `backups.json` file does not have this weakness during the stage of searching for the last non-permanent backup. The task of searching binlogs in a semi-interval can potentially be resolved by the DBMS. Specifically, PostgreSQL can use LSN to determine the `JournalSize`. 

In other words, in theory, PostgreSQL can forego using the `ListFolder` and instead use other methods which can improve performance. However, the approach with `backups.json` still faces issues with concurrent backup creation and deletion operations.

noname0443:
I must admit, I am not particularly fond of my own solution. Prior to my PR, there were N files for N backups. However, after my PR, there will be 2*N + 1 files for N backups. In the last commit, I used the `journal_last` file in S3 to keep track of the last non-permanent backup name, thereby avoiding the need to run `ListFolder` another time. I believe that this will lessen the load on S3 from wal-g's side when there are many simultaneous instances running.

ostinru (approved):
> The new approach encounters a problem with the `ListFolder` method of S3. This operation can become expensive if we have many instances of wal-g running simultaneously.
 
I don't see an issue with backup listing - it is quite cheap operation.
I like how this PR looks at the point @debebantur approved it: the simpler - the better.

PS: it is disabled by default - so we can merge it and check how it works in real life.

noname0443:
Okay, then I suggest to merge it and check how it will work. If we encounter problems with `ListFolder`, we can revert to our previous algorithm or consider devising a more sophisticated one.

PR #1823, opened 2024-10-29
Upgrade pgproto3 to 2.3.3

annielzy:
Upgrade pgproto3 and pglogrepl to resolve: [GHSA-7jwh-3vrq-q3m8](https://github.com/jackc/pgproto3/security/advisories/GHSA-7jwh-3vrq-q3m8)


berler:
I don't think you can upgrade pgproto3 by itself since this version is documented as being compatible with pgx v4, and it currently seems to use pgx v3.

annielzy:
> I don't think you can upgrade pgproto3 by itself since this version is documented as being compatible with pgx v4, and it currently seems to use pgx v3.

Yeah you are right just want to see what the test errors are if we just upgrade pgproto3. Closing this PR since it is blocked on pgx major version upgrade.

PR #1826, opened 2024-11-01
Added working with S3 versions

debebantur:
Wal-g supports setting locks for s3 objects, but it does not manage s3 versions.


debebantur:
Closed in favor of https://github.com/wal-g/wal-g/pull/1900

PR #1829, opened 2024-11-07
Add option to disable 100 continue in s3

fepape-ionos:
### Database name
Not related to specific Database

# Pull request description

### Describe what this PR fix
We ran into an issue where our S3 compatible block storage provider corrupted files when http 100 continue messages were used. Since awsConfig allows disabling this behavior we added a flag to add this to the awsConfig to mitigate this issue.


x4m (approved):

x4m:
Mongodb tests seem flacky

PR #1832, opened 2024-11-12
fix: don't ignore libsodiumKeyPath

fepape-ionos:
### Database name
Postgres, but should not be database specific.

# Pull request description
Change `libsodiumKeyPath` to be derived from the correct setting.

### Describe what this PR fix
There was a typo introduced in #1547 which lead to the `LibsodiumKeyPathSetting` not being read in `ConfigureCrypterForSpecificConfig`. As such `isLibsodium` would not be set if only the key's path was given, no Crypter would be set and no encryption/decryption being used.

This occurred after upgrading from wal-g 2.0.1 installation with `WALG_LIBSODIUM_KEY_PATH` to wal-g 3.0.3 and lead to lz4 errors because the files had incorrect file magic (because decryption was not performed)
New instances in this environment would not encrypt the backups at all.


x4m (approved):

x4m:
Thanks!

PR #1836, opened 2024-11-19
fix IsDirectoryEmpty error

NotAmigo:
### Database name
MongoDB

# Pull request description

### Describe what this PR fix
Prolem is that MongoDB datadir used whitelist regexp, but IsDirectoryEmpty didn't metioned it

Restore MongoDB

<p>
```bash
ERROR: 2024/11/02 15:39:46.576597 directory '/var/lib/mongodb' should be empty
```
</p>
</details>


internal/databases/mongo/binary/storage_local.go
```diff
@@ -70,12 +68,13 @@ func (localStorage *LocalStorage) CleanupMongodDBPath() error {
 		return nil
 	}
 	for _, name := range names {
-		if deletionProtectionWhitelist.MatchString(name) {
+		fullPath := filepath.Join(localStorage.MongodDBPath, name)
+		if localStorage.whitelist != nil && localStorage.whitelist.MatchString(fullPath) {
```
Apelsin234, line 72: checking fullPath here, but in line 103 checking name

internal/databases/mongo/binary/storage_local.go
```diff
@@ -101,21 +100,21 @@ func (localStorage *LocalStorage) EnsureEmptyDBPath() error {
...
+	filesRegexp, ok := conf.GetSetting(conf.MongoDBDeletionProtectionWhitelist)
+	_, err := regexp.Compile(filesRegexp)
 
 	if !ok || err != nil {
-		return regexp.MustCompile(`^lost\+found$`)
+		filesRegexp = `lost\+found`
 	}
 
-	return exp
+	return regexp.MustCompile(filepath.Join(regexp.QuoteMeta(dirPath), filesRegexp))
```
Apelsin234, line 119: for me it bad decision , add parent folder like prefix 
1) this is not obvious and the user can add at the beginning ^ , then everything breaks down
2) can't realise reason why need it prefix

Apelsin234 (changes requested):

internal/databases/mongo/binary/storage_local.go
NotAmigo: So, should we declare full path regexp here? Or should we declare relative path and use regexp.MatchString? 
First one would be too large to be simple imho and second one also could break down easily with ^

Apelsin234: I guess we need to check, as before, just the name(suffix after directory path) . You changed this logic because of IsDirectoryEmpty ,  but we can cut prefix before check with whiteListRegexp.
`relativePath := strings.TrimPrefix(strings.Cut(path, directoryPath), '/') ` after this check `relativePath`

Apelsin234 (approved):

PR #1846, opened 2024-11-22
MySQL: add SpaceIDCollector

ostinru:
xtrabackup uses SpaceID to locate original files to which we should apply deltas (File names could change to reflect table renames).
SpaceIDCollector is a small helper that can collect SpaceID to Path mapping and return it to us.


noname0443 (changes requested):

noname0443 (approved):
LGTM

teem0n (approved):

Issue #1850, opened 2024-12-02
Reduce unnecessary warnings for Non-Physical Objects during PostgreSQL wal-g backup


robertmu:
### Database version
PostgreSQL 14.9 on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0, 64-bit.

### wal-g Version
wal-g version 	55d0801a	2024.12.02_13:40:49	PostgreSQL

# Issue description

### Describe your problem
When performing a backup with wal-g, I encountered numerous warnings related to `relfilenode` conversion failures. These warnings are generated for Non-Physical Objects that inherently do not have a `relfilenode` (no physical storage), leading to conversion failures. The warnings seem unnecessary and clutter the log output.
![image](https://github.com/user-attachments/assets/5d7f2f4a-cde8-4f66-83c8-f957ac3ea0c7)

### Please provide steps to reproduce
* Execute the following command
``` bash
wal-g backup-push /path/to/database
```

* You will observe warning messages in the logs similar to
```
WARN: Failed to get relfilenode for relation test_view: strconv.ParseUint: parsing "": invalid syntax
```

ostinru:
Hi, @robertmu !

It seems that you have observed the same issue as in https://github.com/wal-g/wal-g/pull/1816#issuecomment-2460746524

You can safely ignore it.

Fix here: https://github.com/wal-g/wal-g/pull/1849

ostinru:
Ah! You sent a PR: https://github.com/wal-g/wal-g/pull/1851

PR #1851, opened 2024-12-02
postgres: refactor backup to skip non-physical storage objects (#1850)

robertmu:
Proposed Solution

The solution involves improving how wal-g identifies and handles different types of database objects during backup:
1. Enhanced Object Type Detection
  * Added relkind to the database queries to properly identify different types of relations (tables, views, etc.)
  * This allows wal-g to make informed decisions about how to handle each object type

2. Processing Logic
  * For objects with relfilenode = 0, implement a two-step verification:
    
    First, check if the path is empty (indicating no physical storage)
    If path is empty, skip the object and log a debug message instead of a warning
    If path exists, attempt to parse the actual relfilenode from the path (for system catalogs)

3. Special Handling for Partitioned Tables
 * Explicitly handle partitioned tables (relkind = 'p') as a special case
 * Process them for metadata purposes while acknowledging they don't have physical storage
 * This ensures proper handling of partition hierarchies while avoiding unnecessary warnings
 
This solution maintains all necessary backup functionality while reducing log clutter by eliminating warnings for objects that naturally don't have physical storage.

### DIFF
--- a/internal/databases/postgres/query_runner.go
+++ b/internal/databases/postgres/query_runner.go
@@ -568,7 +568,7 @@ func (queryRunner *PgQueryRunner) GetLockingPID() (int, error) {
 func (queryRunner *PgQueryRunner) BuildGetTablesQuery(getPartitioned bool) (string, error) {
 	switch {
 	case queryRunner.Version >= 90000:
-		query := "SELECT c.relfilenode, c.oid, pg_relation_filepath(c.oid), c.relname, pg_namespace.nspname FROM pg_class " +
+		query := "SELECT c.relfilenode, c.oid, pg_relation_filepath(c.oid), c.relname, pg_namespace.nspname, c.relkind FROM pg_class " +
 			"AS c JOIN pg_namespace ON c.relnamespace = pg_namespace.oid " +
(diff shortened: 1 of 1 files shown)


internal/databases/postgres/query_runner.go
```diff
@@ -684,14 +684,40 @@ func (queryRunner *PgQueryRunner) processTables(conn *pgx.Conn,
...
+
+			// Case 2: Non-empty path with relfilenode=0 indicates a mapped catalog
+			// This happens for:
+			// pg_class itself and other critical system catalogs, shared catalogs
+			// These tables use a separate mapping file to track their actual file locations
 			parts := strings.Split(path.String, "/")
 			chis, err := strconv.ParseUint(parts[len(parts)-1], 10, 32)
 			if err != nil {
-				tracelog.DebugLogger.Printf("Failed to get relfilenode for relation %s\n", tableName)
+				tracelog.WarningLogger.Printf("Failed to get relfilenode for relation %s: %v\n", tableName, err)
```
debebantur, line 719: I believe, here we can skip logging actual error, since it will be not really informative. Also lets change log level to Debug

robertmu: Done, thanks for review. I have updated the log level to Debug and removed unnecessary error messages as suggested.

debebantur (approved):

PR #1854, opened 2024-12-04
release: unify amd64 <--> arm64 names

### LINKED ISSUE #1853: Unified binary names
### Database name
PostgreSQL

# Issue description

The published binaries for AMD64 and ARM64 have different format of filenames. For example, `wal-g-pg-ubuntu20.04-aarch64` has no dash after "ubuntu" while `wal-g-pg-ubuntu-20.04-amd64` does.

### Describe your problem
This difference is a bit inconvenient when I use Ansible to install wal-g binaries to hosts with multiple architectures. My suggestion is that only the arch indicator should change and no other part of the filename, it could be helpful if I could use only `{{ ansible_architecture }}` to select between binaries.

chipitsine:
fixes #1853

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,6 +73,7 @@ jobs:
         include:
           - arch: aarch64
             distro: ubuntu20.04
+            label: ubuntu-20.04
       fail-fast: false
 
     steps:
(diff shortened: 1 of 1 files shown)


chipitsine:
@stl-victor-sudakov fyi

stl-victor-sudakov:
Wow, great, thanks!

chipitsine:
Ping

stl-victor-sudakov:
Pong

x4m:
let's merge?

chipitsine:
> let's merge?
> 

Yes

x4m (approved):

PR #1856, opened 2024-12-06
Fix wal-g copy for pg

munakoiso:


debebantur (approved):

debebantur:
Can you please change without-history flag [here](https://github.com/wal-g/wal-g/blob/fade5e6d54832e584ea8cff0692672ad50f1f7ae/docker/pg_tests/scripts/tests/copy_backup_test.sh#L43C86-L43C101) and [here](https://github.com/wal-g/wal-g/blob/master/docs/PostgreSQL.md#copy)

cmd/pg/copy.go
```diff
@@ -57,11 +57,11 @@ func init() {
...
 	backupCopyCmd.Flags().StringVarP(&fromConfigFile, fromFlag, fromShorthand, "", fromDescription)
-	backupCopyCmd.Flags().BoolVarP(&withoutHistory,
-		withoutHistoryFlag,
-		withoutHistoryShorthand,
+	backupCopyCmd.Flags().BoolVarP(&withAllHistory,
+		withAllHistoryFlag,
+		withAllHistoryShorthand,
 		false,
-		withoutHistoryDescription)
+		withAllHistoryDescription)
```
ostinru, line 64: it is backward incompatible changes... with changed defaults.
Is there real reason to do so?

munakoiso: Backup without any WALs will not work anyway. So there is no chance that someone has used this code before (and then was able to restore the backup).

munakoiso: In addition, there was an error without the "without history" flag - the function copied only from finish_lsn. Therefore, backups from walg copy were useless

PR #1861, opened 2024-12-13
Fixed backward compatibility with files metadata

debebantur:
### Database name
Postgres/Greenplum.

# Pull request description
There was fixes for partial restore which led to change in the structure of files_metadata. This pr makes this changes backward compatible


debebantur:
Too complex solution, since old table info is actually useless

Issue #1871, opened 2024-12-22
PostgreSQL: Two Related Issues in WAL Part File Handling


robertmu:
We have identified two related issues in the WAL delta backup system regarding WAL part file handling:

### Issue 1: Incomplete First Running WAL Part File

Our WAL delta tracking system uses the algorithm `logSegNo % WalFileInDelta` to map WAL files to WAL part files. This creates an inherent limitation:

- When the first WAL file's `logSegNo` is not 0, the earlier parts in the WAL part file will be missing
- For example, if the first WAL file's `logSegNo` is 5, then WAL parts 0-4 will be missing in the part file

### Issue 2: WAL Part File Data Loss During Backup History File Processing

There is a potential data loss issue in `HandleWALPush`(wal_push_handler.go) when processing backup history files:

The `uploader.FlushFiles` method assumes that we have already read the WAL part files into memory before calling it. Its first step is to delete the WAL part files from disk, then write the latest in-memory WAL part file back to disk. However, when handling a backup history file, it will not trigger the action of reading WAL part files into memory. This causes:

1. The WAL part file becomes incomplete (see `WalPartFile.IsComplete()`)
2. Due to the incomplete WAL parts, the WAL part file can never be marked as completed, and its corresponding delta file will not be uploaded
3. When performing delta backup, since the required delta file is missing, it will fail to generate delta map and fall back to full backup

## Environment
- WAL-G version: [master branch]
- PostgreSQL version: [14]

x4m:
Hi! That's very interesting. Are you reffering to WAL_DELTA backups?

robertmu:
Yes, we are discussing WAL_DELTA backups.

x4m:
We have a long-standing bug in WAL DELTA https://github.com/wal-g/wal-g/issues/1296. If you fix it - we will be super grateful to you. We could not trace it during 2019-2020 and then just disabled the feature.
Currentl, only usual deltas are used.
I did not dig into PR details. But second part seems to be something wal-push related.

robertmu:
I have three commits to fix delta map based delta backup. The first PR is https://github.com/wal-g/wal-g/pull/1861, and this is the second PR. Based on my testing, when all three PRs are merged, delta map based delta backup works correctly.

robertmu:
I'm going to bed now. Tomorrow I'll check if this issue is related to #1296.

x4m:
Thank you so much!

robertmu:
[...]

PR #1872, opened 2024-12-22
fix(postgres): handle WAL delta backup edge cases #1871

robertmu:
This commit addresses two related issues in WAL delta backup:

1. First Running WAL Part File Incompleteness
- Add IsFirstRunningPartFile() to detect first running part file
- Add CompleteFirstRunningPartFile() to fill missing parts with empty slices
- Handle cases where logSegNo doesn't start from 0

2. WAL Part File Data Loss During Backup History Processing
- Add HandleBackupHistoryFile() to process backup history files
- Ensure proper ordering of operations in HandleWALPush
- Prevent data loss when flushing WAL part files

Added:
- backup_history.go: Functions to handle backup history files
- Tests for backup history and WAL part file functionality

This ensures WAL delta backups work correctly even when:
- First WAL file's logSegNo is not 0
- Processing backup history files


x4m:
@fanfuxiaoran Can I ask you to help reviewing this PR?

fanfuxiaoran:
> @fanfuxiaoran Can I ask you to help reviewing this PR?

Sure， this pr is to fix the same issue as my pr https://github.com/wal-g/wal-g/pull/1938 - to resolve missing deltal file when doing backup, but use the different methods.  I will compare them in details.

fanfuxiaoran:
For this pr, what I concerned is that it changes the part delta file in `CompletePartFile`.  Consider the following case:
```
1. Do a base backup backup1
2. Generate delta file and upload wal log 
3. But wal log from  xxxxxx01 and  xxxxxx02, due some reason , they cannot be uploaded by wal-g and part delta file not generated
4. Do a base backup backup2 ( start wal log is xxxxxx03)
5. Do a delta backup from backup2, this step will fix the partFilledDeltaFile by `CompletePartFile` , then wal-g can generate delta file for it.
6. Do a delta backup from backup1, this step may use the fixed delta file and success. But it is supposed to fail as the delta file for xxxxxx01 and xxxxxx02 is missing.
```

Maybe we need to add more info in the part delta file  and delta file to mark which segment wal file has been parsed and recorded in it and let the base backup part to decided if a delta file can used for it.

robertmu:
Good catch! However, I want to point out that in the scenario you described, the issue would actually fail at step 1. Here's why:

When WAL segments xxxxxx01 and xxxxxx02 fail to upload, the base backup (backup1) itself would fail to complete successfully. This is because a base backup requires all WAL segments generated during the backup process to be successfully archived. Without these WAL segments, the backup would be considered incomplete and invalid.

So while your concern about the delta file handling is valid from a theoretical perspective, in practice, the scenario wouldn't reach the point where we need to worry about the delta file being incorrectly completed, as the initial base backup (backup1) would have already failed.

That said, it's still valuable to think about improving the robustness of our delta file handling mechanism. Thank you for bringing this up for discussion!

fanfuxiaoran:
> When WAL segments xxxxxx01 and xxxxxx02 fail to upload, the base backup (backup1) itself would fail to complete successfully. This is because a base backup requires all WAL segments generated during the backup process to be successfully archived. Without these WAL segments, the backup would be considered incomplete and invalid.

If I understand correctly, wal log is sent by 'archive_command', even if the backup(backup1) is finished, the 'archive_command' will still run. I mean  xxxxxx01 and xxxxxx02 are the wal logs after the backup1 finished.

robertmu:
I believe this scenario you described should not exist. If backup1 is marked as successful but WAL segments xxxxxx01 and xxxxxx02 are still being archived, it indicates that backup1 has issues and cannot be considered a safe backup

fanfuxiaoran:
> I believe this scenario you described should not exist. If backup1 is marked as successful but WAL segments xxxxxx01 and xxxxxx02 are still being archived, it indicates that backup1 has issues and cannot be considered a safe backup


For example:
1: backup1 starts
2. backup1 finishes and successes at the wal log '000000010000001F0000006F'
3. archive_command still works to send the wal log to s3
4. arhive_command fails to send wal log '000000010000001F00000080' and '000000010000001F00000081' to s3
5. archive_command recovers and successes to send wal log after '000000010000001F00000081'
6. backup2 starts at wal log '000000010000001F00000083'
7. delta backup starts from backup2,  will  fix the delfa file '000000010000001F00000080.delta'
8. If we start a delta backup from backup1, will use the wrong delta file '000000010000001F00000080.delta'

robertmu:
1.  
   It seems that you consider the example to be inappropriate. When a PostgreSQL database is started for the first time, no matter how many times backups are taken afterward, the WAL part file that needs to be “completed” is always the **first** WAL part file. So, in the example you gave, where `backup1` is completed within `000000010000001F0000006F`, it's actually impossible for us to "complete" `000000010000001F00000080_part` (delta).  

   The reason is:  
   The WAL segments corresponding to `000000010000001F00000080_part` (delta) span from `000000010000001F00000080` to `000000010000001F0000008F`. This file doesn't require "complete" because PostgreSQL, during the archiving process, has already invoked the `wal-g` command **for each of the 16 files in that WAL segment range**, and each of those invocations has contributed its corresponding part to the WAL part file, which has been saved properly.

2.  
   Based on my understanding of PostgreSQL's archiving mechanism—which is strictly sequential and serial—and the fact that it doesn't skip failed log files, your scenario involving `000000010000001F00000080` and `000000010000001F00000081` is invalid.  

   When the archiving of `000000010000001F00000080` fails, the archiver will **stop** and **retry** that file. It will **not** proceed to `000000010000001F00000081` until the previous WAL file is archived successfully.  

   Therefore, when the logic to 'complete' the WAL part file is executed—such as in the delta-backup from `backup2` (i.e., triggered by the backup history file corresponding to `delta-backup from backup2`)—the content for `000000010000001F00000080` and `000000010000001F00000081` has **already been generated and stored** within the WAL part file, so it is impossible for an incorrect delta file to be generated.

PR #1874, opened 2024-12-24
Add JournalInfo deletion logic

noname0443:
### Database name
MySQL

# Pull request description

Corresponding to the comments in my previous PR (#1806), I create the PR about the deletion logic of `JournalInfo`.

There is a thing that we should notice. `JournalInfo` of a backup contains information about binlogs we can use for PiTR recovery. So when the oldest backup is deleted, we can't use PiTR with the binlogs from the journal, and consequently the journal can be safely deleted without any calculations. On another hand, if we delete a backup in the middle, we need to take actions in a particular way:
1. Assume the beginning of the binlog semi-interval of the next journal to the beginning of the binlog semi-interval of the deleting `JournalInfo`. The gap between backups was increased after deletion, so we should take it into account for the next `JournalInfo`.
2. Delete the `JournalInfo` of the backup.
3. Calculate the journals size of the semi-interval and assume the new value for the previous `JournalInfo` relate to the deleted one.

We have 4 deletion methods in wal-g, and according to the information above:
1. Delete everything. It deletes all include journals automatically, so we shouldn't do something.
2. Delete before. It deletes only the oldest backups and automatically deletes journals, so we shouldn't do something.
3. Delete retain. It deletes only the oldest backups and automatically deletes journals, so we shouldn't do something.
4. Delete target. It deletes target backup and it can be a backup in the middle. So we should using the algorithm above.

There are a few another changes:
0. `JournalInfo` now is a structure with methods.
1. I refactor code in a more readable way.
2. I refactor tests in a more readable way.

### DIFF
--- a/cmd/mysql/delete.go
+++ b/cmd/mysql/delete.go
@@ -80,12 +80,16 @@ func runDeleteTarget(cmd *cobra.Command, args []string) {
 		mysql.BinlogPath,
 		internal.DefaultLessCmp,
 	)
-	tracelog.ErrorLogger.FatalOnError(err)
-	tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
+	if err != nil {
+		tracelog.WarningLogger.Printf("Can't find the journal info: %s", err.Error())
+	} else {
+		tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
+	}
 
 	deleteHandler.HandleDeleteTarget(backupSelector, confirmed, findFullBackup)
 
-	if confirmed {
+	// Backup could be created without journal
+	if confirmed && err == nil {
 		err := journalInfo.Delete(storage.RootFolder())
 		tracelog.ErrorLogger.PrintOnError(err)
 	}
--- a/cmd/mysql/delete.go
+++ b/cmd/mysql/delete.go
@@ -80,7 +80,6 @@ func runDeleteTarget(cmd *cobra.Command, args []string) {
 		backupName,
 		storage.RootFolder(),
 		mysql.BinlogPath,
-		mysql.BinlogFilenameComparator,
 	)
 	// Backup could be created without journal
 	if err != nil {
--- a/cmd/mysql/delete.go
+++ b/cmd/mysql/delete.go
@@ -80,7 +80,7 @@ func runDeleteTarget(cmd *cobra.Command, args []string) {
 		backupName,
 		storage.RootFolder(),
 		mysql.BinlogPath,
-		internal.DefaultLessCmp,
+		mysql.BinlogFilenameComparator,
 	)
 	// Backup could be created without journal
 	if err != nil {
--- a/cmd/mysql/delete.go
+++ b/cmd/mysql/delete.go
@@ -74,24 +74,30 @@ func runDeleteTarget(cmd *cobra.Command, args []string) {
 	backupSelector, err := internal.NewBackupNameSelector(backupName, true) //todo: add selection by userdata
 	tracelog.ErrorLogger.PrintOnError(err)
 
+	deleteHandler.HandleDeleteTarget(backupSelector, confirmed, findFullBackup)
+
 	journalInfo, err := internal.NewJournalInfo(
 		backupName,
 		storage.RootFolder(),
 		mysql.BinlogPath,
 		internal.DefaultLessCmp,
 	)
+	// Backup could be created without journal
 	if err != nil {
 		tracelog.WarningLogger.Printf("Can't find the journal info: %s", err.Error())
-	} else {
-		tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
+		return
 	}
 
-	deleteHandler.HandleDeleteTarget(backupSelector, confirmed, findFullBackup)
+	if !confirmed {
+		tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
+		return
+	}
 
-	// Backup could be created without journal
-	if confirmed && err == nil {
-		err := journalInfo.Delete(storage.RootFolder())
-		tracelog.ErrorLogger.PrintOnError(err)
+	err = journalInfo.Delete(storage.RootFolder())
+	if err != nil {
+		tracelog.ErrorLogger.Print(err)
+	} else {
+		tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
 	}
 }
 
--- a/cmd/mysql/delete.go
+++ b/cmd/mysql/delete.go
@@ -74,7 +74,21 @@ func runDeleteTarget(cmd *cobra.Command, args []string) {
 	backupSelector, err := internal.NewBackupNameSelector(backupName, true) //todo: add selection by userdata
 	tracelog.ErrorLogger.PrintOnError(err)
 
+	journalInfo, err := internal.NewJournalInfo(
+		backupName,
+		storage.RootFolder(),
+		mysql.BinlogPath,
+		internal.DefaultLessCmp,
+	)
+	tracelog.ErrorLogger.FatalOnError(err)
+	tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
+
 	deleteHandler.HandleDeleteTarget(backupSelector, confirmed, findFullBackup)
+
+	if confirmed {
+		err := journalInfo.Delete(storage.RootFolder())
+		tracelog.ErrorLogger.PrintOnError(err)
+	}
 }
 
 func runDeleteBefore(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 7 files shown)


docker/mysql_tests/scripts/base_tests/journals.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/sh
+set -e -x
+
+. /usr/local/export_common.sh
+
+export WALE_S3_PREFIX=s3://mysqlfullxtrabackupbucket
```
ostinru, line 6: please use your own bucket =) In theory we can run tests concurrently.

Note1: you should add it in wal-g/docker-compose.yaml before use
Note2: new bucket names uses readable 'mysql_xxx' notation instead of messy 'mysqlxxx'

internal/journal.go
```diff
@@ -8,163 +8,304 @@ import (
...
+	if err != nil {
+		return JournalInfo{}, err
+	}
+
+	return backupInfo, nil
+}
+
+// Calculate calculates the size of the JournalInfo in the semi-interval (JournalStart; JournalEnd]
+// using journal files on JournalDirectoryName and save it for the previous JournalInfo
+func (ji *JournalInfo) Calculate(folder storage.Folder) error {
```
ostinru, line 271: 'Calculate' name doesn't tell much about its behaviour =/ Originally I thought that it is pure function.

internal/journal.go
```diff
@@ -8,163 +8,304 @@ import (
...
-	if len(lastSentinelName) == 0 {
-		tracelog.WarningLogger.Printf("last sentinel was not found, we can not evaluate journal size")
-		return nil
+	nextJi, err := ji.GetNext(folder)
+	if err != nil {
+		if err.Error() != cantFindJournal {
+			return err
+		}
+
+		// We should update previous JournalInfo to zero value if there are no JournalInfo after that
```
ostinru, line 200: sorry, haven't got an idea

cmd/mysql/delete.go
```diff
@@ -74,7 +74,25 @@ func runDeleteTarget(cmd *cobra.Command, args []string) {
...
+	journalInfo, err := internal.NewJournalInfo(
+		backupName,
+		storage.RootFolder(),
+		mysql.BinlogPath,
+		internal.DefaultLessCmp,
+	)
+	if err != nil {
+		tracelog.WarningLogger.Printf("Can't find the journal info: %s", err.Error())
+	} else {
+		tracelog.InfoLogger.Printf("Deleted journal info: %+v", journalInfo)
```
ostinru, line 86: Journal info hasn't deleted yet
This message could be confusing

docker/mysql_tests/scripts/base_tests/journals.sh
```diff
@@ -0,0 +1,91 @@
...
+wal-g backup-push --count-journals
+sleep 1
+
+# Check count of backups and content of them
+test "1" -eq $(get_journal_count)
+test "0" -eq $(get_journal_size 1)
+
+# Create backup #2 with journals
+sysbench --time=5 run
+mysql -e 'FLUSH LOGS'
```
ostinru, line 43: 'FLUSH LOGS' works perfectly fine for your test case... so, let's reduce `--time=5` to smaller waits.

internal/journal.go
```diff
@@ -8,163 +8,305 @@ import (
...
-		jt := utility.TrimFileExtension(journalFiles[i].GetName())
+		jf := utility.TrimFileExtension(journalFiles[i].GetName())
 
-		isEqual := !journalCmpLess(jt, end) && !journalCmpLess(end, jt)
+		isEqual := !cmp(jf, ji.JournalEnd) && !cmp(ji.JournalEnd, jf)
 
-		if journalCmpLess(start, jt) && (journalCmpLess(jt, end) || isEqual) {
-			tracelog.InfoLogger.Printf("Found in range: %+v\n", jt)
+		if cmp(ji.JournalStart, jf) && (cmp(jf, ji.JournalEnd) || isEqual) {
+			tracelog.DebugLogger.Printf("Taking into accoutn: %s\n", jf)
```
ostinru, line 289: `accoutn` -> `account`

internal/databases/mysql/backup_push_handler.go
```diff
@@ -132,37 +132,36 @@ func HandleBackupPush(
...
-	err = internal.UploadBackupInfo(folder, backupName, newBackupInfo)
+	err = journalInfo.UpdateIntervalSize(folder)
 	if err != nil {
-		tracelog.ErrorLogger.Printf("failed to upload backup info for %s: %s", backupName, err)
+		tracelog.WarningLogger.Printf("can not calculate journal size: %s", err.Error())
 		return
 	}
-	tracelog.InfoLogger.Printf("uploaded backup info for %s", backupName)
+
+	tracelog.InfoLogger.Printf("uploaded journal info for %s", backupName)
```
ostinru, line 164: This part is mind blowing...

We have following structure:
```
type JournalInfo struct {
...
	JournalStart              string                 `json:"JournalStart"`
	JournalEnd                string                 `json:"JournalEnd"`
	JournalSize               int64                  `json:"JournalSize"`
}
```

We set:
```
JournalStart = (currentBackup-1) binlog position
JournalEnd   = (currentBackup) binlog position
JournalSize  = 0
```

and for previous backup:
```
JournalStart = (currentBackup-2) binlog position
JournalEnd   = (currentBackup-1) binlog position
JournalSize  = SUM of binlog sizes in range (currentBackup-1..currentBackup)
```

so, JournalSize IS NOT size of `JournalStart...JournalEnd` 🤯

internal/journal.go
```diff
@@ -8,163 +8,305 @@ import (
...
 	}
 
 	sum := int64(0)
+	cmp := ji.JournalNameLessComparator
 	for i := 0; i < len(journalFiles); i++ {
-		jt := utility.TrimFileExtension(journalFiles[i].GetName())
+		jf := utility.TrimFileExtension(journalFiles[i].GetName())
 
-		isEqual := !journalCmpLess(jt, end) && !journalCmpLess(end, jt)
+		isEqual := !cmp(jf, ji.JournalEnd) && !cmp(ji.JournalEnd, jf)
```
ostinru, line 286: Is it correct to compare binlogs by name, rather than by upload timestamp?

noname0443: I believe it is a more reliable way to compare them. As was mentioned in the previous PR (https://github.com/wal-g/wal-g/pull/1806#discussion_r1796615611), there could be different situations when we can't rely on uploaded time.

internal/databases/mysql/backup_push_handler.go
noname0443: I agree, so I suggest renaming:
`JournalSize` -> `SizeToNextBackup`,
`JournalStart` -> `PriorBackupEnd`
`JournalEnd` -> `CurrentBackupEnd`
This sounds more appropriate and doesn't form unfulfilled expectations, though it could potentially still be a bit unintuitive... I also suggest not changing the JSON part of the variables to avoid producing a second version of the file format.

internal/journal.go
ostinru: binlog could have any prefix, e.g. users may set `log_bin = mysql-bin-log-FQDN`
Then, after switchover you will count random binlogs rather than range of binlogs.

internal/databases/mysql/backup_push_handler.go
ostinru: If so, please also rename json-fields (it will be easier to debug and reason about).
// note, that this can break backward compatibility...

internal/journal.go
noname0443: I have changed the comment message and done some refactoring. I hope it has become more readable now.

internal/databases/mysql/backup_push_handler.go
noname0443: I have checked compatibility problems. So the last journal info file on the previous version will be broken to:
```
{"PriorBackupEnd":"","CurrentBackupEnd":"","SizeToNextBackup":18708037}
```
It counts all the binlogs in the folder, but... This journal file will be deleted over time, and all normalize, so I suppose we can go for it.

internal/journal.go
noname0443: Yeah, it is so... It seems that we can use the number in the binlog suffix to compare values; parse it and then compare integer values.

ostinru: During PiTR we are applying binlogs in timestamp order, so we can safely assume that all binlogs in between backups should be assigned (counted) as for our backup.

We have almost no control over binlog naming, and binlog suffix doesn't help us to understand where in timeline this binlog was created.

noname0443: Thank you for the explanation! Okay, I'll fix the code to use timestamps.

ostinru (approved):
LGTM

PR #1878, opened 2025-01-02
tests: remove few "sleep"

chipitsine:
speed up gp/cloudberry tests a bit


ostinru (approved):
Let's try it )

chipitsine:
greenplum currently takes ~1hr. too bad for developer velocity

PR #1882, opened 2025-01-04
Refactor: Move CheckArchiveCommand from "connect.go" to queryRunner

boosterKRD:
**Database name**
PostgreSQL

**Pull request description**
The primary goal of this refactor is to eliminate the multiple executions of the CheckArchiveCommand function triggered each time a new PostgreSQL connection was established via the Connect function. By relocating CheckArchiveCommand outside of Connect, we ensure that the archive settings are checked only once

### DIFF
--- a/internal/databases/postgres/backup_push_handler.go
+++ b/internal/databases/postgres/backup_push_handler.go
@@ -173,6 +173,8 @@ func (bh *BackupHandler) createAndPushBackup(ctx context.Context) {
 	tracelog.ErrorLogger.FatalOnError(err)
 	err = bh.checkDataChecksums()
 	tracelog.ErrorLogger.FatalOnError(err)
+	err = bh.CheckArchiveCommand()
+	tracelog.ErrorLogger.FatalOnError(err)
 
 	if orioledbEnabled {
 		chkpNum := orioledb.GetChkpNum(bh.PgInfo.PgDataDirectory)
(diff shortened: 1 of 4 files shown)


internal/databases/postgres/backup_push_handler.go
```diff
@@ -682,6 +684,55 @@ func (bh *BackupHandler) checkDataChecksums() error {
...
+			tracelog.WarningLogger.Println(
+				"archive_command is not configured. This may cause inconsistent backups. " +
+					"Please consider configuring WAL archiving.")
+		} else {
+			tracelog.DebugLogger.Println("WAL archiving settings are configured.")
+		}
+	}
+
+	return nil
+}
```
ostinru, line 734: It seem that after this change we will not show warning for greenplum:
https://github.com/wal-g/wal-g/blob/v3.0.5/internal/databases/greenplum/backup_push_handler.go#L484

boosterKRD: > It seem that after this change we will not show warning for greenplum: https://github.com/wal-g/wal-g/blob/v3.0.5/internal/databases/greenplum/backup_push_handler.go#L484

@ostinru Hi 
I've [added](https://github.com/wal-g/wal-g/pull/1882/commits/a5b0ec22a112dfa243a369e1bf6d191d4d81050d) 'Archive mode' check into Greenplum.

In this commit, I decided to define a separate `func CheckArchiveCommand(conn *pgx.Conn) error` specifically for the Greenplum code. Currently, it references methods from the postgres package. This separation is intended to allow future modifications or overrides of these methods or logic for Greenplum without impacting the postgres implementation.

x4m (approved):

internal/databases/postgres/backup_push_handler.go
boosterKRD: @ostinru Hi could you review my last changes (comment above)?

PR #1896, opened 2025-01-27
pg_tests: try to prevent race condition on background start

chipitsine:
trying to prevent race condition like this one: https://github.com/wal-g/wal-g/actions/runs/12975341294/job/36186448287


chipitsine:
@wal-g/committers  ?

debebantur (approved):

PR #1917, opened 2025-02-20
Fix connection break

debebantur:
### Database name
Postgres/Greenplum

# Pull request description

Sometimes when performing long backups on unstable clusters i get error
```
INFO: 2025/02/18 21:13:53.117534 Calling pg_stop_backup()
ERROR: 2025/02/18 21:13:53.117705 UploadLabelFiles: failed to stop backup: QueryRunner StopBackup: transaction begin failed: conn is dead
```
I believe that it is better to check if connection is alive after processing files. And if it is not, reconnect to database

### DIFF
--- a/internal/databases/postgres/query_runner.go
+++ b/internal/databases/postgres/query_runner.go
@@ -238,6 +238,16 @@ func (queryRunner *PgQueryRunner) StopBackup() (label string, offsetMap string, 
 	defer queryRunner.Mu.Unlock()
 
 	tracelog.InfoLogger.Println("Calling pg_stop_backup()")
+
+	//during long backups conn might break, so we check if connection is still alive
+	errPing := queryRunner.Connection.Ping(context.TODO())
+	if errPing != nil {
+		queryRunner.Connection, err = Connect()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


reshke (approved):

internal/databases/postgres/query_runner.go
```diff
@@ -238,6 +238,16 @@ func (queryRunner *PgQueryRunner) StopBackup() (label string, offsetMap string,
...
+
+	//during long backups connection might break, so we check if connection is still alive
+	errPing := queryRunner.Connection.Ping(context.TODO())
+	if errPing != nil {
+		queryRunner.Connection, err = Connect()
+		if err != nil {
+			return "", "", "", fmt.Errorf("QueryRunner StopBackup: connection is dead: %w %w", errPing, err)
+		}
+	}
+
```
ostinru, line 250: JFYI: this will not work for Cloudberry...

```
* If the client disconnects with a non-exclusive backup running, the backup
is automatically aborted. This is the same thing that pg_basebackup does.
To use these non-exclusive backups the backup software will have to
maintain a persistent connection to the database -- something that should
not be a problem for any of the modern ones out there (but would've been
slightly trickier back in the days when we suggested shellscripts)
```

PR #1924, opened 2025-03-02
Add `wal-g st get-stream` command to fetch backup as a single stream

ostinru:
### MySQL / Mongodb

`wal-g st get-stream` will allow users to fetch stream-like backups as single stream. This command will:
* decrypt backup
* decompress backup
* assemble parts of backups (when `WALG_STREAM_SPLITTER_*` used)
* save to file

### DIFF
--- a/cmd/common/st/get_stream.go
+++ b/cmd/common/st/get_stream.go
@@ -15,7 +15,8 @@ import (
 
 const (
 	getStreamShortDescription = "Download the backup as single stream"
-	getStreamLongDescription  = "Download, decrypt (if encrypted), decompress (if compressed) and assemble (if stream where split) single-stream backups to single stream."
+	getStreamLongDescription  = "Download, decrypt (if encrypted), decompress (if compressed) and assemble " +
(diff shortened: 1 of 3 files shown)


cmd/common/st/get_stream.go
```diff
@@ -0,0 +1,61 @@
+package st
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/wal-g/tracelog"
+
+	"github.com/wal-g/wal-g/internal"
+	"github.com/wal-g/wal-g/internal/databases/mysql"
```
noname0443, line 10: Isn't `get-stream` a mysql specific command? Maybe it would be better to place the file in `cmd/mysql/` instead of `cmd/common`?

ostinru: MongoDB uses split-stream as well

noname0443: Oh, I was just unable to read the description, yeah -__-

noname0443 (approved):

teem0n (approved):

PR #1934, opened 2025-03-10
Fix postgresql tests

debebantur:
### Database name
postgres, mysql

# Pull request description
Some tests started to fail when mysql package was included in common build. Seems that problem is in github.com/go-mysql-org/go-mysql/server package


ostinru (approved):
LGTM

PR #1938, opened 2025-03-18
Fix delta file not exist in getDeltaMap

fanfuxiaoran:
If the first wal file pushed by the wal-g is not started as xxx0 after the archive_mode enabled, and we start the basebackup, then the first delta segment will be missing for that basebackup. So does the last delta file, if the last delta segment is the same as the first delta segment.

The first wal log file after starting the basebackup doesn't contain XLogRecord tail in the head, so we can parse the wal files to extract locations.

Add test 'delta_backup_wal_delta_missing_test.sh' for this commit

### Database name
postgres

# Pull request description

### Describe what this PR fixes
// problem is ...

### Please provide steps to reproduce (if it's a bug)
// it can really help

### Please add config and wal-g stdout/stderr logs for debug purpose

also you can use WALG_LOG_LEVEL=DEVEL for logs collecting
<details><summary>If you can, provide logs</summary>
<p>
```bash
any logs here
```
</p>
</details>

PR #1940, opened 2025-03-20
Add config option WALG_FORCE_WAL_DELTA

fanfuxiaoran:
When WALG_FORBIDDEN_FALLBACK_TO_FULLBACKUP is true that means when delta backup failed, wal-g don't fallback to full scan delta backup. We can use it in delta related tests.


fanfuxiaoran:
Some pg/gp delta tests may fail , #1938 this pr can fix them I think.

internal/config/config.go
```diff
@@ -232,6 +232,9 @@ const (
 	SSHPrivateKeyPath = "SSH_PRIVATE_KEY_PATH"
 
 	SystemdNotifySocket = "NOTIFY_SOCKET"
+
+	// For test
+	ForbiddenFallbackToFullbackup = "FORBIDDEN_FALLBACK_TO_FULLBACKUP"
```
ostinru, line 237: Let's add `WALG_` prefix to this config.

ostinru, line 237: Probably, this feature may be useful in production... Let's add it to documentation (both Pg/Gp).

fanfuxiaoran:
hi @ostinru , I enable `WALG_FORBIDDEN_FALLBACK_TO_FULLBACKUP` in the integration tests, so some delta backup tests failed. We need to fix them in other prs.

x4m:
Can we name this know without double negation? Like "require delta" or "delta only"...

fanfuxiaoran:
> Can we name this knob without double negation? Like "require delta" or "delta only"...

How about 'WALG_FORCE_DELTA' ?

x4m:
> > Can we name this knob without double negation? Like "require delta" or "delta only"...
> 
> How about 'WALG_FORCE_DELTA' ?

sounds waaay better 👍

docker/gp_tests/scripts/configs/common_config.json
```diff
@@ -11,4 +11,5 @@
 "WALG_PGP_KEY_PATH": "/tmp/PGP_KEY",
 "WALG_GP_SEG_POLL_INTERVAL": "5s",
 "WALG_GP_SEG_UPD_INTERVAL": "1s",
-"WALG_GP_AOSEG_SIZE_THRESHOLD": "0"
\ No newline at end of file
+"WALG_GP_AOSEG_SIZE_THRESHOLD": "0",
+"WALG_FORCE_DELTA": "true"
```
x4m, line 15: do we really need it in common config? Is it intended?

docs/Greenplum.md
```diff
@@ -89,6 +89,10 @@ wal-g backup-push --delta-from-name backup_name --config=/path/to/config.yaml
 wal-g backup-push --delta-from-user-data "{ \"x\": [3], \"y\": 4 }" --config=/path/to/config.yaml
 ```
 
+* `WALG_FORCE_DELTA`
+
+To prevent WAL-G from falling back to a full backup if the delta backup fails
```
x4m, line 94: Dot at the end would be good IMO :)

x4m (approved):

docker/gp_tests/scripts/configs/common_config.json
fanfuxiaoran: Found that there is no wal-log based delta backup test in gp_tests. Why doesn't gp support this feature as postgres? @x4m

x4m: Ugh, sorry, I missed notification. This feature was not actively used for PG, so noone stood up to make it work for GP...

x4m (approved):

Issue #1855, opened 2024-12-06
pgproto3 SQL Injection via Protocol Message Size Overflow 


olopost:
### Database name
On that point we use postgreSQL database, but the issues is not mainly base on sqldatabase

# Issue description

### Describe your problem
│ github.com/jackc/pgproto3/v2   in 2.0.7 version have an high security security concern that is fix in 2.3.3 version. Can you fix this dependencies version ?

### Please provide steps to reproduce
Scan with trivy the current version

```

usr/bin/wal-g (gobinary)

Total: 5 (UNKNOWN: 0, LOW: 1, MEDIUM: 1, HIGH: 3, CRITICAL: 0)

┌──────────────────────────────────────────────────┬─────────────────────┬──────────┬────────┬─────────────────────┬───────────────┬───────────────────────────────────────────────────────────┐
│                     Library                      │    Vulnerability    │ Severity │ Status │  Installed Version  │ Fixed Version │                           Title                           │
├──────────────────────────────────────────────────┼─────────────────────┼──────────┼────────┼─────────────────────┼───────────────┼───────────────────────────────────────────────────────────┤
│ github.com/Azure/azure-sdk-for-go/sdk/azidentity │ CVE-2024-35255      │ MEDIUM   │ fixed  │ v1.1.0              │ 1.6.0         │ azure-identity: Azure Identity Libraries Elevation of     │
│                                                  │                     │          │        │                     │               │ Privilege Vulnerability in                                │
│                                                  │                     │          │        │                     │               │ github.com/Azure/azure-sdk-for-go/sdk/azidentity          │
│                                                  │                     │          │        │                     │               │ https://avd.aquasec.com/nvd/cve-2024-35255                │
├──────────────────────────────────────────────────┼─────────────────────┼──────────┤        ├─────────────────────┼───────────────┼───────────────────────────────────────────────────────────┤
│ github.com/jackc/pgproto3/v2                     │ GHSA-7jwh-3vrq-q3m8 │ HIGH     │        │ v2.0.7              │ 2.3.3         │ pgproto3 SQL Injection via Protocol Message Size Overflow │
│                                                  │                     │          │        │                     │               │ https://github.com/advisories/GHSA-7jwh-3vrq-q3m8         │
├──────────────────────────────────────────────────┼─────────────────────┤          │        ├─────────────────────┼───────────────┼───────────────────────────────────────────────────────────┤
[...]

Issue #1824, opened 2024-10-30
pgx major version bump


annielzy:
### Database name
PostgreSQL

# Issue description

We need to bump the `pgx` library to v5 to resolve outstanding CVEs: [CVE-2024-27289](https://ubuntu.com/security/CVE-2024-27289) and [CVE-2024-27304](https://ubuntu.com/security/CVE-2024-27304). This update will also include `pgproto3` which is part of `pgx/v5` and resolve the security advisory: [GHSA-7jwh-3vrq-q3m8](https://github.com/jackc/pgproto3/security/advisories/GHSA-7jwh-3vrq-q3m8).

However, the version bump to `pgx` v5 requires significant changes due to drastic API modifications. Additionally, some dependencies need to be updated or replaced, as they have moved or are only compatible with specific versions of `pgx`.

Additionally, `pgx` v5 only officially support PostgreSQL 12 and above. It seems bogus for `wal-g` to have integration tests on older versions of PostgreSQL, for this PR: https://github.com/wal-g/wal-g/pull/1821.

chipitsine:
are those CVE exploitable with wal-g ?

annielzy:
@chipitsine I am fairly sure that [CVE-2024-27289](https://ubuntu.com/security/CVE-2024-27289) is not exploitable since wal-g does not use the non-default simple protocol. For [CVE-2024-27304](https://ubuntu.com/security/CVE-2024-27304) and [GHSA-7jwh-3vrq-q3m8](https://github.com/jackc/pgproto3/security/advisories/GHSA-7jwh-3vrq-q3m8), do you know if wal-g could generate a query or bind message that exceeds 4 GB in size?

chipitsine:
I'm trying to understand urgency and motivation for addressing this CVE.

x4m:
There are no known vulnerabilities in WAL-G. However, there are two reasons to patch: calm down vuln scanners and... well, one day we will have to update anyway.

debebantur:
This [pr](https://github.com/wal-g/wal-g/pull/1893) updated pgx to v5. Is there anything else to do in this issue?

PR #1941, opened 2025-03-20
move pgconn and pgproto used into pgx

fepape-ionos:
### Database name
Postgres

# Pull request description

Automated vulnerability scans picked up that the used pgproto3 contained vulnerabilities.
This was already mentioned in https://github.com/wal-g/wal-g/issues/1824 and https://github.com/wal-g/wal-g/issues/1855
While the above are closed since pgx was updated to v5 the existing references to pgconn and pgproto were not moved to pgx. This lead to a remaining dependency on (vulnerable) pgproto3 even though the patched version is available in pgx v5.

This PR removes the old dependencies and moves pgconn and pgproto3 references to pgx/v5/pgconn and pgx/v5/pgproto3


fepape-ionos:
Regarding the failing postgres tests:
Error from pglogrepl that it got a command_complete instead of the expected ready_for_query. Which strangely is behaviour which pglogrepl would expect for PG > 12. So that might be a bug there for old PG versions.
Is there a reason why the github actions to test use PG 10? Everything below 13 is currently EOL.

fepape-ionos:
> even taking into account that PG10 is out of support, I know some installation of it in production. not a good idea to break PG10 compatibility

With the change in the latest commit in here (and in pglogrepl) PG10 works again. This most likely was a bug affecting pglogrepl for all PG < 13 and should be resolved. I also have a bunch of "unsupported" Postgres instances which would have been affected.

chipitsine:
> > even taking into account that PG10 is out of support, I know some installation of it in production. not a good idea to break PG10 compatibility
> 
> With the change in the latest commit in here (and in pglogrepl) PG10 works again. This most likely was a bug affecting pglogrepl for all PG < 13 and should be resolved. I also have a bunch of "unsupported" Postgres instances which would have been affected.
> 
> I only fail to understand why some MongoDB tests seem to fail. This should not touch any MongoDB related functionality as far as I am aware.

mongo is known to fail from time to time

chipitsine (approved):

chipitsine:
"merge 'em all, let god sort 'em out"

PR #1950, opened 2025-03-29
refactoring oplog replay

Apelsin234:
1) supported LATEST and LATEST_BACKUP pitr aliases 2) added additionalIntervalAfterBackup for saving additional oplog time for backup

### Database name

# Pull request description

// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/internal/databases/mongo/archive/layout.go
+++ b/internal/databases/mongo/archive/layout.go
@@ -174,7 +174,8 @@ func OldestBackupAfterTime(backups []*models.Backup, after time.Time) (*models.B
 // SelectPurgingOplogArchives builds archive list to be deleted.
 func SelectPurgingOplogArchives(archives []models.Archive,
 	backups []*models.Backup,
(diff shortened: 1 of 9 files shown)


internal/databases/mongo/archive/layout.go
```diff
@@ -187,8 +188,8 @@ func SelectPurgingOplogArchives(archives []models.Archive,
 			continue
 		}
 
-		// retain if arch is part of backup
-		if backup := models.FirstOverlappingBackupForArch(arch, backups); backup != nil {
+		// retain if arch is part of [backup.start: backup.end +
```
VGoshev, line 191: Looks like should be ']' instead of '+'

VGoshev (commented):
Little ussye with comment, but in general looks good for me

VGoshev (approved):

PR #1966, opened 2025-04-16
CI: switch docker tests to dind

chipitsine:
switch docker tests to docker in docker mode

### DIFF
--- a/.github/workflows/dockertests.yml
+++ b/.github/workflows/dockertests.yml
@@ -132,7 +132,9 @@ jobs:
           exit 1
 
       - name: Run test
-        run: ${{ matrix.command }}
+        run: |
+          git config --global --add safe.directory ${GITHUB_WORKSPACE}
+          ${{ matrix.command }}
     env:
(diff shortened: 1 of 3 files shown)


ostinru:
Why do you need docker in docker? We have quite a few things that we are running natively on hardware (not in the docker).

chipitsine:
current tests design is brilliant.  binaries are built on a host (whatever it is, ubuntu 20 ... ubuntu 24 ...) and then delivered to old debian images.

it only works if host is ubuntu 20. which we can achive nowdays only by using containers because ubuntu 20 is removed from github actions

chipitsine:
@boosterKRD, what do you think of merging this ?

it will fix at least several tests.

the rest either will be addressed in some better way or maybe I'll have time to fix dind tests (I suspect they fail due to missing health checks, but I need more time to investigate)

chipitsine:
abandonig in favor of https://github.com/wal-g/wal-g/pull/1973

PR #1968, opened 2025-04-17
Fix WalPageSize and BlockSize for gp and cloudberry

fanfuxiaoran:
In both greenplum and cloudberry, the default WalPageSize and BlockSize are 32768 instead of 8192.

Add config item WALG_PG_WAL_PAGE_SIZE and WALG_PG_BLOCK_SIZE

### Database name
Greenplum, Cloudberry

# Pull request description

// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/cmd/gp/gp.go
+++ b/cmd/gp/gp.go
@@ -15,8 +15,10 @@ import (
 	conf "github.com/wal-g/wal-g/internal/config"
 	"github.com/wal-g/wal-g/internal/databases/greenplum"
 	"github.com/wal-g/wal-g/internal/databases/postgres"
+	"github.com/wal-g/wal-g/internal/databases/postgres/orioledb"
 	"github.com/wal-g/wal-g/internal/multistorage"
 	"github.com/wal-g/wal-g/internal/multistorage/policies"
+	"github.com/wal-g/wal-g/internal/walparser"
 	"github.com/wal-g/wal-g/pkg/storages/storage"
 )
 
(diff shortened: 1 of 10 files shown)


reshke:
Will try to take a look soon

x4m:
LGTM, but there's unused function thus linters fail

fanfuxiaoran:
> LGTM, but there's unused function thus linters fail

Thanks. I forgot to upgrade the golangci-lint to 2.0.2, fixed it.

boosterKRD:
@fanfuxiaoran
Just a thought:
Wouldn't it be simpler and more reliable to fetch block_size and wal_block_size directly from the database; instead of relying on manually configured environment variables like WALG_PG_BLOCK_SIZE and WALG_PG_WAL_PAGE_SIZE? (we could automatically detect the correct page sizes and avoid potential misconfiguration)

Also, could you please clarify what is the approximate performance penalty when using the new convertPageForChecksum() function compared to the previous unsafe-based approach?
Is there any noticeable overhead in practice?

fanfuxiaoran:
> @fanfuxiaoran Just a thought: Wouldn't it be simpler and more reliable to fetch block_size and wal_block_size directly from the database; instead of relying on manually configured environment variables like WALG_PG_BLOCK_SIZE and WALG_PG_WAL_PAGE_SIZE? (we could automatically detect the correct page sizes and avoid potential misconfiguration)

Good idea！ I will try it.

fanfuxiaoran:
> @fanfuxiaoran Just a thought: Wouldn't it be simpler and more reliable to fetch block_size and wal_block_size directly from the database; instead of relying on manually configured environment variables like WALG_PG_BLOCK_SIZE and WALG_PG_WAL_PAGE_SIZE? (we could automatically detect the correct page sizes and avoid potential misconfiguration)

Hi @boosterKRD . The above two configurations are mainly used by wal parser in `wal-push` command.  This command will be invoked by postgres `archive_command` on each segment and master. If we try to fetch them from databases, we need connect to local segment or connect to the master to fetch the page size info. However, if we try to connect to local segment , we don't know the segment port( we only know the master port). If we try to connect to the master,  it requires that the master database can be accessed by segment host.

cmd/gp/gp.go
```diff
@@ -37,6 +39,10 @@ var (
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			// Greenplum uses the 64MB WAL segment size by default
 			postgres.SetWalSize(viper.GetUint64(conf.PgWalSize))
+			walparser.SetWalPageSize(viper.GetUint64(conf.PgWalPageSize))
+			walparser.SetBlockSize(viper.GetUint64(conf.PgBlockSize))
+			postgres.SetDatabasePageSize(viper.GetUint64(conf.PgBlockSize))
+			orioledb.SetDatabasePageSize(viper.GetUint64(conf.PgBlockSize))
```
ostinru, line 45: @x4m , what is PageSize in orioleDB?

Issue #1963, opened 2025-04-15
[BUG] Replace ubuntu-20.04 GitHub Actions runners due to deprecation


boosterKRD:
### Database name

all

### WAL-G Version

master

### Describe your problem

GitHub has officially [announced](https://github.com/actions/runner-images/issues/11101) the deprecation of the ubuntu-20.04 runner image
(I just encountered this error in my workflow):

`This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101`

Currently, the WAL-G project uses ubuntu-20.04 in several workflows, including:
- runs-on: ubuntu-20.04
- Matrix values such as os: [ubuntu-20.04]

We should remove it from the workflows or migrate to using containerisation (e.g., [uraimo/run-on-arch-action](https://github.com/uraimo/run-on-arch-action)б but iit does not support amd64).

chipitsine:
@boosterKRD  what if we replace the following  https://github.com/wal-g/wal-g/blob/master/.github/workflows/dockertests.yml#L25

with

```
    runs-on: ubuntu-latest
    container:
      image: ubuntu:20

```

boosterKRD:
> [@boosterKRD](https://github.com/boosterKRD) what if we replace the following https://github.com/wal-g/wal-g/blob/master/.github/workflows/dockertests.yml#L25
> 
> with
> 
> ```
>     runs-on: ubuntu-latest
>     container:
>       image: ubuntu:20
> ```


Sure, but I’m not entirely sure this setup would work for us — we’re relying on **Docker** in the job, and running inside a container might block access to the Docker daemon.

Also, we have the same issue in the following workflows:

- release.yml — here it looks like the best option would be to drop version 20 and build only for 22 and 24 for amd64.
      * Once I finish implementing DEB provisioning via Launchpad and GitHub Pages, this workflow could be deleted.
- unittests.yml — I have no idea about this one.

boosterKRD:
@chipitsine 
Over the next couple of days I can rewrite the workflow in release.yml to use 
`    runs-on: ubuntu-latest
    container:
      image: ubuntu:20`

and then we can close this issue.

boosterKRD:
The issue can be closed after merging https://github.com/wal-g/wal-g/pull/1970

chipitsine:
[...]

PR #1970, opened 2025-04-21
Switch to container due to Ubuntu 20.04 deprecation on GitHub (fix the worflow release.yml)

boosterKRD:
### Database name
all

# Pull request description
This PR switches the Ubuntu 20.04 job to use a container image instead of the deprecated GitHub-hosted runner.

More details in the issue:  https://github.com/wal-g/wal-g/issues/1963

Here  is a screenshot of the new proposed workflow
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,9 @@
 name: Release workflow for tagged versions
 on:
   push:
+    # Sequence of patterns matched against refs/tags
     tags:
-      - 'v*'
+      - 'v*' # Push events to matching v*, i.e. v0.2.19, v0.2.14a
 
 env:
   GO_VERSION: "1.22"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ env:
   USE_BROTLI: 1
   USE_LIBSODIUM: 1
   USE_LZO: 1
+  DEBIAN_FRONTEND: noninteractive
 
 jobs:
   # Prevents race conditions by creating the release only once before uploads.
@@ -38,7 +39,6 @@ jobs:
     steps:
       - name: Install build dependencies
         run: |
-          export DEBIAN_FRONTEND=noninteractive
           apt-get update
           apt-get install -y liblzo2-dev brotli libsodium-dev curl git cmake build-essential tzdata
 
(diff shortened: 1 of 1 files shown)


boosterKRD:
@chipitsine Hi, if it is possible, could you check this PR?

.github/workflows/release.yml
```diff
@@ -66,7 +89,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   release-ubuntu-arm:
-    runs-on: ubuntu-24.04
+    needs: create-release
+    runs-on: ubuntu-latest
```
chipitsine, line 93: I do not think steps are dependant

.github/workflows/release.yml
```diff
@@ -90,6 +114,7 @@ jobs:
             USE_LZO: ${{ env.USE_LZO }}
           shell: /bin/bash
           run: |            
+            export DEBIAN_FRONTEND=noninteractive
             apt-get update
```
chipitsine, line 118: I think there's already such a variable defined in `env:`

chipitsine:
as you tested, no need to approve from myself ))

ok, I found couple cosmetic issues. nothing really critical

.github/workflows/release.yml
boosterKRD: If you’re referring to the needs section — both ARM and AMD jobs are waiting for the first job (create-release) to finish.
In my tests, I ran into a race condition that caused a failure. Previously, we didn’t run into this issue because all jobs would start immediately.
But now, since we’re using containers, there’s some delay: some jobs start right away, while others wait for available runners.
image

chipitsine: you are right

.github/workflows/release.yml
boosterKRD: This variable is not defined in the global env: section.
I’m exporting it inside the container job to make sure apt-get works in non-interactive mode.
Without this variable, the installation of tzdata prompts for user input, which breaks the workflow in CI.

**Or maybe I misunderstood your comment — could you clarify what exactly you meant?**

chipitsine: I would move variable to https://github.com/wal-g/wal-g/blob/fix/release-workflow/.github/workflows/release.yml#L8-L12

but it is a matter of a taste. export works as well

boosterKRD: Okay, I will test your proposal, and if the global env does not work inside a container, then I’ll keep it as it is now; otherwise, I’ll use your version. 
**Thanks!**

boosterKRD: @chipitsine 
DONE (your proposal hit the spot)  - I have updated PR

chipitsine (approved):

Issue #1952, opened 2025-03-31
[BUG] Failures restoring CloudberryDB


gspiegelberg:
### Database name

CloudberryDB / Greenplum

### WAL-G Version

3.0.5

### Describe your problem

Environment:
- AWS EC2 using Rocky 8
- pre-Apache Cloudberry release 1.6.0
- wal-g 3.0.5 built from github using documented instructions installed in /usr/local/cloudberry-db/bin

Successfully restored base data directories (recovering data files). Many issues with WAL replay.

1. wal-g appends “include_if_exists=recovery.conf” to postgresql.conf. Cloudberry DB does not take include_if_exists directive.
   Workaround: Comment out / remove from all postgresql.conf files.
2. wal-g creates a recovery.conf.
   Workaround: Append contents to postgresql.conf.
3. wal-g recreated recovery.conf assumes wal-g installed in /usr/bin. It is not.
   Workaround: Requires that or symlink on all hosts to /usr/local/cloudberry-db/bin/wal-g. 
4. wal-g manipulates pg_hba.conf on all segments incorrectly not specifying /32 after host IP
   Workaround: Must correct manually on all segments
5. Once WAL replay finally initiated and CB started in recovery mode, some WAL’s are pulled down but not all to advance to restore point.
   Workaround: Unknown still digesting test environment logs.

Regarding point 5 above, some WAL's fetched (see below) however many more required (also see below snip from a aws s3 ls of the archive for segment)
`
[gpadmin@sdw1_cbdb_dev pg_wal]$ pwd
/data1/primary/wal-g-test-1/gpseg4/pg_wal
[gpadmin@sdw1_cbdb_dev pg_wal]$ ls -l
total 458760
-rw-------. 1 gpadmin gpadmin 67108864 Mar 31 15:54 000000010000012700000008 <<<< fetched
-rw-------. 1 gpadmin gpadmin 67108864 Mar 31 15:54 000000010000012700000009 <<<< fetched
-rw-------. 1 gpadmin gpadmin 67108864 Mar 31 15:54 00000001000001270000000A <<<< fetched
-rw-------. 1 gpadmin gpadmin 67108864 Mar 31 15:54 00000001000001270000000B <<<< fetched
-rw-------. 1 gpadmin gpadmin 67108864 Mar 31 15:54 00000001000001270000000C <<<< fetched
-rw-------. 1 gpadmin gpadmin 67108864 Mar 31 15:54 00000002000001270000000C
[...]

PR #1974, opened 2025-04-23
Fix failed restore gp cbdb

Mulily0513:
### Database name
greenplumn and cloudberrydb

# Pull request description

- Instead of hard-coding wal-g in the /usr/bin directory, users are allowed to specify its location by configuring environment variables.
- Fix failed to restore CloudberryDB，After the full backup is completed, a restore point needs to be inserted, and the WAL log segments related to this restore point should be manually archived.This operation ensures that the new cluster can obtain the complete WAL logs containing this restore point, thus enabling the execution of the restoration process.
- Fix wal-g manipulates pg_hba.conf on all segments incorrectly not specifying /32 after host IP

This pull request aims to address the issue reported in https://github.com/wal-g/wal-g/issues/1952. For more detailed information, please refer to the same issue link.

### DIFF
--- a/docker/cloudberry_tests/scripts/tests/restore_point.sh
+++ b/docker/cloudberry_tests/scripts/tests/restore_point.sh
@@ -18,6 +18,38 @@ wal-g --config=${TMP_CONFIG} delete everything FORCE --confirm
 wal-g create-restore-point rp1 --config=${TMP_CONFIG}
 wal-g create-restore-point rp2 --config=${TMP_CONFIG}
 
+# Check whether the WAL log is correctly switched and uploaded to S3 after create-restore-point
+# gpadmin@10f4a227f02b:/usr/local/gpdb_src$ wal-g st ls  segments_005/seg0/wal_005/ --config=${TMP_CONFIG}
+# type size    last modified                     name
+# obj  4624920 2025-05-21 07:17:16.052 +0000 UTC 000000010000000000000001.lz4
+# obj  264275  2025-05-21 07:26:06.265 +0000 UTC 000000010000000000000002.lz4
+
+#wait for wal-g to upload WALs
+sleep 5
+
+check_wal_upload() {
+    local path=$1
+
+    wal-g st ls "$path" --config=${TMP_CONFIG}
+
+    wal-g st ls "$path" --config=${TMP_CONFIG} \
+        | awk '/^obj/ {count++} END {exit !(count >= 2)}'
+}
+
+# Check each segment
+for seg_path in \
+    segments_005/seg-1/wal_005/ \
+    segments_005/seg0/wal_005/ \
+    segments_005/seg1/wal_005/ \
+    segments_005/seg2/wal_005/
+do
+    if ! check_wal_upload "$seg_path"; then
+        echo "Error: WAL files after create-restore-point were not correctly uploaded to S3 for $seg_path"
+        exit 1
+    fi
+done
+
+
 # check verify results to end with 'OK'
 if ! (wal-g st ls basebackups_005 --config=${TMP_CONFIG} | grep -q 'rp1_restore_point.json') then
   echo "Error: restore point rp1 metadata file does not exist"
--- a/internal/databases/greenplum/pg_hba_cfg_maker.go
+++ b/internal/databases/greenplum/pg_hba_cfg_maker.go
@@ -3,6 +3,7 @@ package greenplum
 import (
 	"errors"
 	"fmt"
+	"net"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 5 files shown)


internal/databases/greenplum/pg_hba_cfg_maker.go
```diff
@@ -93,3 +94,10 @@ func (m PgHbaMaker) primarySegments() []*cluster.SegConfig {
 	sort.Slice(primarySegments, func(i, j int) bool { return primarySegments[i].ContentID < primarySegments[j].ContentID })
 	return primarySegments
 }
+
+func (m PgHbaMaker) MaskedHostName(hostName string) string {
```
fanfuxiaoran, line 98: Maybe no need to use a function to fix the host name when it is a ip address. The function name is a little confused.

fanfuxiaoran, line 98: how about change the function name to MaskHostName or TransformHostName?

internal/databases/greenplum/backup_push_handler.go
```diff
@@ -215,6 +215,21 @@ func (bh *BackupHandler) HandleBackupPush() {
...
 	tracelog.ErrorLogger.FatalOnError(err)
 
+	// After completing the full backup, insert a restore point and archive related WAL log segments.
+	// This ensures the new cluster can retrieve complete WAL logs with the restore point for restoration.
+	tracelog.InfoLogger.Println("Switch xlog on cluster")
+	remoteOutput = bh.globalCluster.GenerateAndExecuteCommand("Running wal-g", cluster.ON_SEGMENTS|cluster.INCLUDE_MASTER,
+		func(contentID int) string {
+			seg, ok := bh.globalCluster.ByContent[contentID]
+			if ok {
+				return fmt.Sprintf("PGOPTIONS='-c gp_role=utility' psql -p %d -d postgres -c 'select pg_switch_wal();'", seg[0].Port)
```
ostinru, line 225: For greenplum v6.x you should pass `gp_session_role=utility`...

ostinru: And there is no `pg_switch_wal()` in Pg 9.4 / Gp 6.x:
```
postgres=# select pg_switch_wal();
ERROR:  function pg_switch_wal() does not exist
LINE 1: select pg_switch_wal();
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

fanfuxiaoran:
@Mulily0513 Could you  add an integration test for the part of `pg_switch_wal()`?

Mulily0513:
> @Mulily0513 Could you add an integration test for the part of `pg_switch_wal()`?

ok

fanfuxiaoran:
LGTM!

fanfuxiaoran (approved):

fanfuxiaoran:
Hi ! @ostinru Just checking in to see if this PR is ready to be merged. If there are any concerns, please let us know.

docker/cloudberry_tests/scripts/tests/restore_point.sh
```diff
@@ -18,6 +18,38 @@ wal-g --config=${TMP_CONFIG} delete everything FORCE --confirm
...
+sleep 5
+
+check_wal_upload() {
+    local path=$1
+
+    wal-g st ls "$path" --config=${TMP_CONFIG}
+
+    wal-g st ls "$path" --config=${TMP_CONFIG} \
+        | awk '/^obj/ {count++} END {exit !(count >= 2)}'
+}
```
ostinru, line 37: Probably, wal-g on its own should wait for WAL upload before signaling that backup is ready.
But it is out of scope for this PR.

cc: @debebantur

internal/databases/greenplum/restore_point.go
```diff
@@ -151,12 +152,38 @@ func createRestorePoint(conn *pgx.Conn, restorePointName string) (restoreLSNs ma
...
+							pgOptions = "-c gp_session_role=utility"
+							switchFunction = "pg_switch_xlog()"
+						} else {
+							pgOptions = "-c gp_role=utility"
+							switchFunction = "pg_switch_wal()"
+						}
+						return fmt.Sprintf("PGOPTIONS='%s' psql -p %d -d postgres -c 'select %s;'", pgOptions, seg[0].Port, switchFunction)
+					}
+					return ""
+				})
```
ostinru, line 183: Probably, we should also wait for WAL upload before exiting wal-g (that a signal that backup completed successfully). Obviously, we can do this in bash/python as you do in tests, but for 3rd party people it may not be obvious.

cc: @debebantur , I think that we have observed similar issue in production.

PR #1975, opened 2025-04-23
Disable WALG_FORCE_WAL_DELTA in tests

fanfuxiaoran:
The ci failed due to the bug of wal delta backup

wal-g_pg_copy_backup_test wal-g_pg_copy_backup_test wal-g_pg_copy_backup_test
000000010000000000000003
wal-g_pg_copy_backup_test
000000010000000000000005
wal-g_pg_copy_backup_test locations from delta files. | INFO: 2025/04/22 22:15:47.284658 Delta backup enabled | INFO: 2025/04/22 22:15:47.284771 Timeline: 1, FirstUsedLsn: 0/3000028,
    FirstNotUsedLsn: 0/5000028
| INFO: 2025/04/22 22:15:47.284821 First WAL should participate in
    building delta map:
| INFO: 2025/04/22 22:15:47.284841 First WAL shouldn't participate
    in building delta map:
| ERROR: 2025/04/22 22:15:47.294850 Failed to load delta map from
    previous backup: Error during fetch

Just remove the WALG_FORCE_WAL_DELTA in tests to let them pass (although those tests are not correct as they all fall back to full backup). Will bring it back after the bug fixed.

### Database name

# Pull request description

// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/docker/cloudberry/Dockerfile
+++ b/docker/cloudberry/Dockerfile
@@ -83,7 +83,7 @@ RUN chmod +x /root/setup_gpadmin_user.bash \
 WORKDIR /usr/local/gpdb_src
 RUN locale-gen en_US.utf8 \
  && git log -1 \
- && ./configure --with-perl --with-python --with-libxml --with-gssapi --enable-debug --disable-pxf --disable-orca --prefix=/usr/local/gpdb_src  > /dev/null \
+ && ./configure --with-perl --with-python --with-libxml --with-gssapi --enable-debug --disable-pxf --disable-orca --disable-pax --prefix=/usr/local/gpdb_src  > /dev/null \
  && make -j$(nproc) > /dev/null \
  && make -j$(nproc) install > /dev/null \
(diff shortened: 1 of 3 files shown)


x4m (approved):

boosterKRD:
@x4m am I understanding correctly that wal_delta doesn’t work at all anymore? Regarding GSoC, it would be bad if it suddenly turns out that the feature has completely stopped working and doesn’t generate WAL summary :)

fanfuxiaoran:
> @x4m am I understanding correctly that wal_delta doesn’t work at all anymore? Regarding GSoC, it would be bad if it suddenly turns out that the feature has completely stopped working and doesn’t generate WAL summary :)

Hi,  @boosterKRD , the wal_delta can work, but in some case it will fail: 
If the first wal file pushed by the wal-g is not started as xxx0 after the archive_mode enabled, and we start the basebackup, then the first delta segment will be missing for that basebackup.

PR #1976, opened 2025-04-25
Disable HTTP keepalive in endpoint requests

EinKrebs:


reshke (approved):

ostinru:
@EinKrebs why do you want to disable keepalives?

PR #1977, opened 2025-04-25
CI: run govulncheck

chipitsine:
compliance check beloved by corporate security officers


.github/workflows/golvulncheck.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+        id: go
+
+      - name: Get dependencies
+        run: |
+          go get -v -t -d ./...
+      - id: govulncheck
+        uses: golang/govulncheck-action@v1
```
ostinru: It looks like it produces a lot of irrelevant to PR errors.
I will suggest adding `continue-on-error: true` for this step?

chipitsine: it will not change things. if those errors are ignored now in some way, they will be ignored in another way.

the idea is to fix those findings first. after that it is ok to merge

chipitsine: should be ok now

PR #1981, opened 2025-05-06
golangci-lint 2.1

serprex:


internal/tools/go.mod
```diff
@@ -1,144 +1,209 @@
 module github.com/jaegertracing/jaeger-clickhouse/internal/tools
 
-go 1.22
+go 1.23.0
+
+toolchain go1.24.2
```
serprex, line 5:
```suggestion
```
at least go 1.25 will stop adding this

internal/tools/tools.go
```diff
@@ -1,6 +1,6 @@
 package tools
 
 import (
-	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
+	_ "github.com/golangci/golangci-lint/v2/cmd/golangci-lint"
```
serprex, line 4: what is this used for? could this be removed?

golangci-lint supports sorting go imports now, I'm guessing everyone is using golangci-lint directly

debebantur (approved):

PR #1982, opened 2025-05-06
update azure sdk

serprex:
this removes a few transitive outdated dependencies

autorest repo has been archived, so replaced usage with constants


chipitsine:
I bet we should also bump to 1.23.x in GHA https://github.com/wal-g/wal-g/blob/master/.github/workflows/dockertests.yml#L19  (and few more)

PR #1984, opened 2025-05-08
Refactor: Add context timeouts for pgx queries and connections

### LINKED ISSUE #1969: [FEATURE]  Review how WAL-G connects and works with PostgreSQL via the PGX library

I would like to review and improve the way WAL-G uses the PGX library to interact with PostgreSQL.

Planned improvements:

1. Add support for context with timeout to manage query execution time.
2. Investigate PGX connection modes and switch to one that ensures compatibility with connection poolers that don’t support the extended protocol (e.g., PREPARE),since we can't guarantee what connection string the user will configure.
- Either switch to Simple Protocol mode
- Or use an Extended Protocol mode that sends Parse/Bind/Execute in a single Sync message 

It ensures WAL-G works reliably with all connection poolers, avoiding protocol-related errors during query execution.

boosterKRD:
### Database name
PostgreSQL

# Pull request description
Contexts are now passed to SQL queries and to pgx.ConnectConfig calls in PostgreSQL code. This allows better control over timeouts during both query execution and connection establishment.

ℹ️**NOTICE:**ℹ️
- Greenplum-specific logic (e.g., GpQueryRunner) was not modified — context usage should be added separately by maintainers familiar with that code.
- Additionally, the PostgreSQL Connect function contains fallback logic for Greenplum (e.g., tryConnectToGpSegment). Contexts were intentionally not added there to avoid introducing changes into unclear cross-project code paths. Greenplum maintainers should review this part as well.

Context timeouts are now configured through the **WALG_PG_TIMEOUT** setting.
If this setting is not specified, the default behavior is to have no timeout (i.e., an unlimited execution time).

**Also, the WALG_STOP_BACKUP_TIMEOUT setting is now deprecated. Its functionality is fully covered by the new context-based timeout (WALG_PG_TIMEOUT).
It is recommended to plan its removal from both code and documentation in a future release.**

Fixes #1969

### DIFF
--- a/internal/databases/postgres/configure.go
+++ b/internal/databases/postgres/configure.go
@@ -65,10 +65,5 @@ func getPgTimeoutSetting() (time.Duration, error) {
 		return 0, nil
 	}
 
-	timeout, err := conf.GetDurationSetting(conf.PgTimeout)
-	if err != nil {
-		return 0, err
-	}
-
-	return timeout, nil
+	return conf.GetDurationSetting(conf.PgTimeout)
 }
--- a/internal/databases/postgres/query_runner.go
+++ b/internal/databases/postgres/query_runner.go
@@ -760,7 +760,6 @@ func (queryRunner *PgQueryRunner) GetDataChecksums() (string, error) {
 
 	var dataChecksums string
 	conn := queryRunner.Connection
-	time.Sleep(300 * time.Second)
 	err := conn.QueryRow(pgCtx, "SHOW data_checksums").Scan(&dataChecksums)
 	if err != nil {
 		return "", errors.Wrap(err, "GetDataChecksums: failed to check data_checksums")


chipitsine:
I tried "govulncheck" and it thinks that pgx is vulnerable

https://github.com/wal-g/wal-g/actions/runs/14868038049/job/41749591208?pr=1977

do we have plans of updating pgx as well ?

chipitsine:
somehow govulncheck is thinking on sqlserver

image


thank for the explanation! 
no more questions

internal/databases/postgres/configure.go
```diff
@@ -58,3 +59,16 @@ func getStopBackupTimeoutSetting() (time.Duration, error) {
...
+	if !viper.IsSet(conf.PgTimeout) {
+		return 0, nil
+	}
+
+	timeout, err := conf.GetDurationSetting(conf.PgTimeout)
+	if err != nil {
+		return 0, err
+	}
+
+	return timeout, nil
```
ostinru: could be simplified to `return conf.GetDurationSetting(conf.PgTimeout)`

internal/databases/postgres/connect.go
```diff
@@ -28,8 +29,23 @@ func Connect(configOptions ...func(config *pgx.ConnConfig) error) (*pgx.Conn, er
...
+	conn, err := pgx.ConnectConfig(ctx, config)
 	if err != nil {
+		tracelog.ErrorLogger.Printf("Failed to connect normally: %v", err)
+
+		// TODO(greenplum-split):
+		// This fallback calls Greenplum-specific tryConnectToGpSegment()
+		// even in a plain PostgreSQL build. As a result, the PG server
+		// replies with FATAL: unrecognized configuration parameter "gp_role".
+		// We need to separate GP logic from PG.
 		conn, err = tryConnectToGpSegment(config)
```
ostinru, line 49: Let's pass (probably new) context here as well. Otherwise Pg's connect() may stuck here.

internal/databases/postgres/query_runner.go
```diff
@@ -716,10 +755,13 @@ func (queryRunner *PgQueryRunner) processTables(conn *pgx.Conn,
 func (queryRunner *PgQueryRunner) GetDataChecksums() (string, error) {
 	queryRunner.Mu.Lock()
 	defer queryRunner.Mu.Unlock()
+	pgCtx, cancel := ContextWithPgTimeout(queryRunner.Ctx, queryRunner.PgTimeout)
+	defer cancel()
 
 	var dataChecksums string
 	conn := queryRunner.Connection
-	err := conn.QueryRow(context.TODO(), "SHOW data_checksums").Scan(&dataChecksums)
+	time.Sleep(300 * time.Second)
```
ostinru: should it be here?

internal/databases/postgres/query_runner.go
```diff
@@ -180,19 +192,23 @@ func (queryRunner *PgQueryRunner) buildGetPhysicalSlotInfo() string {
 func (queryRunner *PgQueryRunner) getVersion() (err error) {
 	queryRunner.Mu.Lock()
 	defer queryRunner.Mu.Unlock()
+	pgCtx, cancel := ContextWithPgTimeout(queryRunner.Ctx, queryRunner.PgTimeout)
+	defer cancel()
```
ostinru, line 196: it seems that we can shorten code a bit by removing `defer cancel()` lines

internal/databases/postgres/connect.go
```diff
@@ -28,8 +29,23 @@ func Connect(configOptions ...func(config *pgx.ConnConfig) error) (*pgx.Conn, er
...
+
+	conn, err := pgx.ConnectConfig(ctx, config)
 	if err != nil {
+		tracelog.ErrorLogger.Printf("Failed to connect normally: %v", err)
+
+		// TODO(greenplum-split):
+		// This fallback calls Greenplum-specific tryConnectToGpSegment()
+		// even in a plain PostgreSQL build. As a result, the PG server
+		// replies with FATAL: unrecognized configuration parameter "gp_role".
+		// We need to separate GP logic from PG.
```
ostinru, line 48: x4m doing refactoring here: https://github.com/wal-g/wal-g/pull/2004
feel free to leave feedback there.

internal/databases/postgres/configure.go
boosterKRD: @ostinru 
I have fixed it

internal/databases/postgres/query_runner.go
boosterKRD: @ostinru 
This `"time.Sleep(300 * time.Second)"` was part of the test code, and I just forgot to wipe it out - pardon me :)

internal/databases/postgres/query_runner.go
boosterKRD: Hi @ostinru, thanks for the suggestion!

I just wanted to double-check: did you mean that we should remove defer cancel() entirely?

My understanding is that even if the context isn’t canceled explicitly, calling cancel() is still important to free up internal resources.

internal/databases/postgres/connect.go
boosterKRD: @ostinru 
Yes, I fully agree — as you mentioned in the later comment, it makes perfect sense to separate out the Greenplum-specific logic in a dedicated refactoring PR #2004.

internal/databases/postgres/query_runner.go
```diff
@@ -144,7 +146,17 @@ func NewPgQueryRunner(conn *pgx.Conn) (*PgQueryRunner, error) {
...
 
-	r := &PgQueryRunner{Connection: conn, stopBackupTimeout: timeout}
+	pgTimeout, err := getPgTimeoutSetting()
+	if err != nil {
+		return nil, err
+	}
+
+	r := &PgQueryRunner{
+		Connection:        conn,
+		stopBackupTimeout: timeout, // depricated
```
serprex, line 156:
```suggestion
		stopBackupTimeout: timeout, // deprecated
```

PR #1985, opened 2025-05-11
remove internal/tools

serprex:
this seems to be pretty outdated, I expect this went unnoticed as people are relying on self installed golangci-lint

also adds gci & goimports to golangci config


ostinru (approved):

x4m:
Do we need to override code owners here?

serprex:
@x4m yes

x4m (approved):

PR #1995, opened 2025-05-22
add partially restore to mongodb

NotAmigo:
### Database name
MongoDB

# Pull request description
Add partially restore to MongoDB. Now wal-g makes dir to databases backup for mongodb. Also, it can restore dbs and collections partially. After downloading tar files, it runs mongo with "--restore" flag, which should restore databases and run it correctly. 
// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>


NotAmigo:
So much advices I'm requesting for. 
Firstly, should I do universal dir_database_tar_ball_composer? 
Secondly, should I add restore flag for mongo client to some other methods maybe? 
And thirdly, should I move backup_routes_info, pathFilter and tarFilter methods into any other file or dir? 

Also, requests for some better backup_routes_info structure name

kronwerk:
@NotAmigo hi, LGTM, but definitely need some tests [here](https://github.com/wal-g/wal-g/tree/master/tests_func/features/mongodb) - at least 1 basic positive scenario, but in my opinion it's a rather complex feature and needs 2-3 basic and 4-5 negative scenarios

cmd/mongo/binary_backup_fetch.go
```diff
@@ -39,21 +39,30 @@ const (
 	PitrSinceDescription          = "Timestamp point in time recovery start"
 	PitrUntilFlag                 = "pitr-until"
 	PitrUntilDescription          = "Timestamp point in time recovery finish"
+
+	PartiallyRestorePathsFlag        = "partially-restore-paths"
+	PartiallyRestorePathsDescription = "Comma separated dbname.colname records from wished databases " +
+		"and collections restored partially. Indexes included"
+	PartiallyRestoreWithSystemDBsFlag        = "with-system-dbs"
+	PartiallyRestoreWithSystemDBsDescription = "Always restore 'admin' and 'local' dbs in partially restore. " +
```
VGoshev: Restore `admin` and `local` + restore `local` is RS is set.
`local` present 2 times. looks like it should be `config` in case of rs-name flag is set

internal/databases/mongo/binary/mongod.go
```diff
@@ -316,6 +316,127 @@ func updateRsConfig(ctx context.Context, localDatabase *mongo.Database, rsConfig
...
+	defer cursor.Close(mongodService.Context)
+
+	for cursor.Next(mongodService.Context) {
+		var result bson.M
+		if err := cursor.Decode(&result); err != nil {
+			return nil, err
+		}
+
+		collType, ok := result["type"].(string)
+		if !ok || collType != "view" {
```
VGoshev, line 344: So, we can't restore views? If there is some issue with it?

internal/databases/mongo/binary/restore.go
```diff
@@ -42,59 +42,116 @@ func (restoreService *RestoreService) DoRestore(
 	if err != nil {
 		return err
 	}
+	metadata, err := common.DownloadMetadata(restoreService.Uploader.Folder(), args.BackupName)
```
VGoshev: So if metadata file is absent, backup can't be restored?
If I understand correctly, metadata file need for partial restore only, so in case of full restoration it isn't necessary, so I propose to download it only in case of partial restore.

VGoshev (dismissed):

internal/databases/mongo/binary/mongod.go
NotAmigo: MongoDB does not store views on disk: https://mongodb.prakticum-team.ru/docs/manual/core/views/#:~:text=MongoDB%20does%20not%20persist%20the%20view%20contents%20to%20disk.%20A%20view%27s%20content%20is%20computed%20on%2Ddemand%20when%20a%20client%20queries%20the%20view.

internal/databases/mongo/binary/restore.go
```diff
@@ -198,3 +226,80 @@ func (restoreService *RestoreService) oplogReply(rsConfig RsConfig, replayOplogC
...
+	if err != nil {
+		return errors.Wrap(err, "unable to start mongod in restore mode")
+	}
+	defer mongodProcess.Close()
+
+	mongodService, err := CreateMongodService(
+		restoreService.Context,
+		"wal-g restore",
+		mongodProcess.GetURI(),
+		10*time.Minute,
```
kronwerk: looks like an option to config with default value around this: https://github.com/wal-g/wal-g/blob/9fcee2e8099a21637215552c82a0c20cb3d26c21/internal/config/config.go#L150

cmd/mongo/partial_restore.go
```diff
@@ -0,0 +1,85 @@
...
+		internal.ConfigureLimiters()
+
+		ctx, cancel := context.WithCancel(context.Background())
+		signalHandler := utility.NewSignalHandler(ctx, cancel, []os.Signal{syscall.SIGINT, syscall.SIGTERM})
+		defer func() { _ = signalHandler.Close() }()
+
+		backupName := args[0]
+		mongodConfigPath := args[1]
+		mongodVersion := args[2]
+		whitelist := strings.Split(args[3], ",")
```
kronwerk, line 58: does not really match with <dbname.colname> short description - seems a bit confusing right now from a user perspective, I doubt

internal/concurrent_uploader.go
```diff
@@ -23,6 +23,7 @@ func CreateConcurrentUploader(
 	backupName string,
 	directory string,
 	skipFileNotExists bool,
+	tarBallComposerMaker TarBallComposerMaker,
```
kronwerk: starting from 5 args I prefer converting it to 1 args struct argument - but that's definitely NIT

tests_func/features/mongodb/binary_partially_restore.feature
```diff
@@ -0,0 +1,19 @@
...
+    And mongodb initialized on mongodb01
+
+  Scenario: Partially restore works correctly
+    When mongodb01 has partially test mongodb data
+    And we create binary mongo-backup on mongodb01
+    Then we got 1 backup entries of mongodb01
+
+    Given mongodb02 has no data
+    And mongodb initialized on mongodb02
+    When we restore partially binary mongo-backup #0 to mongodb02 with parts "part1.col2"
```
kronwerk: seems that a few more tests won't be that hard to add:
1) [neg] partial restore with inexistent db;
2) [neg] with inexistent col of actual db;
3) [pos] with some blacklist that actually stops something from source dbs/cols and whitelist dbs/cols;
4) [pos] different results with or without system dbs flag.

not sure about any more tests, but these seem to be perfectly expected in prod envs, I guess.

kronwerk (changes requested):

internal/databases/mongo/binary/dir_database_tar_ball_composer.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (d DirDatabaseTarBallComposer) FinishComposing() (internal.TarFileSets, error) {
+	// Push Headers in first part
+	err := d.addListToTar(make([]*internal.ComposeFileInfo, 0))
+	if err != nil {
+		return nil, err
+	}
+
+	eg := errgroup.Group{}
```
Apelsin234: use WithContext(context.Background())

internal/databases/mongo/binary/dir_database_tar_ball_composer.go
```diff
@@ -0,0 +1,135 @@
...
+	return newDirDatabaseTarBallComposer(
+		maker.files,
+		bundle.TarBallQueue,
+		packer,
+		maker.tarFileSets,
+		bundle.Crypter,
+	), nil
+}
+
+type DirDatabaseTarBallComposer struct {
```
Apelsin234: Why we copy it from postgres , I think we have to put the code in the common `internal` and reuse it

Apelsin234 (changes requested):

kronwerk:
LGTM, pls wait review from Apelsin234@

cmd/mongo/backup_push.go
```diff
@@ -19,10 +19,13 @@ const (
...
 	PermanentFlag              = "permanent"
 	PermanentShorthand         = "p"
+
+	SkipMetadataFlag = "skip-metadata"
 )
 
 var (
-	permanent = false
+	permanent    = false
+	skipMetadata = false
```
Apelsin234: Why you add flag to backuo_push , but use only in binary_backup_push

Apelsin234 (approved):
LGTM for me

kronwerk (approved):

Issue #1152, opened 2021-11-18
PostgreSQL: WAL-g delete will do nothing when one of the backup is marked as permanent


t83714:
### Env & Setup

DB: PostgreSQL
Storage: GCS

# Issue description

### Describe your problem

When run `wal-g backup-list --detail`, there are more than 7 backups are listed (no. 4 backup is marked as permanent):

```
name                          modified             wal_segment_backup_start start_time           finish_time          hostname                                            data_dir                 pg_version start_lsn    finish_lsn   is_permanent
base_000000010000000000000759 2021-10-02T01:15:42Z 000000010000000000000759 2021-10-02T01:02:06Z 2021-10-02T01:15:42Z combined-db-postgresql-backup-jobs-1633136520-g4pvp /bitnami/postgresql/data 130003     31557943336  31574720592  false
base_00000001000000000000075C 2021-10-02T01:39:28Z 00000001000000000000075C 2021-10-02T01:26:12Z 2021-10-02T01:39:28Z combined-db-postgresql-backup-jobs-1633137960-k7ftz /bitnami/postgresql/data 130003     31608274984  31625052240  false
base_000000010000000000000762 2021-10-02T04:15:11Z 000000010000000000000762 2021-10-02T04:04:20Z 2021-10-02T04:15:11Z combined-db-postgresql-backup-jobs-1633147440-h5twg /bitnami/postgresql/data 130003     31708938280  31725715536  false
base_000000010000000700000067 2021-10-02T05:47:35Z 000000010000000700000067 2021-10-02T05:31:35Z 2021-10-02T05:47:34Z combined-db-postgresql-0                            /bitnami/postgresql/data 130003     31792824360  31809631192  true
base_00000002000000000000083E 2021-10-16T01:28:11Z 00000002000000000000083E 2021-10-16T01:15:10Z 2021-10-16T01:28:10Z combined-db-postgresql-backup-jobs-1634346900-qn9nd /bitnami/postgresql/data 130003     35399925800  35506996256  false
base_00000002000000000000084F 2021-10-16T02:28:19Z 00000002000000000000084F 2021-10-16T02:15:12Z 2021-10-16T02:28:19Z combined-db-postgresql-backup-jobs-1634350500-lhd9f /bitnami/postgresql/data 130003     35685138472  35729765400  false
[...]

PR #1998, opened 2025-05-30
Feat: Add new commands to support gpdb/cbdb continuous recovery

Mulily0513:
### Database name
greenplum / cloudberrry

# Pull request description

When using wal-g to restore to a specific restore point, the recovery_target_action is defaulted to promote, which causes the database to start a new timeline after successful startup. This prevents reapplying restore points with later timestamps. This PR adds two commands to address this issue.

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/docker/cloudberry_tests/scripts/tests/follow_primary.sh
+++ b/docker/cloudberry_tests/scripts/tests/follow_primary.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+set -e -x
+CONFIG_FILE="/tmp/configs/create_restore_point_config.json"
+RESTORE_CONFIG="/tmp/configs/follow_primary_restore_config.json"
+COMMON_CONFIG="/tmp/configs/common_config.json"
+TMP_CONFIG="/tmp/configs/tmp_config.json"
+cat ${CONFIG_FILE} > ${TMP_CONFIG}
+echo "," >> ${TMP_CONFIG}
+cat ${COMMON_CONFIG} >> ${TMP_CONFIG}
+/tmp/pg_scripts/wrap_config_file.sh ${TMP_CONFIG}
+source /tmp/tests/test_functions/util.sh
+
+bootstrap_gp_cluster
+setup_wal_archiving
+
+wal-g --config=${TMP_CONFIG} delete everything FORCE --confirm
+
+wal-g backup-push --config=${TMP_CONFIG}
+echo "Inserting sample data..."
+psql -p 7000 -c "DROP DATABASE IF EXISTS test"
+psql -p 7000 -c "CREATE DATABASE test"
+psql -p 7000 -d test -c "CREATE TABLE heap AS SELECT a FROM generate_series(1,10) AS a;"
+wal-g create-restore-point rp1 --config=${TMP_CONFIG}
+psql -p 7000 -d test -c "INSERT INTO heap (a) SELECT a FROM generate_series(1, 10) AS a;"
+wal-g create-restore-point rp2 --config=${TMP_CONFIG}
+psql -p 7000 -d test -c "INSERT INTO heap (a) SELECT a FROM generate_series(1, 10) AS a;"
+wal-g create-restore-point rp3 --config=${TMP_CONFIG}
+wal-g backup-list --config=${TMP_CONFIG}
+wal-g restore-point-list --config=${TMP_CONFIG}
+stop_and_delete_cluster_dir
+
+wal-g backup-fetch LATEST --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+wal-g recovery-action shutdown --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+wal-g follow-primary rp1 --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+sleep 30
+wal-g follow-primary rp2 --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+sleep 30
+wal-g follow-primary rp3 --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+sleep 30
+wal-g recovery-action promote --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+prepare_cluster
+start_cluster
+test "$(psql -p 7000 -d test -t -A -c "SELECT count(*) FROM heap;")" -eq 30 || { echo "Test failed: The count is not equal to 30."; exit 1; }
+
+cleanup
+rm ${TMP_CONFIG}
--- a/internal/databases/greenplum/follow_primary_handler.go
+++ b/internal/databases/greenplum/follow_primary_handler.go
@@ -1,0 +1,110 @@
+package greenplum
+
+import (
+	"fmt"
+	"github.com/spf13/viper"
+	conf "github.com/wal-g/wal-g/internal/config"
+	"path"
+	"strings"
+
+	"github.com/greenplum-db/gp-common-go-libs/cluster"
+	"github.com/wal-g/tracelog"
+)
+
+type FollowPrimaryHandler struct {
+	cluster            *cluster.Cluster
+	stopAtRestorePoint string
+	timeoutInSeconds   int
+}
+
+// nolint:gocritic
+func NewFollowPrimaryHandler(
+	logsDir string,
+	restoreCfgPath, stopAtRestorePoint string,
+	timeoutInSeconds int,
+) *FollowPrimaryHandler {
+	restoreCfg, err := readRestoreConfig(restoreCfgPath)
+	tracelog.ErrorLogger.FatalOnError(err)
+
+	initGpLog(logsDir)
+
+	segmentConfigs := make([]cluster.SegConfig, 0)
+	for contentID, segRestoreCfg := range restoreCfg.Segments {
+		segmentConfigs = append(segmentConfigs, segRestoreCfg.ToSegConfig(contentID))
+	}
+
+	globalCluster := cluster.NewCluster(segmentConfigs)
+	tracelog.DebugLogger.Printf("cluster %v\n", globalCluster)
+
+	return &FollowPrimaryHandler{
+		cluster:            globalCluster,
+		stopAtRestorePoint: stopAtRestorePoint,
+		timeoutInSeconds:   timeoutInSeconds,
+	}
+}
+
+func (fh *FollowPrimaryHandler) Follow() {
+	tracelog.InfoLogger.Println("Updating recovery.conf on segments and master...")
+	fh.updateRecoveryConfigs()
+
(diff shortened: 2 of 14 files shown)


ostinru:
Looks interesting! I will check it in next 1-2 weeks.

fanfuxiaoran:
Hi @Mulily0513 , better to make follow-primary command to support 'LATEST' restore point.

docker/cloudberry_tests/scripts/tests/follow_primary.sh
```diff
@@ -0,0 +1,46 @@
...
+stop_and_delete_cluster_dir
+
+wal-g backup-fetch LATEST --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+wal-g recovery-action shutdown --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+wal-g follow-primary rp1 --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+sleep 30
+wal-g follow-primary rp2 --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+sleep 30
+wal-g follow-primary LATEST --restore-config=${RESTORE_CONFIG} --config=${TMP_CONFIG}
+sleep 30
```
ostinru: usually we try to avoid `sleep()` in tests - it slows down tests =(

ostinru:
@Mulily0513 , what is expected workflow/usecase? Keep second cluster in recovering state (by constantly move it to new restore point) and when primary cluster fails - promote this second cluster?

Mulily0513:
> @Mulily0513 , what is expected workflow/usecase? Keep second cluster in recovering state (by constantly move it to new restore point) and when primary cluster fails - promote this second cluster?

Your understanding is correct. Compared with incremental recovery, continuous recovery is more lightweight, enabling faster creation of recovery points while occupying less storage resources and time. Additionally, it supports frequent creation of recovery points (e.g., every 5-10 minutes), thereby providing more recovery time points. This allows users to more easily obtain the cluster status close to the time of the disaster, thus optimizing data security.

docker/cloudberry_tests/scripts/tests/follow_primary.sh
Mulily0513: During the actual testing process, when attempting to perform rapid consecutive recoveries from RP1 to RP2, I encountered the following error:
image
image

When performing the follow - primary operation to apply recovery points (such as switching from rp1 to rp2 and then to LATEST), the database needs to complete operations like applying recovery logs and switching instance states. These operations are not completed instantaneously, so a sleep is needed to allow sufficient time.

internal/databases/greenplum/segment_config_maker.go
```diff
@@ -18,19 +18,8 @@ type RestoreCfgSegMaker struct {
 	restoreCfg ClusterRestoreConfig
 }
 
-func NewRestoreCfgSegMaker(restoreConfigReader io.Reader) (SegConfigMaker, error) {
```
fanfuxiaoran: Looks there is no need to refactor this in this pr.
And in wal-g, there are a lot places use the io.Reader as the input.
From my point of view, the original code is more readable. `NewRestoreCfgSegMaker` takes the io.Reader as a input and transform the bytes in it to restore config.

internal/databases/greenplum/recovery_cfg_maker.go
```diff
@@ -28,7 +31,11 @@ func (m RecoveryConfigMaker) Make(contentID int, pgVersion int) string {
 	)
 
 	if pgVersion >= 120000 {
```
fanfuxiaoran: Should we support 'recovery_target_action=pause' here ?

internal/databases/greenplum/follow_primary_handler.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (fh *FollowPrimaryHandler) Follow() {
+	tracelog.InfoLogger.Println("Updating recovery.conf on segments and master...")
+	fh.updateRecoveryConfigs()
+
+	tracelog.InfoLogger.Println("Running recovery on segments and master...")
+	// Run WAL-G to restore the each segment as a single Postgres instance
+	remoteOutput := fh.cluster.GenerateAndExecuteCommand("Running wal-g",
```
fanfuxiaoran, line 112: Could we put the code starting each pg segment instance in the cluster  and doing recovery into a separate function , just like updateRecoveryConfigs does?

internal/databases/greenplum/recovery_cfg_maker.go
fanfuxiaoran: recovery_target_action=pause should be used with 'hot_standby=on', to enable user to check if reecovery target is the most desirable point for recovery.  When 'hot_standby=off',  a setting of pause will act the same as shutdown. 
For gpdb6, hot_standby is not supported. 
For cloudberry and gpdb7, we have other guc 'gp_pause_on_restore_point_replay' to help follow the primary node. So here, there is no need to suppose `pause`

ostinru:
> > @Mulily0513 , what is expected workflow/usecase? Keep second cluster in recovering state (by constantly move it to new restore point) and when primary cluster fails - promote this second cluster?
> 
> Your understanding is correct. Compared with incremental recovery, continuous recovery is more lightweight, enabling faster creation of recovery points while occupying less storage resources and time. Additionally, it supports frequent creation of recovery points (e.g., every 5-10 minutes), thereby providing more recovery time points. This allows users to more easily obtain the cluster status close to the time of the disaster, thus optimizing data security.

Ok. Let's consider following scenario: we have greenplum/cloudberry cluster in datacenter DC1 and its copy in DC2. When DC1 fails - we should be able to promote cluster in DC2. There is the chance that recovery point had been created by old master in DC1, however WAL file wasn't uploaded to persistent storage (S3). Your last PR https://github.com/wal-g/wal-g/pull/1974 reduces this probability, but doesn't make it impossible. So, at this point we shouldn't allow wal-g to apply WAL-s in gpdb/cbdb.

I will suggest to add following check: get LSNs from restore-point's json, convert it to WAL file name and check that WAL file was uploaded for all segments and master. And only then allow cluster to apply WALs till restore point.
Probably, validation that _all_ WALs exists between current LSN position and new LSN position would be quite expensive from S3 point of view (file listing is way more expensive than single file check).

PS: probably, LATEST selector should try previous restore point in this case.
PPS: we can reuse this mechanism in backup-push workflow: wait for wall WALs to be uploaded to guarantee that we can restore from backup as soon as backup's sentinel uploaded.

Mulily0513:
> > > @Mulily0513 , what is expected workflow/usecase? Keep second cluster in recovering state (by constantly move it to new restore point) and when primary cluster fails - promote this second cluster?
> > 
> > 
> > Your understanding is correct. Compared with incremental recovery, continuous recovery is more lightweight, enabling faster creation of recovery points while occupying less storage resources and time. Additionally, it supports frequent creation of recovery points (e.g., every 5-10 minutes), thereby providing more recovery time points. This allows users to more easily obtain the cluster status close to the time of the disaster, thus optimizing data security.
> 
> Ok. Let's consider following scenario: we have greenplum/cloudberry cluster in datacenter DC1 and its copy in DC2. When DC1 fails - we should be able to promote cluster in DC2. There is the chance that recovery point had been created by old master in DC1, however WAL file wasn't uploaded to persistent storage (S3). Your last PR #1974 reduces this probability, but doesn't make it impossible. So, at this point we shouldn't allow wal-g to apply WAL-s in gpdb/cbdb.
> 
> I will suggest to add following check: get LSNs from restore-point's json, convert it to WAL file name and check that WAL file was uploaded for all segments and master. And only then allow cluster to apply WALs till restore point. Probably, validation that _all_ WALs exists between current LSN position and new LSN position would be quite expensive from S3 point of view (file listing is way more expensive than single file check).
> 
> PS: probably, LATEST selector should try previous restore point in this case. PPS: we can reuse this mechanism in backup-push workflow: wait for wall WALs to be uploaded to guarantee that we can restore from backup as soon as backup's sentinel uploaded.

ok，I'll add the check to verify that all required WAL files exist before applying them in the cluster.

internal/databases/greenplum/follow_primary_handler.go
Mulily0513: ok

internal/databases/greenplum/segment_config_maker.go
Mulily0513: ok

woblerr:
Hello everyone,
are there any updates on this PR?

A very desirable feature.

Mulily0513:
> Hello everyone, are there any updates on this PR?
> 
> A very desirable feature.

To progress with this PR, we first need to enable support for the Cloudberry kernel's hot standby feature, which is being addressed in PR #1152 (https://github.com/apache/cloudberry/pull/1152). I'm actively working to advance that PR at the moment.

internal/databases/greenplum/follow_primary_handler.go
```diff
@@ -0,0 +1,171 @@
...
+	restoreCfgMaker := NewRecoveryConfigMaker(
+		"wal-g", conf.CfgFile, recoveryTarget, true)
+
+	remoteOutput := fh.cluster.GenerateAndExecuteCommand("Updating recovery.conf on segments and master",
+		cluster.ON_SEGMENTS|cluster.INCLUDE_MASTER,
+		func(contentID int) string {
+			segment := fh.cluster.ByContent[contentID][0]
+			pathToRestore := path.Join(segment.DataDir, viper.GetString(conf.GPRelativeRecoveryConfPath))
+			// For this feature, we expect Cloudberry / Greenplum 6.25+ (in this version some patches from 9.5 were backported)
+			fileContents := restoreCfgMaker.Make(contentID, 90500)
```
ostinru, line 142: It is better to have some way to detect features supported by gpdb/cbdb... =/ 
Will think a bit.

ostinru: Probably, we should do nothing with this. Just wait untill greenplum 6.24- phase out.

ostinru:
Merged this PR. Will test on real clusters next week. Please send changes as new PRs.
Thank everyone for the feature!

ostinru:
@Mulily0513 , another cloudberry PR has been merged: https://github.com/apache/cloudberry/pull/1268
What is in your further plans?

Mulily0513:
> @Mulily0513 , another cloudberry PR has been merged: [apache/cloudberry#1268](https://github.com/apache/cloudberry/pull/1268) What is in your further plans?

In the previous PRs, the default recovery_action was set to shutdown, which meant that the cluster would shut down every time it was restored to a restore point, making it unable to provide query capabilities. With the merged PR [#1268](https://github.com/apache/cloudberry/pull/1268), we’ve now implemented continuous recovery without requiring a shutdown, while also enabling readable query support.
Next, I plan to submit a PR to enable wal-g adaptation for this feature.

PR #2001, opened 2025-06-02
wal-verify flags with specified lsn and timeline to check

instahipsta14:
### Database name
postgresql

# Pull request description

When a replica is down, we need to decide how to restore it: either from storage or from the master. To make this decision, we have to check the WAL integrity. wal-g wal-verify can do this, but in the current version, it must select the current timeline and LSN from a live server. With this feature, we can retrieve the information from the primary and specify it in wal-g wal-verify using the flags --lsn and --timeline.

To reduce the checking time, I added the --backup-name flag to verify the range starting from the specified backup, instead of using the earliest valid backup.

-

-


debebantur:
LGTM, lets fix linters and can merge

debebantur (approved):

PR #2003, opened 2025-06-05
DIRECT-IO Reader for PostgreSQL

vadv:
### Database name

PostgreSQL

# Pull request description

Problem is disk-cache flushing when using wal-g push-backup, this PR solves the problem.

Without patch:
CleanShot 2025-06-05 at 18 35 34@2x

With patch:
CleanShot 2025-06-05 at 18 36 49@2x

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -18,6 +18,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"github.com/wal-g/tracelog"
+
 	"github.com/wal-g/wal-g/internal/logging"
 	"github.com/wal-g/wal-g/internal/webserver"
 )
@@ -77,6 +78,7 @@ const (
 	PgpEnvelopeYcSaKeyFileSetting = "WALG_ENVELOPE_PGP_YC_SERVICE_ACCOUNT_KEY_FILE"
 	PgpEnvelopeYcEndpointSetting  = "WALG_ENVELOPE_PGP_YC_ENDPOINT"
 	PgpEnvelopeCacheExpiration    = "WALG_ENVELOPE_CACHE_EXPIRATION"
+	DirectIO                      = "WALG_DIRECT_IO"
 
 	PgDataSetting                        = "PGDATA"
 	UserSetting                          = "USER" // TODO : do something with it
@@ -271,6 +273,7 @@ var (
 		FailoverStoragesCheckTimeout: "30s",
 		FailoverStorageCacheLifetime: "15m",
 		PgpEnvelopeCacheExpiration:   "0",
+		DirectIO:                     "false",
 		LogLevelSetting:              "NORMAL",
 	}
 
@@ -367,6 +370,7 @@ var (
 		PgpEnvelopeYcKmsKeyIDSetting:  true,
 		PgpEnvelopeYcSaKeyFileSetting: true,
 		PgpEnvelopeYcEndpointSetting:  true,
+		DirectIO:                      false,
 		LibsodiumKeySetting:           true,
 		LibsodiumKeyPathSetting:       true,
 		LibsodiumKeyTransform:         true,
--- a/internal/fsutil/direct_io_reader.go
+++ b/internal/fsutil/direct_io_reader.go
@@ -67,13 +67,17 @@ func (r *reader) copyBuff(p []byte) int {
 }
 
 func (r *reader) Seek(offset int64, whence int) (int64, error) {
+	if offset < 0 {
+		panic(fmt.Errorf("this is programm bug, seek with negative offset is not supported by DirectIOReadSeekCloser"))
+	}
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	r.buffOffset = 0
 	r.buff = nil
 	r.alignedBlock = directio.AlignedBlock(directIOBlockCount * directio.BlockSize)
 	if whence != io.SeekStart {
-		return 0, fmt.Errorf("seek with whence %d currently is not supported by DirectIOReadSeekCloser", whence)
+		panic(fmt.Errorf("this is programm bug, seek with whence %d currently is not supported by DirectIOReadSeekCloser",
+			whence))
 	}
 	blocks := offset / directio.BlockSize
 	if n, errSeek := r.fd.Seek(blocks*directio.BlockSize, io.SeekStart); errSeek != nil {
@@ -82,6 +86,9 @@ func (r *reader) Seek(offset int64, whence int) (int64, error) {
 	nullBytes := make([]byte, offset-blocks*directio.BlockSize)
 	size, errRead := r.read(nullBytes)
 	if errRead != nil {
+		if errors.Is(errRead, io.EOF) && int64(size) < offset {
+			panic(fmt.Errorf("this is program bug, increase with seek is not supported by DirectIOReadSeekCloser"))
+		}
 		return int64(size), errRead
 	}
 	return offset, nil
--- a/internal/fsutil/direct_io_reader.go
+++ b/internal/fsutil/direct_io_reader.go
@@ -1,0 +1,95 @@
+package fsutil
+
+import (
+	"errors"
+	"io"
+	"os"
+	"sync"
+	"syscall"
+
+	"github.com/ncw/directio"
+)
+
+const directIOBlockCount = 32
+
+type reader struct {
+	mu           *sync.Mutex
+	fd           *os.File
+	buff         []byte
+	buffOffset   int
+	alignedBlock []byte
+}
+
+// NewDirectIOReadSeekCloserReadOnly returns read-only io.ReadSeekCloser.
+func NewDirectIOReadSeekCloserReadOnly(path string) (io.ReadSeekCloser, error) {
+	return NewDirectIOReadSeekCloser(path, syscall.O_RDONLY, 0)
+}
+
+// NewDirectIOReadSeekCloser returns io.ReadSeekCloser.
+func NewDirectIOReadSeekCloser(path string, flag int, perm os.FileMode) (io.ReadSeekCloser, error) {
+	in, errOpen := directio.OpenFile(path, flag, perm)
+	if errOpen != nil {
+		return nil, errOpen
+	}
+	return &reader{
+		mu:           &sync.Mutex{},
+		fd:           in,
+		buff:         nil,
+		alignedBlock: directio.AlignedBlock(directIOBlockCount * directio.BlockSize),
+	}, nil
+}
+
+func (r *reader) readBuff() error {
+	if n, err := io.ReadFull(r.fd, r.alignedBlock); err != nil {
+		r.buff = append(r.buff[r.buffOffset:], r.alignedBlock[0:n]...)
+		r.buffOffset = 0
+		if errors.Is(err, io.ErrUnexpectedEOF) {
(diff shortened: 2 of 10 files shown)


internal/tar_ball_file_packer.go
```diff
@@ -76,7 +78,7 @@ func (p *RegularTarBallFilePacker) PackFileIntoTar(cfi *ComposeFileInfo, tarBall
 // TODO : unit tests
 func StartReadingFile(fileInfoHeader *tar.Header, info os.FileInfo, path string) (io.ReadSeekCloser, error) {
 	fileInfoHeader.Size = info.Size()
-	file, err := os.Open(path)
+	file, err := fsutil.NewDirectIOReadSeekCloserReadOnly(path)
```
mialinx: It may affect performance, should be explicitly enabled with config option.

mialinx (changes requested):

internal/tar_ball_file_packer.go
vadv: Yes, this direct-io reader have cpu-penalty (~ 10-20% CPU depends on const `directIOBlockCount`) compared with `io.Open` . I made it just now disabled by default.

fanfuxiaoran:
Hi @vadv  Without patch, the disk-system-hit goes down sometime in the picture, does it mean that it will affect the postgres database performance? And in this pr, you make the wal-g to use directIO that let it not to  impact the os cache when doing 'push-backp'.  If I misunderstand, please correct me.

vadv:
Hi @fanfuxiaoran !

>  the disk-system-hit goes down sometime in the picture, does it mean that it will affect the postgres database performance?

This graph shows the changes in the `/proc/<pid>/io` of PostgreSQL backend processes (`rchar` - physically read  vs `read_bytes` - read physically and logically from storage). This graph demonstrates that PostgreSQL is running stable and is not experiencing any problems with the current backup process. The query p99 from the database pooler improved 4 times compared to vanila wal-g. 

We have tens of thousands of databases in our system, and the process of backing them up can have a significant impact on performance. Therefore, we used additional throttling tools for wal-g (cgroups). However, thanks to this patch, the backup process for particularly sensitive databases is now performed fast and without using cgroups for throttling.

> And in this pr, ... only 'push-backp'.

Yes, but now this code can be used to solve other tasks in which Direct-IO could help.

vadv:
tests `etcd_test` failed: https://github.com/wal-g/wal-g/actions/runs/15471982238/job/43596156079

```
wal-g_etcd_tests  | INFO: 2025/06/06 06:49:46.647643 LATEST backup is: 'stream_20250606T064946Z'
wal-g_etcd_tests  | ERROR: 2025/06/06 06:49:46.656167 Restore command output:
wal-g_etcd_tests  | Error: unknown flag: --data-dir
```

internal/fsutil/direct_io_reader.go
```diff
@@ -0,0 +1,98 @@
...
+	r.buffOffset += n
+	return n
+}
+
+func (r *reader) Seek(offset int64, whence int) (int64, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.buffOffset = 0
+	r.buff = nil
+	r.alignedBlock = make([]byte, len(r.alignedBlock))
```
ostinru: should we use `directio.AlignedBlock()` here as well (or just zero all values in the original slice)?

vadv: Thank you! If there was a seek operation, we would have received an error message about incorrectly allocated memory. Also, if the offset is not a multiple of 4096, it would be an error. I've already fixed that issue, but I still need to add some more test cases to ensure that everything works as expected (error compability as much as possible).

internal/config/config.go
```diff
@@ -77,6 +78,7 @@ const (
 	PgpEnvelopeYcSaKeyFileSetting = "WALG_ENVELOPE_PGP_YC_SERVICE_ACCOUNT_KEY_FILE"
 	PgpEnvelopeYcEndpointSetting  = "WALG_ENVELOPE_PGP_YC_ENDPOINT"
 	PgpEnvelopeCacheExpiration    = "WALG_ENVELOPE_CACHE_EXPIRATION"
+	DirectIO                      = "WALG_DIRECT_IO"
```
ostinru, line 81: Feel free to add it to documentations (`docs/PostgreSQL.md`) unless it too experimental to show to people.

vadv: Added docs as experimental fea.

boosterKRD:
FYI

We should be aware that using Direct I/O introduces more race conditions, since we’re reading files directly while PostgreSQL might still be writing to them. Currently, we already see some false positives in data checksum validation due to torn pages (e.g., 4KB written, 4KB not yet), and Direct I/O can increase these — not only with **Invalid Checksum**, but also with **Invalid Page Header**.

Hopefully, this won’t result in tens of thousands of warning messages. In general, we’ll need to filter out those false positives eventually — but that’s out of scope for this PR.
_Pages with an LSN higher than the backup start LSN will be fixed during WAL recovery anyway, so such errors are safe to ignore._

boosterKRD:
> FYI
> 
> We should be aware that using Direct I/O introduces more race conditions, since we’re reading files directly while PostgreSQL might still be writing to them. Currently, we already see some false positives in data checksum validation due to torn pages (e.g., 4KB written, 4KB not yet), and Direct I/O can increase these — not only with **Invalid Checksum**, but also with **Invalid Page Header**.
> 
> Hopefully, this won’t result in tens of thousands of warning messages. In general, we’ll need to filter out those false positives eventually — but that’s out of scope for this PR. _Pages with an LSN higher than the backup start LSN will be fixed during WAL recovery anyway, so such errors are safe to ignore._

Just to clarify my previous comment — I wasn’t entirely precise.

What I meant is that, in theory, we could skip checksum validation or ignore checksum errors if the page LSN is higher than the backup start LSN, since such pages will be safely restored during WAL recovery.

However, the situation with an Invalid Page Header is more problematic. If the header itself is corrupted, we can’t even read the LSN, so we have no way to determine whether the page is recoverable. This is especially risky with Direct I/O, where even the first 24 bytes (the header) might contain garbage if a write is in progress.

ostinru:
> We should be aware that using Direct I/O introduces more race conditions, since we’re reading files directly while PostgreSQL might still be writing to them.

I have failed to find any information on whether `O_DIRECT` `read()` uses dirty pages from page cache or read directly from disk.

But I think that we can assume that single sector writes (512 bytes) are atomic in modern hardware [1] and in all filesystems[2] after FAT32. Then, Page Header should be written atomically and we will not observe **Invalid Page Header**.

PS: SQLite doesn't[3] assume atomicy, but they are working on tosters and kettles.

[1] https://stackoverflow.com/questions/2009063/are-disk-sector-writes-atomic/61832882#61832882
[2] refer to Table 1 in https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-pillai.pdf
[3] https://www.sqlite.org/atomiccommit.html

boosterKRD:
> I have failed to find any information on whether `O_DIRECT` `read()` uses dirty pages from page cache or read directly from disk.
> 
> But I think that we can assume that single sector writes (512 bytes) are atomic in modern hardware [1] and in all filesystems[2] after FAT32. Then, Page Header should be written atomically and we will not observe **Invalid Page Header**.
> 
> PS: SQLite doesn't[3] assume atomicy, but they are working on tosters and kettles.
> 
> [1] https://stackoverflow.com/questions/2009063/are-disk-sector-writes-atomic/61832882#61832882 [2] refer to Table 1 in https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-pillai.pdf [3] https://www.sqlite.org/atomiccommit.html

👍🏻  @ostinru Thanks for the provided links. I agree, my concerns were unfounded.  👍🏻 


_The nitpicker mode is turned on :)_

The idea of bypassing the OS page cache is great, but I still have doubts regarding the implementation.
The core library here is github.com/ncw/directio v1.0.5 — it was last updated 11 years ago and has only 3 contributors. Maybe that's not critical, since it's mostly a thin wrapper around OS syscalls.
They claim this library supports all OSes (Unix, Windows, macOS), but considering the above, I think it should be tested carefully on all platforms where we expect to run WAL-G (**by the way, do we actually support Windows?**).

_The nitpicker mode is turned off :)_

_And now, the “schizophrenia mode” is on:_
But really, since this option (WALG_DIRECT_IO) is disabled by default and clearly marked as experimental, it’s perfectly fine to just let users try it out and report any problems they hit — along with their OS and filesystem details. If anything breaks, they can always turn it off. So maybe there’s no need for paranoid cross-platform testing just yet.

In short: both modes are right. :D   - @ostinru, @vadv, what do you opt for?

fanfuxiaoran:
> > FYI
> > We should be aware that using Direct I/O introduces more race conditions, since we’re reading files directly while PostgreSQL might still be writing to them. Currently, we already see some false positives in data checksum validation due to torn pages (e.g., 4KB written, 4KB not yet), and Direct I/O can increase these — not only with **Invalid Checksum**, but also with **Invalid Page Header**.
> > Hopefully, this won’t result in tens of thousands of warning messages. In general, we’ll need to filter out those false positives eventually — but that’s out of scope for this PR. _Pages with an LSN higher than the backup start LSN will be fixed during WAL recovery anyway, so such errors are safe to ignore._
> 
> Just to clarify my previous comment — I wasn’t entirely precise.
> 
> What I meant is that, in theory, we could skip checksum validation or ignore checksum errors if the page LSN is higher than the backup start LSN, since such pages will be safely restored during WAL recovery.
> 
I wonder if it is safe to skip checksum validation.  If there is a case that the page is partially flushed , and its LSN is higher than the backup start LSN,  but is the highest LSN between backup start and backup end for that page, so during WAL recovery, no wal log will be applied on it (if the page LSN is higher than the WAL log LSN, the WAL log record wont' be applied) , so it can not be fixed.
> However, the situation with an Invalid Page Header is more problematic. If the header itself is corrupted, we can’t even read the LSN, so we have no way to determine whether the page is recoverable. This is especially risky with Direct I/O, where even the first 24 bytes (the header) might contain garbage if a write is in progress.

gsmolk:
@fanfuxiaoran, are you referring to "torn write" problem? If so, it is covered by `full_page_writes`, and PostgreSQL backup is always taken with `full_page_writes` switched on.
https://wiki.postgresql.org/wiki/Full_page_writes

fanfuxiaoran:
> @fanfuxiaoran, are you referring to "torn write" problem? If so, it is covered by `full_page_writes`, and PostgreSQL backup is always taken with `full_page_writes` switched on. https://wiki.postgresql.org/wiki/Full_page_writes

@gsmolk Yes, I mean the "torn write".  You're right, the `full_page_writes` is turned on during backup, we don't need to worry about it here.

ostinru (approved):

ostinru:
Overall - looks good. I failed to find a scenario where this code will produce invalid result.
We can wait for more experienced reviewer, or merge this change and wait some time to test it in production on "thousands of databases".

vadv:
> wait some time to test it in production on "thousands of databases".

At the moment, this version works with 126+80 PostgreSQL databases, as they were particularly sensitive. We have successfully restored the databases from the staging production environment and deanonymized the data. This process is automated and is carried out every Monday with PITR.

Over the next two weeks, we plan to deploy this version on all of our databases, which are thousands.

PR #2004, opened 2025-06-11
Refactor PG connection from GP stuff

x4m:
Users (me) are very confised to see pg_session_role errors when connecting to Postgres.

### DIFF
--- a/internal/databases/greenplum/restore_point.go
+++ b/internal/databases/greenplum/restore_point.go
@@ -3,11 +3,12 @@ package greenplum
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"os"
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/greenplum-db/gp-common-go-libs/cluster"
 
 	"github.com/wal-g/wal-g/internal/multistorage"
 	"github.com/wal-g/wal-g/pkg/storages/storage"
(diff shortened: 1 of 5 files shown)


x4m:
I'd appreciate a help on finishing this. I struggle to find a time to diff into things that seems to be linter problems.

internal/databases/greenplum/restore_point.go
```diff
@@ -3,12 +3,13 @@ package greenplum
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/greenplum-db/gp-common-go-libs/cluster"
 	"os"
 	"sort"
 	"strings"
 	"time"
 
+	"github.com/greenplum-db/gp-common-go-libs/cluster"
```
leborchuk, line 11: Maybe we should start use of https://github.com/apache/cloudberry-go-libs ?

internal/databases/postgres/connect.go
```diff
@@ -30,22 +29,11 @@ func Connect(configOptions ...func(config *pgx.ConnConfig) error) (*pgx.Conn, er
 
 	conn, err := pgx.ConnectConfig(context.TODO(), config)
 	if err != nil {
+		// Try with GP utility mode as a fallback (for GP/CB segment connections)
```
leborchuk: Since we have moved postgres/connect.go -> greenplum/connect.go, is it still needed here?

leborchuk (changes requested):

internal/databases/greenplum/restore_point.go
x4m: GenerateAndExecuteCommand() is used in this file..

internal/databases/postgres/connect.go
x4m: yeah, that's the leak again... I'll redo

PR #2008, opened 2025-06-18
changed tarballs and tarball_composers

Apelsin234:
add context
add error return instead of fatal

### Database name

# Pull request description

// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/internal/databases/mongo/binary/mongod.go
+++ b/internal/databases/mongo/binary/mongod.go
@@ -131,11 +131,11 @@ func (mongodService *MongodService) GetBackupCursor() (cursor *mongo.Cursor, err
 		if !backupCursorErrorIsRetried(err) {
 			return nil, err
 		}
-		if i < cursorCreateRetries {
-			minutes := time.Duration(i + 1)
-			tracelog.WarningLogger.Printf("%v. Sleep %d minutes and retry", err.Error(), minutes)
-			time.Sleep(time.Minute * minutes)
-		}
+
+		// logging retries
+		minutes := time.Duration(i + 1)
+		tracelog.WarningLogger.Printf("%v. Sleep %d minutes and retry", err.Error(), minutes)
+		time.Sleep(time.Minute * minutes)
 	}
 
 	return cursor, err
--- a/internal/databases/postgres/bundle.go
+++ b/internal/databases/postgres/bundle.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"archive/tar"
+	"context"
 	"fmt"
 	"io"
 	"io/fs"
@@ -102,7 +103,7 @@ func NewBundle(
 }
 
 func (bundle *Bundle) SetupComposer(composerMaker TarBallComposerMaker) (err error) {
-	tarBallComposer, err := composerMaker.Make(bundle)
+	tarBallComposer, err := composerMaker.Make(context.TODO(), bundle)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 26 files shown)


internal/databases/mongo/binary/mongod.go
```diff
@@ -131,11 +131,11 @@ func (mongodService *MongodService) GetBackupCursor() (cursor *mongo.Cursor, err
...
-		if i < cursorCreateRetries {
-			minutes := time.Duration(i + 1)
-			tracelog.WarningLogger.Printf("%v. Sleep %d minutes and retry", err.Error(), minutes)
-			time.Sleep(time.Minute * minutes)
-		}
+
+		// logging retries
+		minutes := time.Duration(i + 1)
+		tracelog.WarningLogger.Printf("%v. Sleep %d minutes and retry", err.Error(), minutes)
+		time.Sleep(time.Minute * minutes)
```
VGoshev, line 138: looks like with `i >= cursorCreateRetries` we need to log only error, without message about new retry and sleep (as we won't retry anymore)

internal/databases/postgres/bundle.go
```diff
@@ -102,7 +103,7 @@ func NewBundle(
 }
 
 func (bundle *Bundle) SetupComposer(composerMaker TarBallComposerMaker) (err error) {
-	tarBallComposer, err := composerMaker.Make(bundle)
+	tarBallComposer, err := composerMaker.Make(context.TODO(), bundle)
```
VGoshev, line 106: Do we really want to use `context.TODO()` here?

internal/databases/postgres/dir_database_tar_ball_composer.go
```diff
@@ -26,7 +27,7 @@ func NewDirDatabaseTarBallComposerMaker(files internal.BundleFiles, filePackerOp
 	}
 }
 
-func (m DirDatabaseTarBallComposerMaker) Make(bundle *Bundle) (internal.TarBallComposer, error) {
+func (m DirDatabaseTarBallComposerMaker) Make(_ context.Context, bundle *Bundle) (internal.TarBallComposer, error) {
```
VGoshev, line 30: New variable, which we don't use. Is it really needed?

internal/directory_uploader.go
```diff
@@ -49,7 +50,7 @@ func (u *CommonDirectoryUploader) Upload(path string) TarFileSets {
 	err := bundle.StartQueue(NewStorageTarBallMaker(u.backupName, u.uploader))
 	tracelog.ErrorLogger.FatalOnError(err)
 
-	err = bundle.SetupComposer(u.tarBallComposerMaker)
+	err = bundle.SetupComposer(context.TODO(), u.tarBallComposerMaker)
```
VGoshev, line 53: yet another `context.TODO()`

internal/tar_ball_test.go
```diff
@@ -47,7 +48,9 @@ func TestS3DependentFunctions(t *testing.T) {
 
 	tarBallQueue.NewTarBall(false)
 	tarBall := tarBallQueue.LastCreatedTarball
-	tarBall.SetUp(nil)
+	if err := tarBall.SetUp(context.TODO(), nil); err != nil {
```
VGoshev, line 51: Don't we want to use context.Background() instead? (But in tests, it doesn't look like it really matters.)

VGoshev (changes requested):

PR #2013, opened 2025-06-24
Add `HandleDetailedBackupList` table output unit-tests

BasedDepartment1:
### Database name
Greenplum

# Pull request description
Add table output unit-tests for `HandleDetailedBackupList`.

The name of the `name` column in pretty table mode was in fact `NAME TIME`. Now it is just `NAME`.

### DIFF
--- a/internal/databases/greenplum/backup_list_handler.go
+++ b/internal/databases/greenplum/backup_list_handler.go
@@ -69,7 +69,7 @@ func (bd *BackupDetail) PrintableFields() []printlist.TableField {
 	return []printlist.TableField{
 		{
 			Name:       "name",
-			PrettyName: "Name time",
+			PrettyName: "Name",
 			Value:      bd.Name,
 		},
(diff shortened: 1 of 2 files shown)


internal/databases/greenplum/backup_list_handler.go
```diff
@@ -69,7 +69,7 @@ func (bd *BackupDetail) PrintableFields() []printlist.TableField {
 	return []printlist.TableField{
 		{
 			Name:       "name",
-			PrettyName: "Name time",
+			PrettyName: "Name",
```
BasedDepartment1, line 72: I suppose this is a bug and the column `name` should be called `NAME` in pretty mode, but if previous name was intentional, I can change it back.

ostinru: No other database uses this weird name - so it is ok to change to what everyone else uses.

PR #2022, opened 2025-07-14
feat: add alicloud OSS storage implementation

imrenagi:
### Database name
Alicloud OSS storage implementation

# Pull request description

While wal-g supports S3 compatible interface for file download and upload, we cant really use S3 storage implementation to donwload and upload walg files to Alicloud OSS if we want to use alicloud STS to assume a particular RAM role. 

In my case, I have Alicloud RAM user which only have permission to assume few RAM Roles. This RAM role is the one which have read/write access to the bucket objects. This cant be achieved if I use S3 implementation, hence I decided to add storage implementation for alicloud OSS.

* Provide this env var.
```
OSS_ACCESS_KEY_ID=
OSS_ACCESS_KEY_SECRET=
OSS_REGION=

// You can use the config below if you want to assume a RAM role.
#OSS_ROLE_ARN=
#OSS_ROLE_SESSION_NAME=
```

WALG_OSS_PREFIX=oss://<bucket>/<path>
```

* Then you can run the walg command as usual.

### CODE REVIEW

**Comment by imrenagi** [comment]
hi @x4m @usernamedt , do you know someone who can review this MR? thanks in advance!

File: docs/STORAGES.md [hunk 1]
```diff
@@ -156,6 +156,30 @@ Overrides the default `upload buffer size` of 8388608 bytes (8 MB). Note that th
 
 Overrides the default `maximum number of upload buffers`. By default, at most 4 buffers are used concurrently.
 
+Alicloud
```
walbertus: Alicloud OSS make more sense with the env variable

internal/config/config.go
```diff
@@ -228,6 +228,10 @@ const (
 
 	GoogleApplicationCredentials = "GOOGLE_APPLICATION_CREDENTIALS"
 
+	AliyunAccessKeyID     = "OSS_ACCESS_KEY_ID"
```
walbertus: using Aliyun and Alicloud create inconsistency in naming, I suggest it should use alicloud everywhere

x4m (approved):
Looks great! It would be very nice to have this feature!

x4m:
Thanks @walbertus for the review. FWIW I'm fine with merging as-is.
But after release we will not be able to change setting names. Are you sure in current naming (WALG_OSS_PREFIX)? Is OSS common abbreviation for Alicloud OSS?

imrenagi:
Thanks @x4m ! 

I have fixed the linter issue. However, the [parallel build job](https://github.com/wal-g/wal-g/actions/runs/16286663087) on the github action pipeline doesnt seems related to my changes. Any pointer on this?

Also, related to the naming convention. Yes, Alicloud is using OSS abbreviation for its object storage solution. Here is their [product page](https://www.alibabacloud.com/en/product/object-storage-service?_p_lc=1) for reference.

x4m:
On their product page everything is Alicloud... :) but if we do not have other options\ideas let's merge as is.

imrenagi:
@x4m I tried to fix the docker pipeline issue by updating `debian:buster` into `debian:bullseye` for the golang image. Looks like buster is too old and some configuration is already missing there. After this change, the same error doesn't appear anymore on my local. 

Also, related to the setting names. When working on this, I found out that alicloud doesnt use consistent naming for their env var prefix. For instance, in their golang OSS sdk, they use `OSS_ACCESS_KEY`. This is key that they will read automatically on the `LoadConfig()` function i used on the PR. But on other sdk or their documentation, they are also using `ALIBABA_CLOUD_ACCESS_KEY`. Since we are focusing more on the storage here, I decided to go with OSS only.

For WALG_OSS_PREFIX, i really think that we should use OSS here since the storage object itself is called OSS. Also, other env var settings used in this MR are indeed configuration used for OSS only. If we have needs to provide configuration for non-OSS specific (e.g. customizing the STS endpoint, which wont be possibly needed), I think we can do something like `ALICLOUD_STS_ENDPOINT` or `ALIBABA_CLOUD_STS_ENDPOINT` later.

imrenagi:
@x4m I found a bug on when doing delta backup. Let me fix that first. Thanks!

ostinru:
> @x4m I tried to fix the docker pipeline issue by updating `debian:buster` into `debian:bullseye` for the golang image. Looks like buster is too old and some configuration is already missing there. After this change, the same error doesn't appear anymore on my local.

Ok, let's check how all tests works with debian:bullseye... we had an issue with libc compatibility between debian <> ubuntu bionic.

imrenagi:
@ostinru the image build was success, but looks like other jobs using the built docker image are failing. Any pointer on how to fix this?

ostinru:
> @ostinru the image build was success, but looks like other jobs using the built docker image are failing. Any pointer on how to fix this?

Ok... 
debian:buster repositories are not available anymore
debian:bullseye is not compatible with ubuntu 18.04:

> wal-g: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by wal-g)

Let's try to use ubuntu everywhere: 
https://github.com/wal-g/wal-g/pull/2025

imrenagi:
Sure. Looks like it works. Will wait for #2025 fix to be merged and I will rebase from there. Thanks @ostinru !

ostinru:
> Sure. Looks like it works. Will wait for #2025 fix to be merged and I will rebase from there. Thanks @ostinru !

Just merged fix. Let's rebase it on master.

imrenagi:
I just rebase the branch. @ostinru @x4m can you please help to trigger the pipeline? thanks!

ostinru:
@imrenagi please also add secret config keys to `secretSettings`  `internal/config/config.go:643`. Otherwise secrets will end up in logs and users will attach it to github bug reports 🙈

imrenagi:
@ostinru Done. Thanks for noticing that. Im not aware of such feature. =)

ostinru (approved):
LGTM

ostinru:
@x4m - let's merge this when tests are good.
// I am on vacation, will not follow up this tread.

x4m:
Thanks everyone! Pushed.

PR #2042, opened 2025-08-19
Second attempt to fix SignatureDoesNotMatch

ostinru:
### Greenplum

previous attempt: https://github.com/wal-g/wal-g/pull/2036

From time to time we are getting HTTP 403 SignatureDoesNotMatch error:
```
ERROR: 2025/08/05 05:04:22.053632 failed to upload 'wal-e/mydatabase/6/segments_005/seg12/basebackups_005/base_000000190000791300000002/tar_partitions/part_512.tar.br' to bucket 'mydatabase': MultipartUpload: upload multipart failed
        upload id: 000FFB97256FFFFF
caused by: SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.
        status code: 403, request id: 8fda35753dffffff, host id:
ERROR: 2025/08/05 05:04:22.053655 Unable to continue the backup process because of the loss of a part 512.
ERROR: 2025/08/05 05:04:23.531918 command failed: exit status 1
```

In this PR I am trying to hide cryptic error with better retry.

### DIFF
--- a/pkg/storages/s3/retryer.go
+++ b/pkg/storages/s3/retryer.go
@@ -25,7 +25,12 @@ func (r ConnResetRetryer) ShouldRetry(req *request.Request) bool {
 	if req.Error != nil && strings.Contains(req.Error.Error(), "SignatureDoesNotMatch") {
 		// It looks like we have some rare issues with request. Sign one more time
 		auth := r.getAuthHeader(req)
+
+		stash := req.Error
+		req.Error = nil // hide req.Error - so, req.Sign() will not fail
 		err := req.Sign()
+		req.Error = stash
+
 		if err != nil {
 			tracelog.ErrorLogger.Printf("Cannot re-sign request: %v", err)
 			return false


wolandspb:
I can confirm we've seen this exact same issue multiple times. It happens in our environment with both PostgreSQL and Greenplum, specifically when we use wal-g in combination with ceph-radosgw.

We've found this error typically occurs under two conditions: an unstable network connection (for example, when a backup is being made between different data centers) or high load on Ceph (for example, during a recovery operation after a node failure). We have caught this error repeatedly, sometimes on dozens of databases at the same time.

My current theory is that this problem happens because a new part of the multipart upload starts before the previous part has successfully finished uploading. The older part may be attempting a retry at the exact moment the new one begins.

I considered adding logic to retry 403 errors within pkg/storages/s3/retryer.go, but I've come to believe that this would likely not resolve the issue.

ostinru:
> We've found this error typically occurs under two conditions: an unstable network connection (for example, when a backup is being made between different data centers) or high load on Ceph (for example, during a recovery operation after a node failure). We have caught this error repeatedly, sometimes on dozens of databases at the same time.

We observing this issue when S3 responses with 503 (Service Temporarily Unavailable). Then, all retries returns 403.

Currently, I am going to test workaround with setting `S3_DISABLE_100_CONTINUE: true`. We will collect some reliable statistics in a week or two. Currently, we can be sure that this PR haven't fixed issue.

wolandspb:
Yes, in cases where Ceph is under high load, we can also get a 500 error from radosgw or nginx, and then, as you wrote, all other uploads receive a 403 error.
But the case with another data center is more interesting. There have been instances where the connection between wal-g and ceph was interrupted by haproxy in the middle due to the long duration of the transfer, and after that, all other uploads also started receiving a 403 error.
I hope some of the information above helps you with debugging.

ostinru:
`S3_DISABLE_100_CONTINUE: true` fixed an issue. No changes required.

PR #2046, opened 2025-08-23
fix cross building of sql server

chipitsine:
This pull request — let me tell you — it’s the best one today. People are talking about it. Developers love it. It’s winning, big league!

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,13 +4,11 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"os/signal"
 	"os/user"
 	"runtime"
 	"sort"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


ostinru:
Am I correct that wal-g logging could be reconfigured at runtime by sending SIGUSR1 signal? Sounds like undocumented feature. Should test it...

About this patch. What do you think about having 2 versions of this function? One for Linux and one for all other systems? (like `PunchHole` in https://github.com/wal-g/wal-g/tree/master/internal/ioextensions )

chipitsine:
I tend to the most simple approach.

first, you are totally correct about undocumented log refresh feature.
the only thing - that is the feature of pg-daemon.

no need to expose it to other databases. let it live somewhere in pg-daemon code.
when pg-daemon will decide to go windows or macos, you suggestion to split the source files makes sense

ostinru:
> no need to expose it to other databases. let it live somewhere in pg-daemon code.

I would say that it could be useful to expose it to all databases: `wal-g` could run for hours and when it stuck - it is good idea to turn logs to `DEVEL` level. Alternatevely - expose /pprof/ endpoints as well. However it is out of scope for this PR.

ostinru (approved):
LGTM

PR #2047, opened 2025-08-24
CI: add sql server cross build

chipitsine:
I acknowledge the addition of SQL Server compilation support, a step which—while long overdue—raises important questions about prior oversight and the broader implications for cross-platform compatibility. It is imperative that such developments be accompanied by transparent documentation and inclusive consultation.


ostinru (approved):
LGTM

PR #2049, opened 2025-08-26
add flag to count journal info

kronwerk:
### Database name
mongo

# Pull request description
add flag to count journal info

we need info about oplog sizes necessary to restore to specified point in time between binary backups - we are adding this info here (pls see mysql as a reference for analogue)

### DIFF
--- a/internal/databases/mongo/binary/backup.go
+++ b/internal/databases/mongo/binary/backup.go
@@ -43,16 +43,16 @@ type CalculateSizesArgs struct {
 	CountJournals bool
 }
 
-func (backupService *BackupService) createInitialJournals() internal.JournalInfo {
+func (backupService *BackupService) createInitialJournals(journalFiles internal.JournalFiles) internal.JournalInfo {
 	backupFolder, err := common.GetBackupFolder()
 	if err != nil {
-		tracelog.ErrorLogger.Printf("can not get backup folder: %s", err.Error())
+		tracelog.ErrorLogger.Printf("can not get backup folder: %+v", err)
 		return internal.JournalInfo{}
 	}
 	backupTimes, err := internal.GetBackups(backupFolder)
 	if err != nil {
 		// no backups is a valid case, no journals should be created then
-		tracelog.WarningLogger.Printf("can not get backups: %s", err.Error())
+		tracelog.WarningLogger.Printf("can not get backups: %+v", err)
 		return internal.JournalInfo{}
 	}
 
@@ -64,6 +64,7 @@ func (backupService *BackupService) createInitialJournals() internal.JournalInfo
 			backupName:            backupTime.BackupName,
 			mostRecentJournalInfo: mostRecentJournalInfo,
 			timeStop:              backupTime.Time,
+			journalFiles:          journalFiles,
 		})
 	}
 	return mostRecentJournalInfo
@@ -73,6 +74,7 @@ type addJournalInfoArgs struct {
 	backupName            string
 	mostRecentJournalInfo internal.JournalInfo
 	timeStop              time.Time
+	journalFiles          internal.JournalFiles
 }
 
 func (backupService *BackupService) addJournalInfo(args addJournalInfoArgs) internal.JournalInfo {
@@ -82,7 +84,7 @@ func (backupService *BackupService) addJournalInfo(args addJournalInfoArgs) inte
 
 	storage, err := internal.ConfigureStorage()
 	if err != nil {
-		tracelog.WarningLogger.Printf("Can't configure storage: %s", err.Error())
+		tracelog.WarningLogger.Printf("Can't configure storage: %+v", err)
 		return internal.JournalInfo{}
 	}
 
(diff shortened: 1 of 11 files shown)


internal/databases/mongo/binary/backup.go
```diff
@@ -35,20 +38,136 @@ func CreateBackupService(ctx context.Context, mongodService *MongodService, uplo
...
+	rootFolder := storage.RootFolder()
+	journalInfo := internal.NewEmptyJournalInfo(
+		args.backupName,
+		args.mostRecentJournalInfo.CurrentBackupEnd, args.timeStop,
+		models.OplogArchBasePath,
+	)
+
+	err = journalInfo.Upload(rootFolder)
+	if err != nil {
+		tracelog.WarningLogger.Printf("can not upload the journal info: %s", err.Error())
```
VGoshev: Maybe it would be better to use %w + err? There and in other places

internal/databases/mongo/binary/backup.go
```diff
@@ -35,20 +38,136 @@ func CreateBackupService(ctx context.Context, mongodService *MongodService, uplo
...
+	}
+
+	mostRecentJournalInfo, err := internal.GetMostRecentJournalInfo(
+		storage.RootFolder(),
+		models.OplogArchBasePath,
+	)
+	if err != nil {
+		// there can be no backups on S3 or we do it first time
+		tracelog.WarningLogger.Printf("can not find the last journal info: %s", err.Error())
+		mostRecentJournalInfo = backupService.createInitialJournals()
```
noname0443: Could we consider invoking `createInitialJournals` only for the first backup? This method can be expensive when there are many backups with a large number of oplogs, due to the oplog listing here: https://github.com/wal-g/wal-g/blob/167b38659c443863a829029ecb17690ca41ef603/internal/journal.go#L231

I suggest introducing a dedicated error type (similar to this one: https://github.com/wal-g/wal-g/blob/167b38659c443863a829029ecb17690ca41ef603/internal/journal.go#L22) and using it at https://github.com/wal-g/wal-g/blob/167b38659c443863a829029ecb17690ca41ef603/internal/journal.go#L211; that way, we wouldn't trigger `createInitialJournals` on S3-related errors.

internal/databases/mongo/binary/backup.go
```diff
@@ -35,20 +38,136 @@ func CreateBackupService(ctx context.Context, mongodService *MongodService, uplo
...
+		models.OplogArchBasePath,
+	)
+
+	err = journalInfo.Upload(rootFolder)
+	if err != nil {
+		tracelog.WarningLogger.Printf("can not upload the journal info: %s", err.Error())
+		return internal.JournalInfo{}
+	}
+
+	err = journalInfo.UpdateIntervalSize(rootFolder)
```
noname0443: I'm concerned that `UpdateIntervalSize` may not be intended for repeated use in a loop, given the potentially expensive oplog listing.

Instead, could we perform a single initial oplog listing in `BackupService` and add a method like `UpdateIntervalSizeWithJournalFiles` that accepts the pre-listed journal files, avoiding a fresh listing on each call?

internal/databases/mongo/binary/backup.go
```diff
@@ -102,14 +221,18 @@ func (backupService *BackupService) DoBackup(backupName string, permanent, skipM
...
 			return err
 		}
 	}
 
-	return backupService.Finalize(concurrentUploader, backupCursor.BackupCursorMeta)
+	err = backupService.Finalize(concurrentUploader, backupCursor.BackupCursorMeta, args.CountJournals)
+	if err != nil {
+		return err
+	}
+	return nil
```
noname0443: Nit: we could simplify to `return backupService.Finalize(...)`. Totally fine to keep the explicit version if that’s the preferred style.

internal/journal.go
```diff
@@ -225,13 +227,26 @@ func GetMostRecentJournalInfo(
...
+	initialized bool
+}
+
 // UpdateIntervalSize calculates the size of the SizeToNextBackup in the semi-interval (PriorBackupEnd; CurrentBackupEnd]
 // using journal files on JournalDirectoryName and save it for the previous JournalInfo
-func (ji *JournalInfo) UpdateIntervalSize(folder storage.Folder) error {
-	journalFiles, _, err := folder.GetSubFolder(ji.JournalDirectoryName).ListFolder()
-	if err != nil {
-		return err
+func (ji *JournalInfo) UpdateIntervalSize(folder storage.Folder, journalFilesObj JournalFiles) error {
```
noname0443: Would it make sense to pass `JournalFiles` by pointer (i.e. `journalFilesObj *JournalFiles`) so we don’t copy the struct? Otherwise, updates to `initialized`/`files` won’t persist outside the function, so `initialized` will effectively remain false across calls.

noname0443 (changes requested):

noname0443 (approved):
Looks good to me!

VGoshev (approved):

PR #2056, opened 2025-09-02
PostgreSQL & Greenplum: Reduce memory usage

ostinru:
### Greenplum / Postgres

`wal-g` may use A LOT of RAM during files metadata upload (hundreds of megabytes). However we don't need to keep whole JSON in memory - we can stream it with new golang JSON v2 API.


internal/databases/postgres/backup_push_handler.go
```diff
@@ -629,17 +632,32 @@ func (bh *BackupHandler) uploadExtendedMetadata(ctx context.Context, meta Extend
...
+	})
+	errorGroup.Go(func() error {
+		err := bh.Arguments.Uploader.Upload(ctx, getFilesMetadataPath(bh.CurBackupInfo.Name), reader)
+		if err != nil {
+			_ = reader.CloseWithError(err)
+			return err
+		}
+		return reader.Close()
+	})
+	return errorGroup.Wait()
```
ostinru: FIXME: move to new method `Uploader.UploadJSON()` ?

ostinru:
Tested. Works well.

1. waiting for golang 1.25.2+ to bump go version
2. cleaning up this PR
3. merge

internal/uploader.go
```diff
@@ -36,9 +47,40 @@ type Uploader interface {
...
+		}
+		return reader.Close()
+	})
+	return errorGroup.Wait()
+}
+
 // RegularUploader contains fields associated with uploading tarballs.
 // Multiple tarballs can share one Uploader.
 type RegularUploader struct {
+	DefaultJSONUploader
```
vovik0134: It would be better to define separate field and initialize it in `New` method (not sure if it exists). Something like that:
```go
type RegularUploader strust {
    jsonUploader jsonUploader
    ...
}

func New() *RegularUploader {
    return &RegularUploader{
        jsonUploader: NewDefaultJSONUploader(),
        ...
    }
}
```

Motivation:
- (Most important) While promoting composition, embedding can still lead to a degree of coupling between the outer and embedded structs, as changes in the embedded struct's public API can affect the outer struct.
- If the outer struct defines fields or methods with the same names as those in an embedded struct, the outer struct's definitions will "shadow" the embedded ones. This can lead to unexpected behavior if not handled carefully.
- While methods can be "overridden" by defining a method with the same signature in the outer struct, this is not true inheritance-style overriding. The original embedded method is still present and can be accessed via the embedded field name.

go.mod
```diff
@@ -53,9 +53,6 @@ require (
 
 require (
 	github.com/ProtonMail/go-crypto v0.0.0-20230426101702-58e86b294756
-	github.com/alibabacloud-go/darabonba-openapi/v2 v2.1.8
-	github.com/alibabacloud-go/sts-20150401/v2 v2.0.4
-	github.com/alibabacloud-go/tea v1.3.9
```
ostinru, line 58: unused for some reason =/

debebantur (approved):

ostinru:
Surprisinly, we already had StreamingJSON (disabled by default)
https://github.com/wal-g/wal-g/blob/v3.0.7/internal/serializer.go#L62-L86

PR #2063, opened 2025-09-12
Greenplum follow-primary: use absolute path to pg_ctl

ostinru:
Minor changes to allow `wal-g follow-primary` work in distributed environment


ostinru:
cc: @Mulily0513 as original author & user of this code

PR #2075, opened 2025-10-02
Add: a reconnection to the binlog server

zkramorev:
## What this PR fixes:
This PR fixes an issue where the WAL-G binlog server could only handle a single connection at a time. When restoring multiple MySQL replicas simultaneously, the server would exit after processing the first connection, causing subsequent replicas to fail with connection errors. If we try to connect from the same host (reconnection), we won't be able to.

**Steps to reproduce:**  
- Cluster restore using wal-g binlog server

**Logs:**  
Error logs from failing replica

```bash
2025-10-02T06:40:53.931816-00:00 9 [ERROR] [MY-010584] [Repl] Replica I/O for channel 'binlog_server_channel': 
Error reconnecting to source 'root@127.0.0.1:4000'. This was attempt 85/86400, 
with a delay of 60 seconds between attempts. Message: Can't connect to MySQL server on '127.0.0.1:4000' (111), 
Error_code: MY-002003
```
```bash
INFO: 2025/10/02 05:15:44.471988 Listening on 127.0.0.1:4000, wait connection
INFO: 2025/10/02 05:15:53.603377 connection accepted
# Server exits after processing this single connection
```

### DIFF
--- a/internal/databases/mysql/binlog_server_handler.go
+++ b/internal/databases/mysql/binlog_server_handler.go
@@ -155,7 +155,7 @@ func sendEventsFromBinlogFiles(logFilesProvider *storage.ObjectProvider, pos mys
 			if err != nil {
 				tracelog.InfoLogger.Println("Error while waiting MySQL applied binlogs: ", err)
 			}
-			os.Exit(0)
+			return
 		}
 		handleEventError(err, s)
 		if err != nil {
@@ -282,22 +282,52 @@ func HandleBinlogServer(since string, until string) {
 	tracelog.ErrorLogger.FatalOnError(err)
 	tracelog.InfoLogger.Printf("Listening on %s, wait connection", l.Addr())
 
-	c, err := l.Accept()
-	tracelog.ErrorLogger.FatalOnError(err)
-	tracelog.InfoLogger.Printf("connection accepted")
-
-	user, err := conf.GetRequiredSetting(conf.MysqlBinlogServerUser)
-	tracelog.ErrorLogger.FatalOnError(err)
-	password, err := conf.GetRequiredSetting(conf.MysqlBinlogServerPassword)
-	tracelog.ErrorLogger.FatalOnError(err)
-	conn, err := server.NewConn(c, user, password, Handler{})
-	tracelog.ErrorLogger.FatalOnError(err)
-	tracelog.InfoLogger.Printf("connection created")
-
 	for {
-		if err := conn.HandleCommand(); err != nil {
-			tracelog.WarningLogger.Printf("Error handling command: %v", err)
-			break
+		c, err := l.Accept()
+		if err != nil {
+			tracelog.ErrorLogger.Printf("Error accepting connection: %v", err)
+			continue
+		}
+		tracelog.InfoLogger.Printf("Connection accepted from %s", c.RemoteAddr())
+
+		user, err := conf.GetRequiredSetting(conf.MysqlBinlogServerUser)
+		if err != nil {
+			tracelog.ErrorLogger.Printf("Failed to get user: %v", err)
+			err := c.Close()
+			if err != nil {
+				return
+			}
+			continue
+		}
+		password, err := conf.GetRequiredSetting(conf.MysqlBinlogServerPassword)
+		if err != nil {
+			tracelog.ErrorLogger.Printf("Failed to get password: %v", err)
+			err := c.Close()
+			if err != nil {
+				return
+			}
+			continue
+		}
+		conn, err := server.NewConn(c, user, password, Handler{})
+		if err != nil {
+			tracelog.ErrorLogger.Printf("Failed to create connection: %v", err)
+			err := c.Close()
+			if err != nil {
+				return
+			}
+			continue
+		}
+		tracelog.InfoLogger.Printf("MySQL connection created for %s", c.RemoteAddr())
+
+		for {
+			if err := conn.HandleCommand(); err != nil {
+				tracelog.WarningLogger.Printf("Error handling command from %s: %v", c.RemoteAddr(), err)
+				break
+			}
+		}
+		err = c.Close()
+		if err != nil {
+			return
 		}
 	}
 }


internal/databases/mysql/binlog_server_handler.go
```diff
@@ -26,9 +27,13 @@ import (
...
-	startTS      time.Time
-	untilTS      time.Time
-	lastSentGTID string
+	startTS        time.Time
+	untilTS        time.Time
+	lastSentGTID   string
+	syncStarted    bool
+	syncMutex      sync.Mutex
+	globalStreamer *replication.BinlogStreamer
+	streamerMutex  sync.Mutex
```
noname0443, line 38: Might we consider moving these variables into `Handler` entity?

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -282,22 +329,33 @@ func HandleBinlogServer(since string, until string) {
 	tracelog.ErrorLogger.FatalOnError(err)
 	tracelog.InfoLogger.Printf("Listening on %s, wait connection", l.Addr())
 
-	c, err := l.Accept()
-	tracelog.ErrorLogger.FatalOnError(err)
-	tracelog.InfoLogger.Printf("connection accepted")
+	for {
```
noname0443, line 327: Is it okay for the loop to run forever? Could we add a cancellation channel and a select statement to stop it, just in case?

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -200,26 +205,68 @@ func (h Handler) HandleRegisterSlave(data []byte) error {
...
+		tracelog.InfoLogger.Println("Returning existing streamer for reconnection")
+		return globalStreamer, nil
+	}
 
-	err := syncBinlogFiles(mysql.Position{Name: "host-binlog-file", Pos: 4}, startTS, s)
-	return s, err
+	globalStreamer = replication.NewBinlogStreamer()
+
+	syncMutex.Lock()
+	if !syncStarted {
```
noname0443: I think `sync.Once` is a more readable and idiomatic way to execute something in only one place. Could we consider using it?

zkramorev: Yes, it will be much better!

internal/databases/mysql/binlog_server_handler.go
zkramorev: of course we can add it, but right now we don't need to programmatically stop the server or process system signals. maybe it's a little overkill?

internal/databases/mysql/binlog_server_handler.go
zkramorev: startTS, untilTS, and lastSentGTID might be better left global for reconnections to work correctly. And globalStreamer and streamerMutex are really better transferred to the Handler structure. what do you think?

noname0443: I can't insist on changes since this isn't part of your PR :)

In my opinion, the problem with `startTS`, `untilTS`, and `lastSendGTID` lies in their scope and hidden dependencies:
1. We should keep variables in the lowest possible scope to prevent unintended access by other functions.
2. When reading the high-level code, it's not immediately obvious that we are modifying the module's global state, which could affect the function's behavior.

If you're in a hurry, let's leave it as is and revisit the refactoring later.

internal/databases/mysql/binlog_server_handler.go
noname0443: Ah, I see it now. The server is designed to stop after it sends all the binlogs - that wasn't obvious at all :( https://github.com/wal-g/wal-g/blob/f69f2e8397f3d2333d1996297c55e66f3da72973/internal/databases/mysql/binlog_server_handler.go#L166
Okay, let's leave it as is for now.

Here are my thoughts on what we could do to improve the code:
- Move the functions `syncBinlogFiles`, `sendEventsFromBinlogFiles`, and `waitReplicationIsDone` to be methods of the Handler struct. This would improve encapsulation.
- Replace all `FatalOnError` and `os.Exit(0)` calls with graceful error handling and a server shutdown mechanism. This would reduce number of side-effects.
- Control the lifecycle of the `go sendEventsFromBinlogFiles` and `go provideLogs goroutines`, rather than leaving them unmanaged.
This is important because the module is part of an internal package, which ideally shouldn't have side effects like terminating the entire server process.

Since you are already familiar with the module, it would be fantastic if you refactor the module in separate PRs. No pressure, of course! :)

noname0443 (approved):

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -282,22 +321,35 @@ func HandleBinlogServer(since string, until string) {
...
+
+		user, err := conf.GetRequiredSetting(conf.MysqlBinlogServerUser)
+		tracelog.ErrorLogger.FatalOnError(err)
+		password, err := conf.GetRequiredSetting(conf.MysqlBinlogServerPassword)
+		tracelog.ErrorLogger.FatalOnError(err)
+
+		conn, err := server.NewConn(c, user, password, globalHandler)
+		if err != nil {
+			tracelog.ErrorLogger.Printf("Error creating connection: %v", err)
+			c.Close()
```
ostinru, line 343: Let's wrap in `utility.LoggedClose()`

internal/databases/mysql/binlog_server_handler.go
ostinru, line 166: Probably, we should distinguish between successful and failed replay =/
`os.Exit(1)` in case of error sounds reasonable?

internal/databases/mysql/binlog_server_handler.go
ostinru: Let's add a comment that this loop will be stopped due to `os.Exit()` in `sendEventsFromBinlogFiles`. This will simplify understanding...

ostinru (approved):
Looks great!

PR #2076, opened 2025-10-02
tests: migrate from "bitnami" to "bitnamilegacy"

chipitsine:
💅 Bye Bitnami, we’re legacy now — migrated with grace, attitude, and zero tolerance for vendor lock-in.


ostinru:
Lol, Bitnami is part of Broadcom as well

PR #2080, opened 2025-10-04
prepare test scripts for Postgresql > 10

chipitsine:
currently there were a lot of hardcoded "10" version in tests. Let's carefully replace them

### DIFF
--- a/docker/pg_tests/scripts/tests/archiving_ready_rename.sh
+++ b/docker/pg_tests/scripts/tests/archiving_ready_rename.sh
@@ -3,7 +3,7 @@ set -e -x
 CONFIG_FILE="/tmp/configs/archiving_ready_rename.json"
 COMMON_CONFIG="/tmp/configs/common_config.json"
 TMP_CONFIG="/tmp/configs/tmp_config.json"
-export PGDATA="/var/lib/postgresql/10/main"
+
 cat ${CONFIG_FILE} > ${TMP_CONFIG}
 echo "," >> ${TMP_CONFIG}
 cat ${COMMON_CONFIG} >> ${TMP_CONFIG}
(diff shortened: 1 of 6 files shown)


docker/pg_tests/scripts/tests/archiving_ready_rename.sh
```diff
@@ -13,29 +13,29 @@ wal-g delete everything FORCE --confirm --config=${TMP_CONFIG}
...
 /usr/lib/postgresql/10/bin/initdb ${PGDATA}
 
-echo "archive_mode = on" >> /var/lib/postgresql/10/main/postgresql.conf
-echo "archive_timeout = 600" >> /var/lib/postgresql/10/main/postgresql.conf
-echo "archive_command = 'exit 1'" > /var/lib/postgresql/10/main/postgresql.auto.conf
+echo "archive_mode = on" >> ${PGDATA}/postgresql.conf
+echo "archive_timeout = 600" >> ${PGDATA}/postgresql.conf
+echo "archive_command = 'exit 1'" > ${PGDATA}/postgresql.auto.conf
 
 /usr/lib/postgresql/10/bin/pg_ctl -D ${PGDATA} -w start
```
debebantur, line 20: Is version left here on purpose? Will we change bin path later?

chipitsine: it will be part of separate discussion. I do not want to mix everthing together

chipitsine: regarding pg_ctl, initdb ... I think of adding `/usr/lib/.../bin` to PATH in Dockerfile. but it will likely raise many questions that I do not want to resolve now

PR #2088, opened 2025-10-15
Reduce walg-exporter binary size by removing wal-g dependency

jorsol:
### Database name
PostgreSQL

# Pull request description

The exporter previously depended on the wal-g module, pulling in a large set of transitive dependencies. By copying `lsn.go` locally and removing the wal-g import, the dependency tree was greatly simplified. 

This optimization cuts the final binary size by over 80% (78 MB → 13 MB) while keeping all existing functionality intact.


jorsol:
Hi @x4m, @chipitsine, @ostinru, can you please have a look ?

chipitsine:
I wonder whether we do have automatic tests for exporter. Is there a way to see whether it is working or not

chipitsine:
does it make sense to add ? or not

```
promtool check metrics
```

jorsol:
> does it make sense to add ? or not
> 
> ```
> promtool check metrics
> ```

Yes, it make sense, but adding automated tests should be done in another PR.

chipitsine (approved):

PR #2090, opened 2025-10-21
Use Valkey instead of Redis, step 1

VGoshev:
* Everywhere use Valkey instead of Redis (without breaking backward compatibility)
* `sed -e 's/redis/valkey/gi'`
* Add Valkey 7.2, 8.0, 8.1, 9.0 to Wal-G for Valkey tests

Future steps:
1. Use `valkey-go` as client to Valkey instead of `redis-go`.
2. Change `WALG_REDIS_*` parameters to `WALG_VALKEY_*`. But need to think about supporting backward compatibility (Looks like add special wrapper, which will try to get `VALKEY` version with fallback to `REDIS` version of config).
3. Rename all files from `redis` to `valkey`
4. In workflow do release for `valkey` *and* `redis` (for backward compatibility)

## WHY???

Because of Redis new (not so new, actually) license, which isn't so open now.
Valkey is new true Open  In Memory Key-Value Database. And it's much better, btw.

### Database name
~Redis~ Valkey


serprex:
redis came out alright, & we shouldn't break compat here

PR #2100, opened 2025-11-01
perf: reduce command executions and requests in WAL-G exporter

jorsol:
### Database name
PostgreSQL

## Pull request description

Optimized command usage to improve performance by reducing the number of executed commands and external requests.

Previously, the exporter ran:
  - `wal-g backup-list --detail --json` for backup information
  - `wal-g wal-show --detailed-json` for WAL segment information
  - `wal-g st ls` to check storage connectivity

Now it only runs:
  - `wal-g wal-show --detailed-json` to retrieve both WAL segment and backup info
  - `wal-g st check read` to verify storage connectivity

This reduces command executions from three to two and lowers the total number of
requests, resulting in faster metrics collection and lower resource usage.

### DIFF
--- a/cmd/pg/exporter/README.md
+++ b/cmd/pg/exporter/README.md
@@ -112,9 +112,9 @@ The exporter requires WAL-G to be properly configured and accessible. Ensure tha
 ### WAL-G Commands Used
 
 The exporter executes the following WAL-G commands:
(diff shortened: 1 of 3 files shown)


jorsol:
This is now ready!

Issue #1781, opened 2024-09-10
Snapshot backups [Design discussion]


x4m:
There are numerous requests to implement this feature, so I think time has come.
Let's start with Postgres, maybe extend it further later.

I propose to implement ```snapshot-push``` command. This command must:
1. call pg_start_backup() just like ```backup-push```
2. Run predefined command which mush push FS snapshot.
3. call pg_stop_backup()
4. Upload snapshot sentinel: usual metadata and config params.

When snapshot backup is deleted, corresponing WALs are removed from archive. Also, there call some pre-defined function.

WDYT?

x4m:
Some observations:
1. We still need to place backup_label on FS after snapshot restoration
2. IMO having one more option of backup-push would suffice

ostinru:
Nice idea... Official MySQL documentations states[1] that we can do same backups for MySQL as well (people from Percona disagrees[2]):

> 1. From a client program, execute `FLUSH TABLES WITH READ LOCK`.
> 2. From another shell, execute mount vxfs snapshot.
> 3. From the first client, execute `UNLOCK TABLES`.
> 4. Copy files from the snapshot.
> 5. Unmount the snapshot.
> Similar snapshot capabilities may be available in other file systems, such as LVM or ZFS.

Haven't tried it... Sounds like it will make database read-only during steps 1-3.
We should have deeper investigation before doing the same for MySQL.

---
[1] https://dev.mysql.com/doc/refman/8.4/en/backup-methods.html
[2] https://forums.percona.com/t/backup-mysql-using-lvm-snapshot/16778

PR #2101, opened 2025-11-01
Add snapshot backup support for PostgreSQL

x4m:
Implement snapshot backups that leverage filesystem-level or cloud disk snapshots (e.g., AWS EBS, Azure Managed Disks, GCP Persistent Disks, ZFS, LVM) for creating PostgreSQL backups while maintaining proper database consistency and point-in-time recovery capabilities.

Snapshot backups store only metadata in WAL-G storage while the actual database files remain in externally managed snapshots. This approach provides near-instantaneous backups regardless of database size and significantly reduces storage costs in WAL-G's object storage.

Key components:

1. snapshot-push command: Coordinates backup creation by calling pg_start_backup(), executing a user-defined snapshot command, calling pg_stop_backup(), and uploading metadata. The snapshot command receives environment variables (WALG_SNAPSHOT_NAME, WALG_PG_DATA, WALG_SNAPSHOT_START_LSN, WALG_SNAPSHOT_START_WAL_FILE) for proper snapshot tagging and identification.

2. snapshot-fetch command: Prepares restored snapshots for PostgreSQL recovery by creating backup_label and tablespace_map files from stored metadata. Supports automatic recovery configuration for both PostgreSQL 12+ (recovery.signal) and earlier versions (recovery.conf), with optional point-in-time recovery target specification.

3. Automatic WAL protection: Critical safety feature that prevents deletion of WAL segments required by snapshot backups. During any delete operation, WAL-G identifies all snapshot backups and protects their required WAL range (start LSN to finish LSN) from deletion, ensuring snapshot backups remain recoverable even with aggressive retention policies.

4. Exact backup_label preservation: Stores the exact content returned by pg_stop_backup() for backup_label and tablespace_map files rather than reconstructing them. This ensures compatibility across all PostgreSQL versions and handles any future format changes automatically, as PostgreSQL-generated files are guaranteed to be readable by PostgreSQL.

Implementation stores snapshot metadata in BackupSentinelDto with new BackupLabel and TablespaceMap fields. Snapshot backups are identified by FilesMetadataDisabled=true, CompressedSize=0, and presence of BackupLabel content. Delete operations check IsSnapshotBackup() and protect required WAL segments through GetPermanentBackupsAndWals() modifications.

Configuration uses WALG_SNAPSHOT_COMMAND (required) for snapshot creation and WALG_SNAPSHOT_DELETE_COMMAND (optional) for cleanup. Commands execute in shell context with environment variables for maximum flexibility across different infrastructure providers.

Testing includes comprehensive snapshot_test.sh with 10 test cases covering backup creation, restoration, PITR, deletion, retention policies, and WAL protection verification. Tests use cp -al (hardlinks) to emulate filesystem snapshots without requiring cloud infrastructure.

Documentation in docs/PostgreSQL_Snapshot.md provides complete usage examples for major cloud providers (AWS, Azure, GCP) and on-premises solutions (ZFS, LVM), along with technical implementation details, security considerations, and best practices.

Snapshot backups integrate seamlessly with existing WAL-G features including backup-list, delete commands, permanent backup flag, encryption, compression (for WAL files), and multiple storage backends.

Author: Cursor, Sonnet 4.5, some whacking by me
Discussion: https://github.com/wal-g/wal-g/issues/1781

### Database name

# Pull request description

// problem is ...

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/internal/databases/postgres/snapshot_push_handler.go
+++ b/internal/databases/postgres/snapshot_push_handler.go
@@ -32,6 +32,7 @@ type SnapshotBackupHandler struct {
 	QueryRunner      *PgQueryRunner
 	PgInfo           BackupPgInfo
 	startWalFileName string
+	isStandby        bool
 	// Exact content from pg_stop_backup() - don't reconstruct, use what Postgres gives us
 	backupLabel   string
 	tablespaceMap string
@@ -143,8 +144,11 @@ func (sbh *SnapshotBackupHandler) startBackup() error {
 		return errors.Wrap(err, "pg_start_backup() failed")
 	}
 
+	// Note: pg_backup_start() can be called on standby servers
+	// On standby, walFileName will be empty and we'll use the LSN instead
+	sbh.isStandby = inRecovery
 	if inRecovery {
-		return errors.New("Cannot perform snapshot backup on a standby server")
+		tracelog.InfoLogger.Println("Snapshot backup is being performed on a standby server")
 	}
 
 	// Parse start LSN
--- a/cmd/pg/snapshot_push.go
+++ b/cmd/pg/snapshot_push.go
@@ -22,26 +22,28 @@ const (
 var (
 	// snapshotPushCmd represents the snapshot-push command
 	snapshotPushCmd = &cobra.Command{
-		Use:   "snapshot-push db_directory",
+		Use:   "snapshot-push [db_directory]",
 		Short: snapshotPushShortDescription,
 		Long: `Creates a snapshot backup using filesystem or cloud disk snapshots.
 
 This command:
-1. Calls pg_start_backup() to ensure database consistency
-2. Executes a user-defined snapshot command (WALG_SNAPSHOT_COMMAND)
-3. Calls pg_stop_backup()
-4. Uploads backup metadata to storage
+1. Connects to PostgreSQL and gets the data directory (if not provided as argument)
+2. Calls pg_start_backup() to ensure database consistency
+3. Executes a user-defined snapshot command (WALG_SNAPSHOT_COMMAND)
+4. Calls pg_stop_backup()
+5. Uploads backup metadata to storage
 
 The snapshot command receives the following environment variables:
   - WALG_SNAPSHOT_NAME: The backup name
   - WALG_PG_DATA: The PostgreSQL data directory path
   - WALG_SNAPSHOT_START_LSN: The backup start LSN
+  - WALG_SNAPSHOT_START_WAL_FILE: The WAL file name at backup start
 
 Example snapshot command for AWS EBS:
   aws ec2 create-snapshot --volume-id vol-xxxxx --description "$WALG_SNAPSHOT_NAME"
 
 WAL archiving must be properly configured for point-in-time recovery.`,
-		Args: cobra.ExactArgs(1),
+		Args: cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
 			internal.ConfigureLimiters()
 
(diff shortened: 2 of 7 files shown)


internal/databases/postgres/snapshot_push_handler.go
```diff
@@ -0,0 +1,330 @@
...
+	var lsnString string
+	var inRecovery bool
+	err = sbh.QueryRunner.Connection.QueryRow(context.TODO(), startBackupQuery, backupName).Scan(
+		&walFileName, &lsnString, &inRecovery)
+	if err != nil {
+		return errors.Wrap(err, "pg_start_backup() failed")
+	}
+
+	if inRecovery {
+		return errors.New("Cannot perform snapshot backup on a standby server")
```
boosterKRD: Possible issue: when pg_start_backup() runs successfully but returns inRecovery = true, the function immediately returns an error without calling pg_stop_backup().

x4m: thanks for reviewing! Well, this code was written by Cursor and it is...well... I see no reason to disallow snapshot backup on standby.

boosterKRD: > thanks for reviewing! Well, this code was written by Cursor and it is...well... I see no reason to disallow snapshot backup on standby.

Hi !
I thought about it for a while — why standby is forbidden — but in the end, I accepted the “religion of the code.” Probably for the best: there’s no real load on the primary, and using a replica doesn’t add much value.

If you run pg_stop_backup() on a replica, it finishes immediately without waiting for the WAL segments (the range mentioned in backup_label) to be archived — especially since archive_mode is usually disabled on replicas.
So you get a “successful” backup, but at restore time, some required WAL segments might be missing.

On the primary, pg_stop_backup() always waits until all WALs are archived, ensuring consistency.

From my view — if backups on replicas are ever allowed, docs must clearly state that users are fully responsible for controlling WAL archiving.

internal/uploader.go
```diff
@@ -215,7 +215,7 @@ func (uploader *RegularUploader) UploadJSON(ctx context.Context, path string, da
 
 	errorGroup, _ := errgroup.WithContext(ctx)
 	errorGroup.Go(func() error {
-		err := json2.MarshalWrite(writer, data)
+		err := json.NewEncoder(writer).Encode(data)
```
ostinru, line 218: Is there any reason to rollback https://github.com/wal-g/wal-g/pull/2056 ?
If it breaks compilation - upgrade to 1.25 and add `GOEXPERIMENT=jsonv2` to your env (this part is not obvious).

internal/databases/postgres/snapshot_push_handler.go
```diff
@@ -0,0 +1,332 @@
...
+	// Connect to postgres
+	tracelog.DebugLogger.Println("Connecting to Postgres for snapshot backup.")
+	conn, err := Connect()
+	if err != nil {
+		return err
+	}
+
+	sbh.QueryRunner, err = NewPgQueryRunner(conn)
+	if err != nil {
+		return errors.Wrap(err, "failed to build query runner")
```
ostinru, line 126: NIT: it seems that `github/pkg/errors` is public archive and we can use `fmt.Errorf()` instead.

internal/databases/postgres/snapshot_push_handler.go
```diff
@@ -0,0 +1,332 @@
...
+	startBackupQuery, err := sbh.QueryRunner.BuildStartBackup()
+	if err != nil {
+		return errors.Wrap(err, "Building start backup query failed")
+	}
+
+	var walFileName string
+	var lsnString string
+	var inRecovery bool
+	err = sbh.QueryRunner.Connection.QueryRow(context.TODO(), startBackupQuery, backupLabel).Scan(
+		&walFileName, &lsnString, &inRecovery)
```
ostinru, line 142: Why not use `sbh.QueryRunner.StartBackup()`?

internal/databases/postgres/snapshot_push_handler.go
x4m: We already support backups on standby since 2017. So there's no way back in forbidding them somewhere :)

internal/uploader.go
x4m: yup, this is bogus change, I'll keep json2

internal/databases/postgres/snapshot_push_handler.go
x4m: I think at some point we should make this consistent across codebase...

internal/databases/postgres/snapshot_push_handler.go
x4m: Good idea!

PR #2104, opened 2025-11-05
tests: unleash the power of "prepare_config", pg

chipitsine:
test scripts refactoring, no functional change

### DIFF
--- a/docker/pg_tests/scripts/tests/archiving_ready_rename.sh
+++ b/docker/pg_tests/scripts/tests/archiving_ready_rename.sh
@@ -1,13 +1,8 @@
 #!/bin/sh
 set -e -x
-CONFIG_FILE="/tmp/configs/archiving_ready_rename.json"
-COMMON_CONFIG="/tmp/configs/common_config.json"
-TMP_CONFIG="/tmp/configs/tmp_config.json"
 
-cat ${CONFIG_FILE} > ${TMP_CONFIG}
-echo "," >> ${TMP_CONFIG}
-cat ${COMMON_CONFIG} >> ${TMP_CONFIG}
-/tmp/scripts/wrap_config_file.sh ${TMP_CONFIG}
+. /tmp/tests/test_functions/prepare_config.sh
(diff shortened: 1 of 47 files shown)


docker/pg_tests/scripts/tests/archiving_ready_rename.sh
```diff
@@ -1,13 +1,8 @@
...
-COMMON_CONFIG="/tmp/configs/common_config.json"
-TMP_CONFIG="/tmp/configs/tmp_config.json"
-
-cat ${CONFIG_FILE} > ${TMP_CONFIG}
-echo "," >> ${TMP_CONFIG}
-cat ${COMMON_CONFIG} >> ${TMP_CONFIG}
-/tmp/scripts/wrap_config_file.sh ${TMP_CONFIG}
+
+. /tmp/tests/test_functions/prepare_config.sh
+prepare_config "/tmp/configs/archiving_ready_rename.json"
```
ostinru, line 5: I was thinking about moving to some `jq` magic - explicitly merge several JSONs.
In this scenario we will achieve 2 goals:
* we use standard tooling
* we have valid JSONs in `docker/pg_tests/scripts/configs`

Unfortunately, quick google-ing about deep JSON merge with `jq` haven't show best solution...

chipitsine: once this is done by single function, we can change that function accordingly

ostinru: `jq` is little bit more explicit: it is clear that you joins several JSONs... this script hides the fact that it merges `commong_config.json`.
And I have no idea how `TMP_CONFIG` works below... probably I am bad at bash.

chipitsine: that function was introduced 2 years ago:

https://github.com/wal-g/wal-g/commit/75d7334af04176722575c00785bb6ff30a5ad91f

why not to use existing function.
is a lot of code duplication really better ?

PR #1956, opened 2025-04-05
releases: add ubuntu22 for aarch64

chipitsine:

PR #2115, opened 2025-11-19
releases: add ubuntu24 for aarch64

sixcorners:
### Database name

PostgreSQL

24.04 arm builds

There is some discussion here: https://github.com/wal-g/wal-g/pull/1956
I wasn't able to get any of the arm things building 24.04 or not.
https://github.com/sixcorners/walg/actions/runs/19503691114/job/55824875713

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ { distro: ubuntu20.04, label: ubuntu-20.04 }, { distro: ubuntu22.04, label: ubuntu-22.04 } ]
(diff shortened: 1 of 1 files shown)


chipitsine:
@sixcorners , can you try the following ?

here are defined variables for "run-on-arch"

https://github.com/sixcorners/walg/blob/master/.github/workflows/release.yml#L112-L115

I think it does not inherit GOEXPERIMENT from https://github.com/sixcorners/walg/blob/master/.github/workflows/release.yml#L10

can you try adding GOEXPERIMENT ?

(obsiously, it would be more elegant to switch to native arm64 runners)

chipitsine:
I've made similar PR  https://github.com/wal-g/wal-g/pull/2140

chipitsine:
aarch64 releases are implemented in https://github.com/wal-g/wal-g/pull/2140

I think we can close this PR

PR #2123, opened 2025-11-29
feat: allow configuration of S3_ENABLE_VERSIONING

jorsol:
### Database name
PostgreSQL

# Pull request description
By default, WAL-G checks whether versioning is enabled on the bucket. When the user already knows the bucket’s versioning state, this setting allows them to specify `enabled` or `disabled`, avoiding an unnecessary round trip to the server.


ostinru (approved):

ostinru:
LGTM. 
Will leave for a couple days in case if someone else would like to review this PR as well.

Issue #1960, opened 2025-04-14
[BUG] s3 versioning breaks wal-g delete


dbarrosop:
### Database name

PostgresSQL

### WAL-G Version

v3.0.6

### Describe your problem

When using an s3 bucket with versioning enabled the following issue arises:

```
$ wal-g delete retain 7
INFO: 2025/04/14 10:51:47.625436 Backup to delete will be searched in storages: [default]
INFO: 2025/04/14 10:51:47.625480 retrieving permanent objects
INFO: 2025/04/14 10:51:47.740142 Backup base_0000000100000029000000A1 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:47.749351 Backup base_000000010000002A000000C2 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:47.763387 Backup base_000000010000002B000000E3 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:47.772679 Backup base_000000010000002D00000004 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:48.117691 Backup base_0000000100000029000000A1 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:48.127546 Backup base_000000010000002A000000C2 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:48.136454 Backup base_000000010000002B000000E3 lacks metadata to check if it's permanent, ignoring...
INFO: 2025/04/14 10:51:48.144452 Backup base_000000010000002D00000004 lacks metadata to check if it's permanent, ignoring...
ERROR: 2025/04/14 10:51:48.180620 object 'base_0000000100000029000000A1_backup_stop_sentinel.json' not found in storage
```

All of those files were basically deleted on a previous run but it looks like wal-g is still processing them.

If I remove the permission `s3:GetBucketVersioning` (or disable versioning) then the issue goes away but it makes our compliance tool sad. Is this intended? Is this a bug? Does wal-g not support buckets with versioning enabled? Am I missing some setting?

### Steps to reproduce

1. use s3 as storage
2. enable versioning on the bucket
3. run `wal-g delete retain ...` multiple times

### Config and WAL-G logs

logs:

```
$ wal-g delete retain 7
INFO: 2025/04/14 10:51:47.625436 Backup to delete will be searched in storages: [default]
[...]

PR #2132, opened 2025-12-06
fix(s3): exclude deleted objects from ListFolder when versioning is enabled

oleksiytsyban:
When S3 bucket versioning is enabled, ListFolder() incorrectly returned objects that have been deleted (objects where the latest version is a delete marker). This caused:
- backup-list to show deleted backups
- backup-list --detail to fail with 'object not found in storage'
- st ls to show deleted objects

Modified listVersions() to filter out objects where the LATEST version is a delete marker. Uses a two-pass approach to handle pagination correctly, since delete markers and versions might appear on different pages of the ListObjectVersions API response.

Added unit tests to verify the fix.

### Database name
Tested with PostgreSQL

# Pull request description
## Problem

When S3 bucket versioning is enabled, `ListFolder()` incorrectly returns objects that have been deleted (objects where the latest version is a delete marker). This causes:

- `backup-list` to show deleted backups
- `backup-list --detail` to fail with "object not found in storage" when trying to read metadata of deleted backups
- `st ls` to show deleted objects

## Root Cause

The `listVersions()` function in `pkg/storages/s3/folder.go` was including delete markers as regular objects in the returned list.

## Solution

Modified `listVersions()` to filter out objects where the LATEST version is a delete marker. The fix uses a two-pass approach to handle pagination correctly, since delete markers and versions might appear on different pages of the `ListObjectVersions` API response.

Added `st ls --all-versions` flag. 
The `--all-versions` output will show:
- Regular objects with LATEST label
- Old versions with version ID
- Delete markers with DELETE or LATEST DELETE label

## Testing

Added 4 unit tests to verify:
- Objects with LATEST delete marker are excluded
- Non-deleted objects include all versions  
- ALL versions of deleted objects are excluded
- Objects recreated after deletion are included

### DIFF
--- a/pkg/storages/s3/folder.go
+++ b/pkg/storages/s3/folder.go
@@ -187,6 +187,93 @@ type versionInfo struct {
 	isLatest     bool
 }
 
+// addListedSubfolders converts S3 CommonPrefixes to Folder handles.
+// These handles are used by callers (e.g. `st ls` non-recursive output, or recursive listing via subfolders).
+func (folder *Folder) addListedSubfolders(subFolders *[]storage.Folder, commonPrefixes []*s3.CommonPrefix) {
+	for _, p := range commonPrefixes {
+		subFolder := NewFolder(folder.s3API, folder.uploader, *p.Prefix, folder.config)
+		// Propagate the showAllVersions setting to subfolders created during listing
+		subFolder.showAllVersions = folder.showAllVersions
+		*subFolders = append(*subFolders, subFolder)
+	}
+}
+
+// collectDeleteMarkers gathers delete marker entries and records which keys are "deleted"
+// (i.e. their LATEST is a delete marker). `deletedKeys` drives filtering in buildObjectsFromVersions.
+func (folder *Folder) collectDeleteMarkers(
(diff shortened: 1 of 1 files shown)


debebantur:
Thank you for your pr! It is a great addition. However, when calling wal-g st ls, it is better to show everything we have in bucket. Can you, please, add some flag to return all versions?

oleksiytsyban:
@debebantur Do you want to show deleted files and folders when calling  wal-g st ls? Why?

debebantur:
> @debebantur Do you want to show deleted files and folders when calling  wal-g st ls? Why?

There are 2 main reasons

1) If you accidentally st rm some object in versioned bucket you could undo it via wal-g if you see all versions. Main purpose of wal-g is not working with s3, however, i believe that basic functions will be useful

2) One of the benefits of versioning buckets is safety. In future I want to make an option to restore backup from oldest accessible versions. ( In case someone deleted or encrypted original backup )

oleksiytsyban:
@debebantur Added flag, as requested.

oleksiytsyban:
Also fixes https://github.com/wal-g/wal-g/issues/1960

PR #2134, opened 2025-12-08
Ensure wal-g-binlog-server handles reconnections correctly

zkramorev:
## Problem
If a client disconnects and then reconnects:
A new connection is accepted, but it reuses the old Handler.
The sync.Once.Do() call, which is responsible for starting the binlog streaming process (syncBinlogFiles), does not execute again because it has already run for this Handler instance.
As a result, wal-g-binlog-server does not send any more binlog data to the reconnected client.
The server gets stuck in waitReplicationIsDone, waiting for the replica to apply GTIDs that will never be sent, while the replica is stuck waiting for data that will never arrive.

---
wal-g-binlog-server.log on the failed host:
```bash
DEBUG: 2025/12/08 10:36:22.927071 Expected GTID set: 1a9***a:1-311; MySQL GTID set: 1a9***a:1-269,e***a:1-1105
DEBUG: 2025/12/08 10:36:24.928453 Expected GTID set: 1a9***a:1-311; MySQL GTID set: 1a9***a:1-269,e***a:1-1105
... (repeats indefinitely)
```

## Solution
- Each new connection now gets a new Handler instance with context.Context. If a connection drops, the context cancels ongoing downloads/streams safely.
- waitReplicationIsDoneSafe now gracefully waits for the replica to catch up only after the entire S3 binlog queue is fully processed and sent
## Dependencies Upgrade
This PR upgrades `github.com/go-mysql-org/go-mysql` to the commit
`498f8104b8ffc4905a7c3172a5a186668d371470`, where I made [this change](https://github.com/go-mysql-org/go-mysql/pull/1115) to further refine the binlog-server in this PR.
To adapt the new versions of the libraries, I have corrected only `MariadbGTIDSet` in `wal-g/internal/databases/mysql/mysql_binlog.go` and deprecated `NewConn()` in `wal-g/internal/databases/mysql/binlog_server_handler.go`

### DIFF
--- a/docker/mysql_tests/scripts/base_tests/pitr_binlog_server_reconnection_test.sh
+++ b/docker/mysql_tests/scripts/base_tests/pitr_binlog_server_reconnection_test.sh
@@ -138,14 +138,14 @@ if [ $WAIT_COUNT -eq $MAX_WAIT ]; then
     exit 1
 fi
 
-sleep 5
+sleep 3
 
 log "Starting proxy with reconnections..."
 python3 "$PROXY_SCRIPT" $PROXY_PORT "127.0.0.1" $BINLOG_SERVER_PORT $PLANNED_DISCONNECTS > $PROXY_LOG 2>&1 & proxy_pid=$!
 log "Started proxy with PID: $proxy_pid"
 
 log "Waiting for proxy to start..."
-sleep 15
+sleep 7
 if ! kill -0 $proxy_pid 2>/dev/null; then
     log "ERROR: Proxy process died"
     cat $PROXY_LOG
@@ -253,4 +253,3 @@ fi
 
 
 log "Test completed successfully!"
-sleep 11
--- a/internal/databases/mysql/binlog_server_handler.go
+++ b/internal/databases/mysql/binlog_server_handler.go
@@ -115,22 +115,21 @@ func addRotateEvent(s *replication.BinlogStreamer, pos mysql.Position) error {
 	return s.AddEventToStreamer(&rotateBinlogEvent)
 }
 
+func (h *Handler) buildGTIDStr() string {
+	parts := make([]string, 0, len(h.lastSentGTIDs))
+	for u, gno := range h.lastSentGTIDs {
+		parts = append(parts, fmt.Sprintf("%s:1-%d", u, gno))
+	}
+	return strings.Join(parts, ",")
+}
+
 func (h *Handler) waitReplicationIsDoneSafe() {
 	if len(h.lastSentGTIDs) == 0 {
 		tracelog.InfoLogger.Println("S3 objects finished. No GTIDs were sent. Shutting down immediately.")
 		os.Exit(0)
 	}
 
-	var sb strings.Builder
-	first := true
-	for u, gno := range h.lastSentGTIDs {
-		if !first {
-			sb.WriteString(",")
-		}
-		sb.WriteString(fmt.Sprintf("%s:1-%d", u, gno))
-		first = false
-	}
-	targetGTIDStr := sb.String()
+	targetGTIDStr := h.buildGTIDStr()
 
 	tracelog.InfoLogger.Printf("All S3 binlogs processed. Waiting for replica to catch up to GTID: %s", targetGTIDStr)
 
--- a/pkg/storages/storage/object_provider.go
+++ b/pkg/storages/storage/object_provider.go
@@ -107,6 +107,6 @@ func (p *ObjectProvider) Close() {
 	if p.closed {
 		return
 	}
-	p.closed = true
 	close(p.ch)
+	p.closed = true
 }


internal/databases/mysql/binlog_server_handler.go
```diff
@@ -35,27 +37,25 @@ var (
...
-	for !ok {
-		ok = s.AddErrorToStreamer(err)
-	}
+	producerRunning bool
+
+	replicationDone bool
+	doneMutex       sync.Mutex
 }
 
-// see: https://dev.mysql.com/doc/dev/mysql-server/latest/classbinary__log_1_1Rotate__event.html
```
ostinru, line 53: Let's replace it with link, that shouldn't change:
https://github.com/percona/percona-server/blob/8.0/libbinlogevents/include/control_events.h#L53-L108

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -108,9 +116,28 @@ func waitReplicationIsDone() error {
 	if err != nil {
 		return err
 	}
+	defer func(db *sql.DB) {
+		_ = db.Close()
+	}(db)
```
ostinru:
```
defer utility.LoggedClose(db, "some text here")
```

pkg/storages/storage/object_provider.go
```diff
@@ -10,78 +14,99 @@ var (
...
-	case p.ech <- ErrProviderClosed:
-	default:
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.closed {
+		return
 	}
-	close(p.ech)
+	p.closed = true
+	close(p.ch)
```
teem0n, line 110: Shouldn't we call close() before p.closed = true ?
In case of error we will return in  `if p.closed {`, and won't close channel

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -99,166 +117,217 @@ func addRotateEvent(s *replication.BinlogStreamer, pos mysql.Position) error {
...
+	p.SetFlavor(mysql.MySQLFlavor)
+	// check checksum on our side - we should exit with error here rather than stuck waiting for MySQL apply all binlogs till `lastSentGTID`.
+	p.SetVerifyChecksum(true)
+
+	err := p.ParseFile(binlogPath, 4, func(e *replication.BinlogEvent) error {
+		if e.Header.EventType == replication.GTID_EVENT {
+			gtidEvent := &replication.GTIDEvent{}
+			if err := gtidEvent.Decode(e.RawData[replication.EventHeaderSize:]); err == nil {
+				u, _ := uuid.FromBytes(gtidEvent.SID)
+				result = u.String() + ":1-" + strconv.Itoa(int(gtidEvent.GNO))
```
teem0n: We can use stringbuilder here

docker/mysql_tests/scripts/base_tests/pitr_binlog_server_reconnection_test.sh
```diff
@@ -73,183 +82,175 @@ safe_kill_process() {
...
-done
+log "Starting proxy with reconnections..."
+python3 "$PROXY_SCRIPT" $PROXY_PORT "127.0.0.1" $BINLOG_SERVER_PORT $PLANNED_DISCONNECTS > $PROXY_LOG 2>&1 & proxy_pid=$!
+log "Started proxy with PID: $proxy_pid"
 
-if [ $WAIT_COUNT -eq 15 ]; then
-    echo "ERROR: Proxy failed to start within 15 seconds"
-    echo "=== Proxy log ==="
+log "Waiting for proxy to start..."
+sleep 15
```
ostinru: it is bad idea to use `sleep` in tests - tests became SLOW. When tests are slow - developers start procrastinate: drink coffee, play kicker, etc.

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -50,12 +64,16 @@ func handleEventError(err error, s *replication.BinlogStreamer) {
...
-	tracelog.ErrorLogger.FatalOnError(err)
-	ServerIDNum, err := strconv.Atoi(serverID)
-	tracelog.ErrorLogger.FatalOnError(err)
+	if err != nil {
+		return err
+	}
+	serverIDNum, err := strconv.Atoi(serverID)
+	if err != nil {
+		return err
+	}
```
ostinru: these errors unrecoverable. I think it is fine to leave it as `tracelog.ErrorLogger.FatalOnError`

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -99,166 +117,217 @@ func addRotateEvent(s *replication.BinlogStreamer, pos mysql.Position) error {
...
+	maxGTID := extractMaxGTID(binlogPath)
+	if maxGTID == "" {
+		return false
+	}
+	fileSet, parseErr := mysql.ParseGTIDSet("mysql", maxGTID)
+	if parseErr != nil {
+		return false
+	}
+	if clientGTIDSet.Contain(fileSet) {
```
ostinru: You are parsing first GTID in binlog and skipping whole file if mysqld applied first transaction from file. There is a risk of loosing transactions here.

docker/mysql_tests/scripts/base_tests/pitr_binlog_server_reconnection_test.sh
```diff
@@ -73,183 +82,175 @@ safe_kill_process() {
...
+    log "- Expected $EXPECTED_ROWS rows, got $FINAL_ROW_COUNT"
+    log "- Proxy reconnects: $PROXY_RECONNECTS"
     exit 1
 fi
 
-echo "Test completed successfully!"
\ No newline at end of file
+
+log "Test completed successfully!"
+sleep 11
```
ostinru: Why do you need `sleep` here?

zkramorev: No need, yes. I will remove all unnecessary sleep

internal/databases/mysql/binlog_server_handler.go
zkramorev: returned the old mechanics of sending all transactions

pkg/storages/storage/object_provider.go
zkramorev: I did it based on the logic that in Go, the close() function doesn't return an error. The p.closed flag is essential here to prevent a panic: close of closed channel in case Close() is called multiple times. 
Do I think I should leave the current option or did I miss something?

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -99,166 +115,182 @@ func addRotateEvent(s *replication.BinlogStreamer, pos mysql.Position) error {
...
-	replicaSource, err := conf.GetRequiredSetting(conf.MysqlBinlogServerReplicaSource)
+func (h *Handler) waitReplicationIsDoneSafe() {
+	if len(h.lastSentGTIDs) == 0 {
+		tracelog.InfoLogger.Println("S3 objects finished. No GTIDs were sent. Shutting down immediately.")
+		os.Exit(0)
+	}
+
+	var sb strings.Builder
+	first := true
+	for u, gno := range h.lastSentGTIDs {
```
teem0n: Let's put this into separate method "BuildGtidStr" etc

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -311,46 +344,83 @@ func HandleBinlogServer(since string, until string) {
...
 	tracelog.ErrorLogger.FatalOnError(err)
 	serverPort, err := conf.GetRequiredSetting(conf.MysqlBinlogServerPort)
 	tracelog.ErrorLogger.FatalOnError(err)
+
 	l, err := net.Listen("tcp", serverAddress+":"+serverPort)
 	tracelog.ErrorLogger.FatalOnError(err)
 	tracelog.InfoLogger.Printf("Listening on %s, wait connection", l.Addr())
 
-	globalHandler := &Handler{}
+	srv := server.NewServer("5.7.42", mysql.DEFAULT_COLLATION_ID, mysql.AUTH_NATIVE_PASSWORD, nil, nil)
```
teem0n, line 358: Why 5.7.42?)

zkramorev: I think 5.7.x is for maximum compatibility when initially establishing a connection. it also helps us to stick to mysql_native_password and avoid problems with caching_sha2_password authentication, which may occur if we present the server as 8.0+. And (.42) doesn't really matter

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -99,166 +115,182 @@ func addRotateEvent(s *replication.BinlogStreamer, pos mysql.Position) error {
...
+		if h.ctx.Err() != nil {
+			tracelog.WarningLogger.Println("Client disconnected while waiting for completion. Handler shutting down, awaiting reconnect...")
+			return
 		}
 
-		lastSentGTIDSet, err := mysql.ParseGTIDSet("mysql", lastSentGTID)
-		if err != nil {
-			return err
+		var executedStr string
+		err := db.QueryRowContext(h.ctx, "SELECT @@global.gtid_executed").Scan(&executedStr)
```
teem0n, line 155: What if we failed to connect at line 143?
Should not we reconnect first?

Issue #631, opened 2020-04-16
Documentation for tablespace support?


thedevelopnik:
The release for 2.14 states that there is a preview of tablespace support. This is a critical feature for us, but I can't find any documentation on it. Does it work using a json file similar to wal-e? I'd love to preview it, give feedback and even contribute but can't find any information other than that one-liner in the release notes. Thanks!

thedevelopnik:
Happy to contribute to documentation as well.

x4m:
@thedevelopnik Hi! Well, the idea was to implement WAL-E interface.
@Van10000 can you please comment on docs and usage of tablespaces?

kamholz:
I'd love to see documentation for this too. I tried a `backup-fetch` on a backup with tablespaces, and ran into two issues:

1. The JSON file I passed to `--restore-spec` wouldn't work without the `base_prefix` key. This isn't documented as required for WAL-E, although it seems to be supported there. I looked through the WAL-G code to understand what it's for.

2. After adding the `base_prefix`, it was ignored by `backup-fetch`. I fetched the backup into a different directory than it was backed up from, but `backup-fetch` still tried to write to the original directory (which, thankfully, it didn't have permission to do). I looked through the code for how it takes the supplied `base_prefix` value and saw that it puts it into the TablespaceSpec structure, but apparently that code isn't working properly.

PR #2137, opened 2025-12-14
Support WAL-E tablespace spec format for better compatibility

x4m:
Hopefully partially address #631.
Accept both "Spec" (WAL-G) and "spec" (WAL-E) JSON keys when reading
tablespace specifications during backup-fetch. This allows WAL-G to
restore backups created by WAL-E without requiring users to modify
the spec file.

Output continues to use uppercase "Spec" to maintain backward
compatibility with existing user scripts that parse WAL-G backups.

Add integration test for --restore-spec with tablespace relocation.


x4m:
This PR is superseded by #2139

PR #2139, opened 2025-12-15
Fix tablespace restore-spec

NikkoFox:
### Database name
PostgreSQL

# Pull request description

During `deltaFetchRecursionOld(New)`, we attempt to override the backup's `TablespaceSpec` with a user-provided specification `--restore-spec` (tablespaceSpec). However, the previous implementation was only updating the local pointer `sentinelDto.TablespaceSpec`, while the actual object referenced by `backup.SentinelDto` remained unchanged.

As debug logs confirmed, backup.SentinelDto.TablespaceSpec retained the original (old) basePrefix, causing subsequent functions like backup.unwrapToEmptyDirectory (which uses *backup.SentinelDto) to use the wrong path.

Use backup-fetch with custom base_prefix in `--restore-spec`

```
// Correct spec (user provided) --restore-spec json
INFO: ... &{basePrefix:/pgdata/test-1 ...} 
// Local variable updated correctly; tablespaceSpec = chooseTablespaceSpecification(sentinelDto.TablespaceSpec, tablespaceSpec)
INFO: ... &{basePrefix:/pgdata/test-1 ...} 
// But the original backup struct still holds the old value; err := checkDBDirectoryForUnwrap(dbDataDirectory, *backup.SentinelDto, *backup.FilesMetadataDto)
INFO: ... &{basePrefix:/pgdata/test-m1019 ...} 
```

Maybe this will fix the problem from https://github.com/wal-g/wal-g/issues/631#issuecomment-620208001

### DIFF
--- a/docker/pg_tests/scripts/tests/walg_tablespace_restore_spec_test.sh
+++ b/docker/pg_tests/scripts/tests/walg_tablespace_restore_spec_test.sh
@@ -138,9 +138,13 @@ wal-g backup-fetch --restore-spec /tmp/modified_restore_spec.json ${PGDATA} LATE
 echo "=== Verifying tablespace symlinks ==="
 ls -la ${PGDATA}/pg_tblspc/
 
-# TODO: i don't know what to do with tablespace_map in walg
-cat ${PGDATA}/tablespace_map
-rm ${PGDATA}/tablespace_map
+# Verify tablespace_map is not existing
+if [ ! -f ${PGDATA}/tablespace_map ]; then
+    echo "tablespace_map does not exist"
+else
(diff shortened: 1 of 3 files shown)


NikkoFox:
If postgres<15 and tablespace_map exists, postgres will recreate the symlinks and break the restore process. 
I don't know what to do safely about this.

```
# walg_tablespace_restore_spec_test.sh
# TODO: i don't know what to do with tablespace_map in walg
cat ${PGDATA}/tablespace_map
rm ${PGDATA}/tablespace_map
```

x4m (approved):

PR #2148, opened 2026-01-06
run orioledb tests sequentially to save time

chipitsine:
current orioledb test includes orioledb build which takes about 8 minutes.

so far, before: 8 min 20 sec three times = 25 minutes
        after:  10 min

this commit introduces breaking change, no previous behaviour is kept, i.e. passing TEST=test_name is not supported anymore for the sake of simplicity

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ pg_integration_test: clean_compose
 
 orioledb_integration_test: install_and_build_pg clean_compose load_docker_common
 	docker compose build orioledb
-	docker compose up --exit-code-from $(TEST) $(TEST)
+	docker compose up --exit-code-from orioledb orioledb
 	make clean_compose
 
 .PHONY: clean_compose


Makefile
```diff
@@ -106,7 +106,7 @@ pg_integration_test: clean_compose
 
 orioledb_integration_test: install_and_build_pg clean_compose load_docker_common
 	docker compose build orioledb
-	docker compose up --exit-code-from $(TEST) $(TEST)
+	for f in docker/orioledb/scripts/tests/*; do T=$$(basename "$${f}" .$${f##*.}); docker compose up --exit-code-from orioledb_$${T} orioledb_$${T}; done
```
ostinru: It is cryptic.

May be move to `run_integration_tests.sh` as it done for many databases?

chipitsine: I checked current approach. In all run_integration_tests.sh scripts are executed in one containter.
while for orioledb I just keep current container layout and launch containers one by one.

maybe just several "docker compose" invocation instead of cryptic calculation in "for" loop ?

chipitsine: I'll try to run scripts in the same container

chipitsine:
@homper can you have a look please ?

Issue #767, opened 2020-09-23
Add disable compression.


AlexGluck:
Can anyone add `WALG_COMPRESSION_METHOD=none`?

PR #2156, opened 2026-01-19
Add 'none' compression method

den-ibr:
### Database name
Any database

# Pull request description
Add 'none' compression method ([Issue #767](https://github.com/wal-g/wal-g/issues/767))


internal/compression/none_enabled.go
debebantur, line 1: I believe, you can just add this compression to compression_default.go.

PR #2161, opened 2026-01-21
Feature/prometheus exporter

santiagolertora:
### Database name
MySQL/MariaDB

# Pull request description

This PR addresses critical code review feedback for the Prometheus exporter implementation:

**Problems fixed:**
1. **High cardinality metrics** - Using `backup_name` as a label was creating a new time series for each backup, leading to potential Prometheus memory issues
2. **Missing error handling** - GTID parsing errors were silently ignored, which could lead to data inconsistency
3. **Fragile binlog parsing** - Text parsing lacked validation and error logging
4. **Non-functional Dockerfile** - WAL-G installation was a placeholder comment
5. **Weak test script** - No health check retry logic, could fail intermittently
6. **Documentation issues** - Incorrect relative paths

**Changes made:**
- Fixed metric cardinality by using "latest" constant instead of backup_name
- Added proper error handling for GTID parsing operations in binlog_push_handler.go
- Improved binlog parsing with validation and error logging
- Fixed WAL-G installation in Dockerfile with proper release download
- Added health check retry loop in test.sh (10 retries with 1s interval)
- Fixed documentation path to use absolute path from repo root

N/A - These are code quality improvements based on code review feedback

N/A - No runtime issues, these are preventive fixes for production readiness

**Build verification:**
cd cmd/mysql/exporter
go build .
# Build successful, no errors


chipitsine:
are these folders included in purpose ? first one looks related to prometheus, not sure about yellow folders

image

cmd/mysql/exporter/.gitignore
```diff
@@ -0,0 +1,40 @@
+# Binaries
+walg-mysql-exporter
+walg-mysql-exporter-linux
+*.exe
```
chipitsine, line 4: do you use wal-g on windows ?
if so, shall we add it to https://github.com/wal-g/wal-g/blob/master/.github/workflows/windows-native.yml

chipitsine:
I think currently we do not run 

```
cd cmd/mysql/exporter
go build .
```

in ci.


shall we maybe create dedicated build job which will build exporters

cmd/mysql/exporter/.gitignore
santiagolertora: Good point. Added a dedicated CI workflow (.github/workflows/exporter-build.yml) that runs go build and go vet on cmd/mysql/exporter for every push and PR. Since the exporter has its own go.mod, it runs independently from the main project build.

PR #2162, opened 2026-01-21
Feature/mariadb gtid support

santiagolertora:
### Database name
MySQL/MariaDB

# Pull request description

This PR addresses critical code review feedback for MariaDB GTID support:

**Problems fixed:**
1. **Missing bounds check (CRITICAL)** - Accessing `e.RawData[5:9]` without checking if the array has at least 9 bytes could cause a panic in production with malformed binlog events
2. **Code duplication** - `waitReplicationIsDone()` had identical code in both MariaDB and MySQL branches, reducing maintainability

**Changes made:**
- Added length check before accessing event header to prevent panic: `if len(e.RawData) < 9`
- Removed duplicate code in `waitReplicationIsDone()` by consolidating identical branches
- Improved error messages with contextual information (shows actual byte length)

**Potential panic scenario (now prevented):**
1. A corrupted or truncated binlog event with header < 9 bytes arrives
2. Without the bounds check, accessing `e.RawData[5:9]` would panic
3. This would crash the binlog server in production

**Code duplication issue:**
1. Both MariaDB and MySQL branches in `waitReplicationIsDone()` called the exact same functions
2. This made the code harder to maintain and understand

N/A - These are preventive fixes identified during code review

**Build verification:**
go build ./internal/databases/mysql/...
# Build successful, no errors
```

**Error message improvement example:**
Before: Silent panic on malformed event
After: `Binlog event header too short for server ID: 7 bytes` (logged as warning, function returns empty string safely)

### DIFF
--- a/internal/databases/mysql/binlog_server_handler.go
+++ b/internal/databases/mysql/binlog_server_handler.go
@@ -133,39 +133,6 @@ func waitReplicationIsDone(flavor string) error {
 	}
 }
 
-// extractMariaDBGTIDFromEvent extracts the MariaDB GTID from a GTID event.
-// MariaDB GTID format in binlog events: domain-server-sequence
-func extractMariaDBGTIDFromEvent(e *replication.BinlogEvent) string {
-	// Skip event header (19 bytes)
-	data := e.RawData[replication.EventHeaderSize:]
-
-	// MariaDB GTID event structure:
-	// - Sequence number: 8 bytes (uint64)
-	// - Domain ID: 4 bytes (uint32)
-	// - Flags: 1 byte
-
-	if len(data) < 13 {
-		tracelog.WarningLogger.Printf("MariaDB GTID event too short: %d bytes", len(data))
-		return ""
-	}
-
-	sequence := binary.LittleEndian.Uint64(data[0:8])
-	domain := binary.LittleEndian.Uint32(data[8:12])
-
-	// Server ID is in the event header; ensure the raw data is long enough
-	if len(e.RawData) < 9 {
-		tracelog.WarningLogger.Printf("Binlog event header too short for server ID: %d bytes", len(e.RawData))
-		return ""
-	}
-	serverID := binary.LittleEndian.Uint32(e.RawData[5:9])
-
-	// Format: domain-server-sequence
-	gtid := strconv.FormatUint(uint64(domain), 10) + "-" +
-		strconv.FormatUint(uint64(serverID), 10) + "-" +
-		strconv.FormatUint(sequence, 10)
-
-	return gtid
-}
 
 // detectBinlogFlavor detects whether we're working with MySQL or MariaDB binlogs.
 // It tries to detect from the replica source connection, or defaults to MySQL.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by chipitsine** [comment]
is `cmd/mysql/exporter` included in purpose ?

another question - can code deduplication be separate PR ? for the sake of review simplicity

**Comment by chipitsine** [comment]
this test failure looks related to the PR itself https://github.com/wal-g/wal-g/actions/runs/22400432066/job/64847478007?pr=2162#step:8:7587

File: internal/databases/mysql/binlog_server_handler.go [hunk 1]
```diff
@@ -148,8 +213,18 @@ func sendEventsFromBinlogFiles(logFilesProvider *storage.ObjectProvider, pos mys
 			gtidEvent := &replication.GTIDEvent{}
 			err = gtidEvent.Decode(e.RawData[replication.EventHeaderSize:])
 			tracelog.ErrorLogger.FatalOnError(err)
-			u, _ := uuid.FromBytes(gtidEvent.SID)
-			lastSentGTID = u.String() + ":1-" + strconv.Itoa(int(gtidEvent.GNO))
+
+			// Check flavor to determine how to extract GTID
+			if flavor == mysql.MariaDBFlavor {
```
ostinru: Is this code reachable? As I can see - MariaDB has its own `MARIADB_GTID_EVENT` type. If it uses own events - we can use event parser from `go-mysql-org` project.

https://github.com/go-mysql-org/go-mysql/blob/v1.7.0/replication/const.go#L99-L100

santiagolertora: You're right, that code was unreachable. MariaDB uses MARIADB_GTID_EVENT (type 162), not GTID_EVENT (type 33), so the flavor check inside the GTID_EVENT case would never trigger for MariaDB binlogs.Fixed it by switching on the event type directly: MARIADB_GTID_EVENT is now handled with the MariadbGTIDEvent parser from go-mysql-org, and GTID_EVENT stays for MySQL. Also removed the manual binary decoding since the library already does that.

internal/databases/mysql/mariadb_gtid_filter.go
```diff
@@ -0,0 +1,177 @@
...
+	for domainID, gtid := range minuend.Sets {
+		subGtid, existsInSub := subtrahend.Sets[domainID]
+
+		if !existsInSub {
+			// Domain only exists in minuend, include it
+			result.Sets[domainID] = gtid.Clone()
+		} else if gtid.SequenceNumber > subGtid.SequenceNumber {
+			// Domain exists in both, but minuend has a higher sequence
+			// Include the difference (we can't represent ranges, so we use the higher value)
+			result.Sets[domainID] = gtid.Clone()
```
ostinru: Since MySQL GTID sets can represent ranges of transactions (with gaps) and we can merge this sets without loosing any information. Here it looks more dangerous.

At high level: all this filtering code helps reduce amount of binlogs uploaded & downloaded (when master swithced, we are trying to upload all binlogs it have). Keeping in mind the fact that it is better to upload binlog rather than loose it - we can keep old behavior (upload all binlogs for MariaDB).

OR

we should have 100% clear picture of what we are doing here. @santiagolertora do you have good deep dive into MariaDB GTIDs and more importantly how MariaDB applies it during PiTR? I thought that we had issue with PiTR before - it worked something like binlog+position navigation in MySQL and then failed due to duplicate events (on an attempt to apply event that we already applied before)

ostinru:
Merged as https://github.com/wal-g/wal-g/pull/2475

PR #2163, opened 2026-01-25
fix(deps): update libsodium to 1.0.21

jorsol:
Update libsodium to 1.0.21

### DIFF
--- a/link_libsodium.sh
+++ b/link_libsodium.sh
@@ -4,7 +4,8 @@ set -e
 
 readonly CWD=$PWD
 readonly OS=$(uname)
-readonly LIBSODIUM_VERSION=${LIBSODIUM_VERSION:-1.0.20}
+readonly ARCH=$(uname -m)
+readonly LIBSODIUM_VERSION=${LIBSODIUM_VERSION:-1.0.21}
 
 test -d tmp/libsodium || mkdir -p tmp/libsodium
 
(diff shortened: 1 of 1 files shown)


link_libsodium.sh
```diff
@@ -13,13 +14,18 @@ cd tmp/libsodium
...
+CONFIGURE_ARGS="--prefix ${PWD} --disable-debug --disable-dependency-tracking --enable-static --disable-shared"
 if [[ "${OS}" == "SunOS" ]]; then
   # On Illumos / Solaris libssp causes linking issues when building wal-g.
   CONFIGURE_ARGS="${CONFIGURE_ARGS} --disable-ssp"
-fi      
+fi   
 
-./configure ${CONFIGURE_ARGS}
+LOCAL_CFLAGS="-O2"
+if [[ "${OS}" == "Linux" ]] && [[ "${ARCH}" == *arm* || "${ARCH}" == "aarch64" ]]; then
```
chipitsine, line 24: wow.  you run db on arm64 ?

chipitsine:
did you maybe compared performance before and after ?

link_libsodium.sh
jorsol: Not on production (yet), but I work a lot with containers under Mac so I end up using arm64 and hit and issue without that flag.

jorsol: This is the error when building under arm64:

`crypto_ipcrypt/ipcrypt_armcrypto.c:84:5: note: use '-flax-vector-conversions' to permit conversions between vectors with differing element types or numbers of subparts`

chipitsine: I wonder how did we made recent aarch64 binaries ...

https://github.com/wal-g/wal-g/actions/runs/21212997721

chipitsine: maybe we shall add macos ci job

PR #2171, opened 2026-02-02
fix: close file at creation

wooffie:
### Database name
Any

# Pull request description

Call of `os.Create(filePath)` create file handle. Clean code should close this handle to avoid leak after function exit


chipitsine:
thank you for contribution!

PR #2172, opened 2026-02-03
Add AZURE_BLOB_STORE_API_VERSION env var

annielzy:
### Database name
NA

# Pull request description

The problem is that we have some Azure Stack Hub environments that don't support the azure blob store api version of the current SDK.

Without setting AZURE_BLOB_STORE_API_VERSION to "2020-10-02", when trying to connect to Azure Blob Storage, wal-g error with:
```
<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidHeaderValue</Code><Message>The value for one of the HTTP headers is not in the correct format.
RequestId:6a029202-bdfd-0551-46b5-91b0f69a4298
Time:2026-01-29T21:31:51.3910908Z</Message><HeaderName>x-ms-version</HeaderName><HeaderValue>2026-02-06</HeaderValue></Error>
```

NA

### DIFF
--- a/pkg/storages/azure/storage.go
+++ b/pkg/storages/azure/storage.go
@@ -73,8 +73,8 @@ func buildClientOptions(config *Config) *container.ClientOptions {
 		},
 	}
 	if config.BlobStoreAPIVersion != "" {
-		opts.ClientOptions.PerCallPolicies = append(
-			opts.ClientOptions.PerCallPolicies,
+		opts.PerCallPolicies = append(
+			opts.PerCallPolicies,
 			&apiVersionPolicy{apiVersion: config.BlobStoreAPIVersion},
 		)
 	}


chipitsine:
just curious. if I would need to set env variable, I would probably do it either by running "export" before invoking wal-g or passing it directly "VAR=value wal-g ..."

why do code need to be modified to set a variable ?

annielzy:
The current version of wal-g does not provide a way to configure the Azure Blob Storage API version—it's hardcoded in the SDK/client. This PR introduces a new environment variable (AZURE_BLOB_STORE_API_VERSION) specifically to make the API version configurable for environments like Azure Stack Hub, which may not support the latest default version.

The environment variable is new and was named for this purpose; it is not something that could previously be set via the environment or config. That's why code changes are needed—to add support for overriding the API version via this env var.

PR #2183, opened 2026-02-19
ci: cleanup some space on runner

chipitsine:
when increasing tests density, I encountered lack of disk space. Let's make some preparation and clean some space in advance


debebantur (commented):
Hi! Can you give some more details, please?
- What was the problem, you encountered? Is it about tests amount, or huge tests?
- I thought that github runners here use plain ubuntu-latest image.  Does it really have swapfile?

chipitsine:
I tried to combine redis aof and rdb tests in single run instead of two


github runners use ubuntu + lot of cached software and apparently even swapfile

PR #2184, opened 2026-02-20
Change folder DeleteObjects([]string) to DeleteObjects([]Object)

debebantur:
Now we pass only object keys to DeleteObjects func. However some storages use additional info for deletion (s3 uses VersionID for versioned buckets). Passing whole object list allows to use all available info.


ostinru (approved):
LGTM

kronwerk (approved):

PR #2158, opened 2026-01-19
Add production-ready Docker image for MariaDB PITR

santiagolertora:
### Database name
MySQL / MariaDB
### Title
```
feat(mysql): Add production-ready Docker image for MariaDB PITR
```

### Description
```markdown
## Summary

Add production-ready Docker image with automated Point-in-Time Rec

- Intelligent 4-phase entrypoint for automated restore and PITR
- Kubernetes manifests with StatefulSet and CronJob examples
- E2E test suite adapted to mariadb_tests format
- Automatic backup detection and restoration on empty datadir

## Changes

- `docker/mariadb-walg/`:
  - `Dockerfile` - Multi-stage build with WAL-G integration
  - `docker-entrypoint-walg.sh` - Orchestrates restore → prepare →
  - `binlog-replay-helper.sh` - Helper for binlog application
  - `kubernetes.yml` - Production K8s deployment examples
  - `README.md` - Complete setup and usage documentation
- `docker/mariadb_tests/scripts/tests/pitr_mariabackup_full_test.s

## Features

### 4-Phase Entrypoint Logic
1. **Detection**: Check if datadir is empty
2. **Restore**: Fetch backup and prepare with mariabackup
3. **Permissions**: Fix ownership (mysql:mysql)
4. **PITR**: Start MariaDB and apply binlogs to target timestamp

### Environment Variables
- `WALG_RESTORE_FROM` - Backup name or `LATEST`
- `WALG_PITR_TIMESTAMP` - Target time for PITR
- `WALG_S3_PREFIX` / cloud storage config
- Standard MariaDB variables (`MYSQL_ROOT_PASSWORD`, etc.)

## Test Plan

- Full E2E test: backup → insert data → binlog backup → PITR resto
- Tested with MariaDB 10.5, 10.6, 10.11
- Verified GTID handling and binlog replay accuracy
- Tested empty datadir detection and automatic restore
- Kubernetes manifests validated on production cluster

## Usage Example

```bash
docker run -e WALG_S3_PREFIX=s3://bucket/backups \
  -e WALG_RESTORE_FROM=LATEST \
  -e WALG_PITR_TIMESTAMP=2024-01-15T10:30:00Z \
  -e MYSQL_ROOT_PASSWORD=secret \
  mariadb-walg:latest
```

PR #2185, opened 2026-02-26
pitr_mariabackup_full_test.sh: skip if mariadb is lower than 10.8

chipitsine:


chipitsine:
@santiagolertora


test added in https://github.com/wal-g/wal-g/pull/2158 fails
is it supposed to pass on mariadb-10.3.39 ?

(due to current docker test design, we are bound to mariadb-10.3.x, which seems to be too old for "mariadb + pitr +gtid" machinery)

ostinru:
> is it supposed to pass on mariadb-10.3.39 ?

According to https://endoflife.date/mariadb - 10.3 had reached End-of-Life. I think that we can bump version in tests to 10.6+ or even 10.11+?

@santiagolertora which MariaDB versions are popular in MariaDB world? Is it common to keep MariaDB versions that reached end of life?

chipitsine:
it would be nice to bump, however starting with https://github.com/wal-g/wal-g/commit/aee1f180f1da9afce1efdbd14c44f7051080c81e we use mariadb shipped with ubuntu distro (because mariadb do not make releases for ubuntu 20.04 anymore)

chipitsine:
@santiagolertora , I'm trying to run `pitr_mariabackup_full_test.sh` manually command by command and I wonder how it was supposed to work.

variables $WALG_MYSQL_BINLOG_END_TS, $WALG_MYSQL_CURRENT_BINLOG set in https://github.com/wal-g/wal-g/blob/master/docker/mariadb_tests/scripts/tests/pitr_mariabackup_full_test.sh#L7 are empty 

I added debug "echo" to the test and it confirmed my observation.

I'm thinking of removing that test, we do not benefit from it currently. If you have it working for you, it would be nice to have it fixed

santiagolertora:
@chipitsine thanks for looking into this! You're right  the test was designed for MariaDB 10.8+ where GTID support is more mature.
regarding the empty vars,
WALG_MYSQL_BINLOG_END_TS and WALG_MYSQL_CURRENT_BINLOG are meant to be set dynamically by WAL-G during binlog-replay
These are passed to the WALG_MYSQL_BINLOG_REPLAY_COMMAND which uses them as template variables
The test might need better setup to populate these before use, minimum MariaDB version to 10.8+ (or 10.11 LTS) and 11.8x as you suggested with Ubuntu 24.04.

MariaDB 10.3 is EOL and GTID support there was incomplete. In production, most users are on 10.6+ or 10.11 LTS and 11.8.x. The test is valuable for these versions.

 I have a more complete testing suite in docker/mariadb-walg/ that works with newer MariaDB versions - I initially kept it separate to avoid touching the existing docker/mariadb_tests/ structure, but we could consider unifying them later.

chipitsine:
but it fails even on 10.11 (I've updated image definition to switch to more recent mariadb): https://github.com/wal-g/wal-g/actions/runs/22498863895/job/65181221399

chipitsine:
do you have it passed in your own fork ?

ostinru:
> (because mariadb do not make releases for ubuntu 20.04 anymore)

I'd like to update test images from 20.04 to 22.04. (And left 18.04 for retro-versions of MySQL).

chipitsine:
np, I'll move that to dedicated PR.

chipitsine:
https://github.com/wal-g/wal-g/pull/2189

btw, with ubuntu 22.04 is shipped mariadb-10.6
likely we'll have to add mariadb from mariadb repo as it was previously

chipitsine:
@santiagolertora what shall we do with pitr_mariabackup_full_test.sh ? it fails even on mariadb-10.11 (at least for me). if it works for you, please let us know.

otherwise I'm afraid we have to delete it

santiagolertora:
@chipitsine @ostinru I've looked into this in detail. There were two separate problems: bugs in the test script itself, and a limitation in WAL-G's binlog-replay logic. My PR #2195fixes both.
Problems in pitr_mariabackup_full_test.sh:
The S3 bucket mariadb_pitr_mariabackup_full was missing from docker-compose.yml, so the test would fail on storage initialization
No FLUSH LOGS + binlog-push before backup-push, so pre-backup events were not archived and the binlog boundary was unclear
No cleanup of WALG_MYSQL_BINLOG_DST at the start or before replay, so leftover binlogs from previous runs could interfere
The row count assertion checked for 5 users but only 4 were inserted (Alice, Bob, Charlie, David)
The xtrabackup_binlog_info path was hardcoded instead of using $MYSQLDATA

About the empty variables: $WALG_MYSQL_BINLOG_END_TS and $WALG_MYSQL_CURRENT_BINLOG are set by WAL-G internally when it executes the replay command for each binlog file. They are not meant to be set by the test. If you ran the commands manually outside of wal-g binlog-replay, they would naturally be empty. This is the same behavior as in the MySQL tests at least as i find
Issues on the restore with binlog-replay fetches binlogs starting from BinLogStart (the earliest binlog uploaded before the backup). But the backup already contains all data up to a specific position in a later binlog file. When mysqlbinlog replays from the start, it re-applies CREATE TABLE, INSERT, etc. that already exist in the restored data, causing "Table 'users' already exists" errors. This has nothing to do with the MariaDB version.
The WAL-G fix: Both xtrabackup and mariabackup write the exact binlog file and position in xtrabackup_info (the binlog_pos field), which is already available in --extra-lsndir but WAL-G was not parsing it. My PR reads this during backup-push, stores it in the sentinel (BinLogFileName, BinLogFilePosition), and during binlog-replay skips binlogs already covered by the backup and passes --start-position for the boundary binlog. I also added WALG_MYSQL_BINLOG_START_POSITION as a new env var available in the replay command, and the test script uses it via ${WALG_MYSQL_BINLOG_START_POSITION:+--start-position="$WALG_MYSQL_BINLOG_START_POSITION"}.

Changes on testenvironment:
docker-compose.yml: added the mariadb_pitr_mariabackup_full S3 bucket
docker/mariadb/export_common.sh: added mariadb_version_check and mariadb_get_version helpers, improved mariadb_kill_and_clean_data reliability, and made mysql_set_gtid_from_backup compatible with both xtrabackup_binlog_info and mariadb_backup_binlog_info (MariaDB 11.8+ renames these files)
The test passes cleanly on 10.11 with no workarounds needed. Please don't remove it

chipitsine:
abandoning in favour of https://github.com/wal-g/wal-g/pull/2195

PR #2191, opened 2026-03-04
tests: do not compress docker images (caches are compressed already),

chipitsine:
improvements:

- time (1 minute less)
- space, images are not needed after load
- wal-g/golang is reduced by 50Mb

### DIFF
--- a/docker/golang/Dockerfile
+++ b/docker/golang/Dockerfile
@@ -22,7 +22,7 @@ ENV PATH=$GOPATH/bin:/usr/local/go/bin:$PATH
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 
 RUN wget https://github.com/Kitware/CMake/releases/download/v3.31.0/cmake-3.31.0-linux-x86_64.tar.gz && \
-    tar xf cmake-3.31.0-linux-x86_64.tar.gz -C /usr --strip-components=1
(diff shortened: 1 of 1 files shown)


chipitsine:
copilot suggestions are irrelevant, but interesting to try

Makefile
```diff
@@ -85,7 +85,7 @@ pg_integration_test: clean_compose
 		make install_and_build_pg;\
 		make pg_build_image;\
 	else\
-		docker load -i ${CACHE_FILE_DOCKER_PREFIX};\
+		docker load -i ${CACHE_FILE_DOCKER_PREFIX} && rm ${CACHE_FILE_DOCKER_PREFIX};\
```
ostinru, line 88: What is idea behind removing cache after loading it into docker?

chipitsine: it is one-time cache, i.e. it is built in "docker build" job, delivered to "test" job and after that it will not going anywhere. but if we delete it, we'll save some space on runner

maybe that use case should be better documented in comments

ostinru: For local development test runs save/load is meaningless... so we should never get to this IF branch.
Looks good.

ostinru (approved):

Makefile
chipitsine: well, I thought of an idea to load all images when job start and remove them.
currently, we delete only images that are loaded.

but we can think about it later

chipitsine: and for local dployment you also need to define CACHE_xxx variables, without them it is undefined behaviour

PR #2195, opened 2026-03-07
mysql: add precise binlog position tracking for PITR

santiagolertora:
## Summary

Store exact binlog file and position from xtrabackup_info/mariadb_backup_info
in backup sentinel during backup-push. Use this information during binlog-replay
to skip already-restored binlogs and start from the exact position, eliminating
duplicate DDL/DML errors during point-in-time recovery.

- Parse binlog_pos from xtrabackup_info (xtrabackup) and mariadb_backup_info
  (MariaDB 11.8+) with fallback between naming conventions
- Add BinLogFileName and BinLogFilePosition to StreamSentinelDto
- Pass WALG_MYSQL_BINLOG_START_POSITION env var to replay command
- Detect mariabackup in addition to xtrabackup/xbstream
- Update PITR test to use --start-position for precise replay
- Support both xtrabackup_binlog_info and mariadb_backup_binlog_info

### Database name

MySQL / MariaDB

### What this PR fixes

During binlog-replay for point-in-time recovery, WAL-G replays binlogs starting
from BinLogStart (the first binlog uploaded before the backup). Since the backup
already contains all data up to a specific position within a later binlog,
mysqlbinlog re-applies DDL/DML that already exists in the restored data, causing
"Table already exists" and similar errors. This made the PITR test
(pitr_mariabackup_full_test.sh) fail even on MariaDB 10.11+.

Both xtrabackup and mariabackup already record the exact binlog file and position
in their info file (xtrabackup_info / mariadb_backup_info) via the binlog_pos
field. This file is available in --extra-lsndir but WAL-G was not parsing it.

This PR parses that information, stores it in the backup sentinel, and uses it
during binlog-replay to skip binlogs already covered by the backup and start
from the exact position in the boundary binlog.

Related: #2185


ostinru:
Hi, @santiagolertora ! It is an improvement to existing codebase. This patch starts applying transactions from `filename`+`position`. This will work well for single-node clusters. 
However, for HA-clusters it will be more complicated. I suppose that after switchover/failover we will upload binlogs from new master. And during PiTR we will try to apply all binlogs. And I think it will fail because new binlogs may have transactions from previous master (relay logs). In PostgreSQL it should work because they have history files that explicitly tells which binlog range should be applied. For MariaDB it doesnt?

So, I don't understand how it should work.

santiagolertora:
Superseded by #2497. Closing.

PR #2196, opened 2026-03-08
tests: switch mysql to ubuntu 22.04 && drop ubuntu 20.04

chipitsine:


docker-compose.yml
```diff
@@ -32,13 +32,6 @@ services:
     image: wal-g/ubuntu:18.04
     container_name: wal-g_ubuntu
 
-  ubuntu_20_04:
-    build:
-      dockerfile: ./docker/ubuntu/Dockerfile_20_04
```
ostinru, line 37: Do we need `docker/ubuntu/Dockerfile_20_04` left in workspace?

chipitsine: I do not think so

PR #2206, opened 2026-03-19
Optimize oplog listing for large-scale MongoDB deployments

Apelsin234:
Summary                                                                                                                                                                       
   
  - When dealing with a large number of oplog archives in storage, the previous implementation listed all oplog archives in the S3 prefix before selecting the required segment.
   This caused significant performance degradation for databases with extensive oplog history.
  - For oplog_push, we only need the last oplog archive to determine the starting point, yet the entire prefix was being listed.                                                
  - For oplog_replay, we needed a specific range of oplogs but had to fetch the full listing first.                                                                             
                                                                                                                                                                                
  Changes                                                                                                                                                                       
                                                                                                                                                                                
  - Introduced ListOplogArchivesSegment(startAfter, endBefore) method that fetches only the required range of oplog archives instead of listing everything.                                                                                                  
  - Added unit tests for the new functionality.

### DIFF
--- a/internal/databases/mongo/archive/loader.go
+++ b/internal/databases/mongo/archive/loader.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
 	"github.com/wal-g/wal-g/internal/compression"
@@ -130,9 +131,13 @@ func (sd *StorageDownloader) ListOplogArchives() ([]models.Archive, error) {
 	return sd.ListOplogArchivesSegment(nil, nil)
 }
 
-// ListOplogArchivesStartAfter fetches all oplog archives existed in storage after the given startAfter.
+// ListOplogArchivesSegment fetches all oplog archives existed in storage after the given startAfter.
 func (sd *StorageDownloader) ListOplogArchivesSegment(startAfter *string, endBefore *string) ([]models.Archive, error) {
-	tracelog.DebugLogger.Printf("Listing %s", sd.oplogsFolder.GetPath())
+	tracelog.DebugLogger.Printf("Listing %s with startAfter `%s` and endBefore `%s`",
+		sd.oplogsFolder.GetPath(),
+		aws.StringValue(startAfter),
+		aws.StringValue(endBefore),
+	)
 	var objects []storage.Object
 	var err error
 
@@ -144,7 +149,6 @@ func (sd *StorageDownloader) ListOplogArchivesSegment(startAfter *string, endBef
 		}
 	} else {
 		objects, _, err = folder.ListFolderSegment(startAfter, endBefore)
-
 		if err != nil {
 			return nil, fmt.Errorf("can not list oplog archives folder: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


internal/databases/postgres/names_metadata_test.go
```diff
@@ -256,7 +256,7 @@ func TestFetchDtoWithFilesMetadata(t *testing.T) {
 	}
 
 	oldObjects := DatabaseObjectsInfoOld{Oid: 12, Tables: map[string]uint32{"t1": 1, "t2": 2}}
-	newObjects := postgres.DatabaseObjectsInfo{Oid: 12, Tables: map[string]postgres.TableInfo{"t1": postgres.TableInfo{Oid: 1, Relfilenode: 2}, "t2": postgres.TableInfo{Oid: 3, Relfilenode: 4}}}
```
Apelsin234, line 259: it was `make fmt`

internal/databases/mongo/archive/loader.go
```diff
@@ -126,10 +127,27 @@ func (sd *StorageDownloader) DownloadOplogArchive(arch models.Archive, writeClos
 
 // ListOplogArchives fetches all oplog archives existed in storage.
 func (sd *StorageDownloader) ListOplogArchives() ([]models.Archive, error) {
+	return sd.ListOplogArchivesSegment(nil, nil)
+}
+
+// ListOplogArchivesStartAfter fetches all oplog archives existed in storage after the given startAfter.
```
NotAmigo: ListOplogArchivesSegment

pkg/storages/s3/folder.go
```diff
@@ -218,6 +219,46 @@ func (folder *Folder) ListFolder() (objects []storage.Object, subFolders []stora
...
+
+	listFunc := func(commonPrefixes []*s3.CommonPrefix, contents []*s3.Object) bool {
+		cont := true
+		for _, prefix := range commonPrefixes {
+			subFolder := NewFolder(folder.s3API, folder.uploader, *prefix.Prefix, folder.config)
+			subFolders = append(subFolders, subFolder)
+		}
+		for _, object := range contents {
+			key := *object.Key
+			if endBefore != nil && key >= *endBefore {
```
NotAmigo: Are you sure you want to make endBefore exclusive?

.gitignore
```diff
@@ -45,3 +45,9 @@ venv/
 
 #Ctags tags file
 tags
+
+
+# AI tools files and directories
```
NotAmigo, line 50: Is it supposed to be added?

internal/databases/mongo/binary/oplog_reply.go
```diff
@@ -80,6 +78,38 @@ func RunOplogReplay(ctx context.Context, mongodbURL string, replayArgs ReplyOplo
 	return HandleOplogReplay(ctx, replayArgs.Since, replayArgs.Until, oplogFetcher, oplogApplier)
 }
 
+func resolveOplogReplaySequence(
+	downloader archive.Downloader,
+	since, until models.Timestamp,
+) (archive.Sequence, error) {
+	// because of oplog archives are write every 30 second intervals, we need to expand segment
+	sinceStr := buildOplog(models.Timestamp{TS: since.TS - 300, Inc: 0})
```
NotAmigo: Why Since.Ts - 300 if every 30 second?

Apelsin234: 30 seconds is the default interval, and I chose to add a small buffer here since listing ~10 archives is relatively cheap.

.gitignore
Apelsin234: Yes, I think so — a lot of people are likely using AI tools at the moment

pkg/storages/s3/folder.go
Apelsin234: In general, ranges are usually defined as exclusive, so I went with that approach. That said, I’m not entirely sure how appropriate it is in this specific case.

internal/databases/mongo/archive/loader.go
Apelsin234: ok

pkg/storages/s3/folder.go
Apelsin234: fixed, guess in this case better is inclusive end

internal/databases/mongo/archive/loader.go
```diff
@@ -126,10 +127,25 @@ func (sd *StorageDownloader) DownloadOplogArchive(arch models.Archive, writeClos
 
 // ListOplogArchives fetches all oplog archives existed in storage.
 func (sd *StorageDownloader) ListOplogArchives() ([]models.Archive, error) {
-	tracelog.DebugLogger.Printf("Listing %s", sd.oplogsFolder.GetPath())
```
VGoshev, line 129: Looks like we missed this debug print. Let's go back and add it (it is very useful for understanding what is happening when you read the logs).

Apelsin234: fixed

PR #2211, opened 2026-03-20
ci: add trivy vuln scanner

chipitsine:
let's add trivy
later we'll decide whether to keep govulncheck or not


debebantur:
Thank you for this pr. Govulncheck still fails, is this WIP?

chipitsine:
golvulncheck is telling us to update `github.com/jackc/pgproto3/v2@v2.3.3`
however, no fix has been released yet

it will be red until resolved, hope couple weeks or so

ostinru:
It seems that `trivy-action` was Ok at the time this PR created... 
=)
https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23

chipitsine:
I was a bit wrong regarding `github.com/jackc/pgproto3/v2@v2.3.3`
that transitive dependency comes together with `github.com/greenplum-db/gp-common-go-libs@v1.0.22` which is archived and no chance it will update its pgproto dependency

chipitsine:
let's harden permission as well

ostinru:
> I was a bit wrong regarding `github.com/jackc/pgproto3/v2@v2.3.3` that transitive dependency comes together with `github.com/greenplum-db/gp-common-go-libs@v1.0.22` which is archived and no chance it will update its pgproto dependency

We have https://github.com/apache/cloudberry-go-libs - however we should test compatibility with Greenplum 6. Created an issue for research: #2219

chipitsine:
superseded  by https://github.com/wal-g/wal-g/pull/2400

PR #2222, opened 2026-03-25
catchup: add parallel send/receive via WALG_UPLOAD_DISK_CONCURRENCY

x4m:
After the initial handshake on the coordinator connection, the sender reads WALG_UPLOAD_DISK_CONCURRENCY (default 1, backward-compatible), announces the worker count via a new IsStart/NumWorkers message, and opens N TCP connections to the same port.  Each connection is served by a dedicated goroutine on both sides.

On the send side, filepath.Walk is now done in a separate goroutine that feeds a buffered channel; N worker goroutines pull file tasks from the channel and call sendOneFile concurrently.  sendDeletedFiles, StopBackup, and the final pg_control write always run on worker 0.

On the receive side, the listener stays open until all N workers have connected, then each worker processes its stream in a goroutine.

The old sendFileCommands function is replaced by walkIntoChannel + the channel/goroutine fan-out in HandleCatchupSend.

### DIFF
--- a/internal/databases/postgres/catchup_send_recieve_handler.go
+++ b/internal/databases/postgres/catchup_send_recieve_handler.go
@@ -10,10 +10,12 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
 	"github.com/wal-g/wal-g/internal/compression"
+	conf "github.com/wal-g/wal-g/internal/config"
 	"github.com/wal-g/wal-g/internal/databases/postgres/errors"
 	"github.com/wal-g/wal-g/internal/ioextensions"
 	"github.com/wal-g/wal-g/utility"
@@ -27,10 +29,12 @@ func HandleCatchupSend(pgDataDirectory string, destination string) {
 		tracelog.ErrorLogger.Fatal("Our system lacks System Identifier, cannot proceed")
 	}
 	tracelog.ErrorLogger.FatalOnError(err)
-	writer, decoder, encoder := startSendConnection(destination)
+
+	// Coordinator connection (worker 0): handshake only.
+	writer0, decoder0, encoder0 := startSendConnection(destination)
 
 	var control PgControlData
-	err = decoder.Decode(&control)
+	err = decoder0.Decode(&control)
 	tracelog.ErrorLogger.FatalOnError(err)
 	tracelog.InfoLogger.Printf("Destination control file %v", control)
 	tracelog.InfoLogger.Printf("Our system id %v", *info.systemIdentifier)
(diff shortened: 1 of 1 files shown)


internal/databases/postgres/catchup_send_recieve_handler.go
```diff
@@ -54,29 +59,78 @@ func HandleCatchupSend(pgDataDirectory string, destination string) {
 			lsn, control.Checkpoint)
 	}
 
-	sendFileCommands(encoder, pgDataDirectory, fileList, control.Checkpoint)
+	numWorkers, err := conf.GetMaxUploadDiskConcurrency()
```
ostinru, line 62: We always encourage contributors to document their features. I think that it will be useful to add/update it in `docs/PostgreSQL.md`

ostinru (approved):
LGTM

internal/databases/postgres/catchup_send_recieve_handler.go
```diff
@@ -10,10 +10,12 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
```
vadv, line 13: Need `"time"` for `SetDeadline` below.

```suggestion
	"sync"
	"time"
```

internal/databases/postgres/catchup_send_recieve_handler.go
```diff
@@ -239,47 +304,77 @@ func sendOneFile(path string, info fs.FileInfo, wasInBase bool, checkpoint LSN,
...
+		tracelog.ErrorLogger.Fatal("Expected IsStart command from sender")
+	}
+	numWorkers := startCmd.NumWorkers
+	if numWorkers < 1 {
+		numWorkers = 1
+	}
+	tracelog.InfoLogger.Printf("Receiving with %v parallel workers", numWorkers)
+
+	// Accept the remaining worker connections on the same port.
+	decoders := make([]*gob.Decoder, numWorkers)
```
vadv, line 351: If sender dies after `IsStart` but before opening all worker connections, receiver hangs here forever.

All worker connections arrive within milliseconds in normal operation (sequential `net.Dial` right after `IsStart`). 30s is a safe upper bound — if they haven't arrived by then, sender is dead.

```suggestion
	// Accept the remaining worker connections on the same port.
	// SetDeadline on a TCPListener only affects Accept calls, not already accepted connections.
	listen.(*net.TCPListener).SetDeadline(time.Now().Add(30 * time.Second))

	decoders := make([]*gob.Decoder, numWorkers)
```

vadv (commented):
If sender crashes after sending `IsStart` but before opening all worker connections, the receiver blocks on `Accept()` forever — no timeout, no context. The process hangs and holds the port.

See inline suggestion for a minimal fix.

PR #2225, opened 2026-03-26
add save configfile env

### LINKED ISSUE #2031: [FEATURE] set config file path from environment

Instead of specifying the --config /path parameter each time you run the wal-g command, set the WALG_CONFIGFILE (or something like this) variable to specify the path to the configuration file.

Removes the need to specify the parameter for the command each time you run it

TiredBeer:
Fixes #2031

### DIFF
--- a/internal/databases/greenplum/backup_push_handler.go
+++ b/internal/databases/greenplum/backup_push_handler.go
@@ -140,13 +140,11 @@ func (bh *BackupHandler) buildBackupPushCommand(contentID int) string {
 		backupPushArgsLine,
 		// pass the config file location
 		fmt.Sprintf("--config=%s", conf.CfgFile),
-	}
-	cmd = append(cmd,
 		// forward stdout and stderr to the log file
 		"&>>", formatSegmentLogPath(contentID),
 		// run in the background and get the launched process PID
 		"& echo $!",
-	)
+	}
 
 	cmdLine := strings.Join(cmd, " ")
 	tracelog.InfoLogger.Printf("Command to run on segment %d: %s", contentID, cmdLine)


cmd/common/common.go
```diff
@@ -51,7 +51,7 @@ func Init(cmd *cobra.Command, dbName string) {
 	cmd.InitDefaultVersionFlag()
 	conf.AddConfigFlags(cmd, hiddenConfigFlagAnnotation)
 
-	cmd.PersistentFlags().StringVar(&conf.CfgFile, "config", "", "config file (default is $HOME/.walg.json)")
+	cmd.PersistentFlags().StringVar(&conf.CfgFile, "config", "", "config file (default is $HOME/.walg.json, can also be set via WALG_CONFIGFILE env var)")
```
debebantur: Seems that this line triggers linters because it is too long

cmd/common/common.go
debebantur, line 49: On command init we call conf.InitConfig. We can redefine conf.CfgFile here if it is empty. This will allow not to call GetConfigFilePath() each time

cmd/common/common.go
debebantur: Lets name this var WALG_CONFIG_PATH it sounds better

internal/databases/greenplum/ao_check_length_handler.go
```diff
@@ -95,15 +95,16 @@ func (checker *AOLengthCheckHandler) buildCheckAOLengthCmd(contentID int, backup
 	cmd := []string{
 		// nohup to avoid the SIGHUP on SSH session disconnect
 		"nohup", "wal-g",
-		// config for wal-g
```
debebantur, line 98: Seems that now there is no need to change this. If conf.CfgFile equals to "", segment command will try to read path from env

internal/databases/greenplum/backup_push_handler.go
```diff
@@ -138,13 +138,16 @@ func (bh *BackupHandler) buildBackupPushCommand(contentID int) string {
 		fmt.Sprintf("--content-id=%d", segment.ContentID),
 		// actual arguments to be passed to the backup-push command
 		backupPushArgsLine,
-		// pass the config file location
-		fmt.Sprintf("--config=%s", conf.CfgFile),
```
debebantur, line 142: same here

internal/databases/greenplum/backup_fetch_handler.go
```diff
@@ -290,7 +290,9 @@ func (fh *FetchHandler) buildFetchCommand(contentID int) string {
 		fmt.Sprint(segment.DataDir),
 		fmt.Sprintf("--content-id=%d", segment.ContentID),
 		fmt.Sprintf("--target-user-data=%s", segUserData.QuotedString()),
-		fmt.Sprintf("--config=%s", conf.CfgFile),
```
debebantur, line 293: same here

internal/databases/greenplum/ao_check_length_handler.go
TiredBeer: fixed it

internal/databases/greenplum/backup_push_handler.go
```diff
@@ -140,11 +140,13 @@ func (bh *BackupHandler) buildBackupPushCommand(contentID int) string {
 		backupPushArgsLine,
 		// pass the config file location
 		fmt.Sprintf("--config=%s", conf.CfgFile),
+	}
+	cmd = append(cmd,
```
debebantur: Seems there is no need to append this separately. You can just leave it in definition

PR #2231, opened 2026-04-06
Add timeline switch check

debebantur:
### Database name
Postgres

# Pull request description

In some cases (when cluster was restored) there are records in .history file, that are not actually present in cluster history.
Integrity check fails because it tries to switch timeline that was not actually switched.
The logic is: if wal file with current timeline is present on LSN before switch, then this switch was not actually present

### DIFF
--- a/internal/databases/postgres/wal_segment_runner.go
+++ b/internal/databases/postgres/wal_segment_runner.go
@@ -107,7 +107,7 @@ func (r *WalSegmentRunner) getNextSegment() WalSegmentDescription {
 		// for example after restoring cluster
 		// We will skip such records
 		if _, fileExists := r.walFolderSegments[nextSegment]; fileExists {
-			tracelog.WarningLogger.Printf("timeline switch at LSN %s is invalid. Skipping it\n", record.lsn.String())
+			tracelog.WarningLogger.Printf("timeline switch at LSN %s is not from our history. Skipping it\n", record.lsn.String())
 			return nextSegment
 		}
 		// switch timeline if current WAL segment number found in .history record


internal/databases/postgres/wal_segment_runner.go
```diff
@@ -99,12 +99,21 @@ func (r *WalSegmentRunner) ForceMoveNext() {
...
 	nextTimeline := r.currentWalSegment.Timeline
+	nextSegmentNo := r.currentWalSegment.Number.previous()
+	nextSegment := WalSegmentDescription{Timeline: nextTimeline, Number: nextSegmentNo}
 	if record, ok := r.timelineSwitchMap[r.currentWalSegment.Number]; ok {
+		tracelog.DebugLogger.Printf("found timeline switch at LSN %s\n", record.lsn.String())
+		// Sometimes there are records in .history that are not actually present in cluster history
+		// for example after restoring cluster
+		// We will skip such records
+		if _, fileExists := r.walFolderSegments[nextSegment]; fileExists {
+			tracelog.WarningLogger.Printf("timeline switch at LSN %s is invalid. Skipping it\n", record.lsn.String())
```
ostinru:
```suggestion
			tracelog.WarningLogger.Printf("timeline switch at LSN %s is not from our history. Skipping it\n", record.lsn.String())
```

internal/databases/postgres/wal_segment_runner.go
```diff
@@ -99,12 +99,21 @@ func (r *WalSegmentRunner) ForceMoveNext() {
 // getNextSegment calculates the next segment
 func (r *WalSegmentRunner) getNextSegment() WalSegmentDescription {
 	nextTimeline := r.currentWalSegment.Timeline
+	nextSegmentNo := r.currentWalSegment.Number.previous()
```
ostinru, line 102: `NextSegmentNo := previous()`. Why?

debebantur: We run check in reverse order. From now to oldest backup

PR #2245, opened 2026-04-23
use atomic types over atomic ops

serprex:


internal/testutils/network_error_folder.go
```diff
@@ -46,14 +47,14 @@ func NewNetworkErrorFolder(sourceFolder storage.Folder, networkErrorAfterByteSiz
...
+		readFromFile: make(map[string]*atomic.Int64),
 	}
 }
 
 type TestFolder struct {
 	storage.Folder
 	maxReadSize  int64
-	readFromFile map[string]*int64
+	readFromFile map[string]*atomic.Int64
 	mutex        sync.Mutex
```
ostinru, line 58: This one seems to be mutex-protected.

ostinru: Ah... it is used as atomics in `utility/with_size_reader.go`.
👍

ostinru (approved):

PR #2247, opened 2026-04-23
use slices from stdlib

serprex:
generic methods offer better typing than sort.Slice

### DIFF
--- a/internal/backup_file_description.go
+++ b/internal/backup_file_description.go
@@ -1,7 +1,7 @@
 package internal
 
 import (
-	"sort"
+	"slices"
 	"time"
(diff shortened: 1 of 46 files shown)


ostinru (approved):

ostinru:
Not sure that `slices.ContainsFunc` without lambdas is more readable than old school `for`-loop... But golang doesn't have lambdas =/

ostinru (approved):

ostinru:
@x4m, it seems that this one should be merged with code ownership overrides.

VGoshev (approved):

PR #2249, opened 2026-04-24
Skip transactions mysql replica has already applied

serprex:
Fixes flakey mysql_base_tests failures in pitr_binlog_server_reconnection_test.sh:
```
Slave SQL for channel '': Fatal error: ... unexpected statement was found in the middle of the transaction. Query: 'CREATE DATABASE sbtest;', Error_code: 1593
```

`HandleBinlogDumpGTID` ignored replica's GTID set, streaming every binlog file from Pos=4 Pre-backup events (e.g. init.sql's CREATE DATABASE) reached replicas whose GTID_PURGED already covered them. Now compare each GTID event's GTID to `requiredGTIDs` set by COM_BINLOG_DUMP_GTID. Transactional events are skipped; specific non-transactional events reset it

`buildGTIDStr` formatted every tracked UUID as `uuid:1-N`, ignoring that group number ranges need not start at 1 (such as after SET GLOBAL GTID_PURGED or any GTID gap)

Replace custom GTID tracking with go-mysql's `GTIDSet`, replacing `buildGTIDStr`

### DIFF
--- a/internal/databases/mysql/binlog_server_handler.go
+++ b/internal/databases/mysql/binlog_server_handler.go
@@ -17,7 +17,6 @@ import (
 	"github.com/go-mysql-org/go-mysql/replication"
 	"github.com/go-mysql-org/go-mysql/server"
 	mysqldriver "github.com/go-sql-driver/mysql"
-	"github.com/google/uuid"
 	"github.com/wal-g/tracelog"
 
 	"github.com/wal-g/wal-g/internal"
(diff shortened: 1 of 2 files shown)


teem0n (dismissed):

ostinru:
Am I correct that scenario for this error
```
Fatal error: ... unexpected statement was found in the middle of the transaction. Query: 'CREATE DATABASE sbtest;', Error_code: 1593
```
is following:
1. run PiTR
2. reconnect in the middle of transaction
3. start replaying from the very beginning (DDL)
4. MySQL throws an error, instead of skipping transaction

Is this correct?

ostinru (approved):

serprex:
@ostinru yes

ostinru:
> @ostinru yes

sounds like a bug in MySQL?

PR #2251, opened 2026-04-25
fix: skip checksum validation for torn pages covered by WAL  (FPI)

boosterKRD:
### Database name
PostgreSQL

### Pull request description

During `backup-push`, WAL-G validates page checksums as it reads data files. When PostgreSQL writes an 8KB page, the OS may split the write into two 4KB operations. If WAL-G reads the page between those two writes (a "torn page"), the checksum will not match — this is reported as corruption, but it could be a false positive.

With `full_page_writes=on`, PostgreSQL writes a Full Page Image (FPI) into WAL for the first modification of every page after a checkpoint. `pg_backup_start` forces a checkpoint, so any page whose LSN > `backupStartLSN` was modified after that checkpoint and therefore has an FPI in WAL. Even if that page is torn or genuinely corrupted at backup time, recovery will restore it from the WAL FPI. Reporting these pages as corrupt has no value and causes false alarms.

In `isPageCorrupted`: if `full_page_writes` is enabled AND `page.LSN > backupStartLSN`, skip the checksum check and return `false`.
Pages written before the backup started (`LSN <= backupStartLSN`) are still fully validated — no FPI guarantee exists for them.

### Testing:
- Unit and integration tests were written with Claude's assistance; I have reviewed and verified each one manually twice.
- The Docker integration test completes in ~15 seconds on an ancient MacBook (Intel chip)) and will not meaningfully increase the overall test suite runtime.

### P.S. 
The full_page_writes check is intentional and important. On Copy-on-Write filesystems (ZFS, btrfs, etc.) torn pages cannot occur, so administrators may disable full_page_writes for performance. In that case — regardless of page LSN — a checksum mismatch is genuine corruption, and WAL-G must always emit a WARNING. The fix correctly gates the skip on full_page_writes=on, so COW users are fully protected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -361,6 +361,11 @@ services:
     container_name: wal-g_pg_delete_retain_after_test
     command: /tmp/tests/delete_retain_after_test.sh
 
+  pg_checksum_corruption_test:
+    <<: *pg_test_common
+    container_name: wal-g_pg_checksum_corruption_test
+    command: /tmp/tests/checksum_corruption_test.sh
+
   pg_full_backup_test:
     <<: *pg_test_common
     container_name: wal-g_pg_full_backup_test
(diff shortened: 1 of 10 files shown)


boosterKRD:
Hi folks
I would be glad if someone could review this PR

PR #2255, opened 2026-04-28
Make local & sftp filesystem storage PutObject atomic and durable

saolof:
The "local" filesystem storage option is occasionally used with mounted network shares, so it is worth making it correct so that wal files are always durably archived when the archive command returns successfully.

The sftp implementation is also common in a plain unix server setting. So while sftp limits the durability guarentees you can offer to "the server has acked it", using the atomic rename pattern should make it a bit harder to corrupt a wal file or leave it in an unfinished state, and the code changes are very similar so I did that too.

### Database name
All, this changes the storage layer only.

# Pull request description
This makes the PutObject implementation for fs and sftp atomic with a posix rename, after fsyncing the data in the fs case.

The "local" filesystem storage option is occasionally used with mounted network shares, so it is worth making it correct so that wal files are always durably archived when the archive command returns successfully.

The sftp implementation is also common in a plain unix server setting. So while sftp limits the durability guarentees you can offer to "the server has acked it", using the atomic rename pattern should make it a bit harder to corrupt a wal file or leave it in an unfinished state, and the code changes are very similar so I did that too.

The PutObject is not atomic.

Creating a network partition during transfer, or using lazyfs.

### DIFF
--- a/pkg/storages/fs/folder.go
+++ b/pkg/storages/fs/folder.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


saolof:
Just a note on sftp: it's possible to tell most sftp servers to fsync if they support the fsync@openssh.com extension to the protocol. But wal-g depends on github.com/pkg/sftp v1.11.0  while the sftp package added support to wrap that in 1.13 , which confused me for a while. And it is using a protocol extension, not the base protocol.

I'm leaving that change for a potential followup commit, since the most important thing is ensuring that the files are on a different machine which requires an fsync for nfs but is handled by flush on close for sftp.

pkg/storages/fs/folder.go
```diff
@@ -101,21 +107,44 @@ func (folder *Folder) ReadObject(objectRelativePath string) (io.ReadCloser, erro
 func (folder *Folder) PutObject(name string, content io.Reader) error {
 	tracelog.DebugLogger.Printf("Put %v into %v\n", name, folder.subPath)
 	filePath := folder.GetFilePath(name)
-	file, err := OpenFileWithDir(filePath)
+	randomPostfix, err := storage.NewTimestampRandomTag()
```
serprex: should consistently call suffix

saolof: Done

saolof:
Fixed the windows build issue which was the last remaining test failure. 

It turns out that windows has no directory fsync equivalent and unlike linux it provides no way of enforcing that a file won't be orphaned if the power cuts out. Also rebased and squashed while I was at it to make the git history cleaner if this gets merged

saolof:
Rebased, no further changes expected.

serprex (approved):

PR #2269, opened 2026-05-04
ci cleanup: remove go get

serprex:
make already handles this

### DIFF
--- a/.github/workflows/dockertests.yml
+++ b/.github/workflows/dockertests.yml
@@ -66,11 +66,6 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
         id: go
 
-      - name: Get dependencies
-        if: steps.cache-images.outputs.cache-hit != 'true'
-        run: |
-          go get -v -t ./...
-
       # build images
       - name: Build images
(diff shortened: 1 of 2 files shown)


.github/workflows/dockertests.yml
ostinru: This step is also... weird

.github/workflows/dockertests.yml
```diff
@@ -98,10 +93,6 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
         id: go
 
-      - name: Get dependencies
-        run: |
-          go get -v -t -d ./...
-
```
ostinru, line 111: Then, we don't need `liblzo2-dev` and `Set up Go 1.x`. However it may be used in Makefile > deps > go_deps

serprex: it is used by Makefile > deps > go_deps

.github/workflows/dockertests.yml
serprex: #2183

ostinru (approved):

PR #2270, opened 2026-05-04
ci: cache mongo base images, drop noop mongo_test matrix entry

serprex:
Split tests_func/Dockerfile.mongodb into two stages stored separately:
- Dockerfile.mongodb-base: ubuntu+mongo apt install + supervisor/configs, parameterized only by MONGO_VERSION/PACKAGE/REPO
- Dockerfile.mongodb: install wal-g over wal-g/mongo-base

Image building job now runs `make mongo_save_images` alongside pg_save_image

No longer re-downloads mongodb-{org,enterprise} from repo.mongodb.{org,com} for each variant

Also drop `make mongo_test` matrix entry: target only runs `mongo_build`, which every other mongo matrix entry already exercises inside its docker image

### DIFF
--- a/.github/workflows/dockertests.yml
+++ b/.github/workflows/dockertests.yml
@@ -36,6 +36,7 @@ jobs:
             ${{ env.CACHE_FILE_UBUNTU_18_04 }}
             ${{ env.CACHE_FILE_UBUNTU_22_04 }}
             ${{ env.CACHE_FILE_GOLANG }}
+            ${{ env.CACHE_FOLDER }}/wal-g_mongo-base-*.tar
           key: ${{ env.IMAGES_CACHE_KEY }}
 
       - name: Cleanup space on runner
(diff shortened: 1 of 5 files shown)


chipitsine:
I have some proposal on that, but I didn't have a time to try :(

currently we build some "base" image with "wal-g" included and later we use it as source image and install mongodb to it.

what if we instead will use mongodb image (either existing one or build our own and push it to ghcr.io) and will build wal-g and add it to that image ?

mongodb changes very infrequently. no need to install it each time

chipitsine:
no questions on dropping mongo_test, can we merge is as separate PR ?

chipitsine:
it might work for all heavily built packages like cloudberry. etc

chipitsine:
hmm... your implementation is similar to what I proposed.
the only concern is to avoid building it each time and packing/unpacking via gha cache ...

chipitsine:
I'll have a look at timings.
we do not build mongodb, installing should be pretty fast (especially compared with packing/unpacking over cache layer)


well, for packages we build (like redis, cloudberry, ... ) it would be definitely more significant.

chipitsine:
well, all packages manipulation (not only mongodb, but also other packages) takes ~30 sec
which is much less than building redis, cloudberry, ...

PR #2273, opened 2026-05-05
daemonclient-rs

serprex:
introduce a parallel daemonclient written in Rust, whose binary is a tenth the size of daemonclient

postgres wants to run without overcommit, which doesn't play nice with Go GC,
leading to having to fence off memory with cgroups in https://github.com/ubicloud/ubicloud/pull/5266


chipitsine:
Wow. So frim the beginning we run postgres and wal-g on the same machine and them have different preferences on overcommit????

serprex:
> Wow. So frim the beginning we run postgres and wal-g on the same machine and them have different preferences on overcommit????

indeed, at least with wal-g as daemon it's easier to configure cgroups / GOMEMLIMIT to reign it in. but this was something to give credit to pgBackRest's use of C, & it'd've been nice if wal-g had been wal-rs for postgres at least, but it's taken a much longer time to get suitably good cloud sdks / database clients on Rust compared to Go, so I understand why things didn't go that way

interested in having https://github.com/wal-g/wal-rs project?

@byucesoy recently wrote blog on this: https://www.ubicloud.com/blog/postgresql-and-the-oom-killer-why-we-use-strict-memory-overcommit

x4m:
FWIW I'm fine with in-repo or separate repo, as you wish.

@serprex, did I understand it right that you will support it in future? :)

ostinru:
> Wow. So frim the beginning we run postgres and wal-g on the same machine and them have different preferences on overcommit????

`vm.overcommit_memory` is system wide property. There is no way to set it differently for other cgroups.

However in my tests (on `influxdata/telegraf`) golang is quite good at memory management. `memory committed` that is counted in `vm.overcommit_memory` was close to RES (resident set size).

@serprex , do you use `vm.overcommit_memory = 2` in production? Have you had a `/proc/[pid]/maps` snapshot for a leaking wal-g?

cmd/daemonclient-rs/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+fn split_flag(s: &str) -> (&str, Option<&str>) {
+    match s.find('=') {
+        Some(i) => (&s[..i], Some(&s[i + 1..])),
+        None    => (s, None),
+    }
+}
+
+fn run(argv: Vec<String>) -> ExitCode {
```
ostinru, line 192: Let's use some argument parsing librare? e.g. `clap`?

chipitsine:
> introduce a parallel daemonclient written in Rust, whose binary is a tenth the size of daemonclient
> 
> postgres wants to run without overcommit, which doesn't play nice with Go GC,
> leading to having to fence off memory with cgroups in https://github.com/ubicloud/ubicloud/pull/5266

That was my observation of hosting redis and MySQL on the same server, as they have opposite expectation from overcommit

It is system wide and not managed by cgroup as far as I know

serprex:
> FWIW I'm fine with in-repo or separate repo, as you wish.
> 
> @serprex, did I understand it right that you will support it in future? :)

will bring up with ubicloud today, if we're actively using it we'll definitely be supporting it

cmd/daemonclient-rs/src/main.rs
serprex: goal was to match go parsing, but seems fair to not have this be drop-in replacement, especially since it drops connection timeout parameter already

serprex:
@ostinru yes, =2, no leak, but without https://github.com/ubicloud/ubicloud/pull/5041 we were finding OOM killer making VM unreachable. we're trying to tighten our margins as much as possible, in 5041 Burak allocates a budget for go services, & in a pending PR I'm trying to enforce that with GOMEMLIMIT/cgroups. wal-g in go isn't untenable, only unideal. in the end otel machinery means the problem is here to stay regardless

serprex:
based on experiments we're thinking https://github.com/wal-g/walg_archive is the better way to go for reducing daemon client overhead

Issue #1805, opened 2024-10-10
Feature request: add statsd walg_s3_upload_time metric


M1ha-Shvn:
### Database name
PostgreSQL

# Issue description

### Describe your problem
It would be useful to monitor s3 upload latency. Currently I'm doing it using a bash wrapper on wal-push command, it would be cool if wal-g could produce this metric itself

ostinru:
Could you please define 'upload latency'? Is it roundtrip to storage (S3)?

M1ha-Shvn:
A time of WAL upload to S3 storage. According to our research, different S3-compatible storages have very different upload time differing from 1 second to 10 seconds. Also storages can lag sometimes and give high WAL concurrency on uploading server. I want to see this timers in metrics. Currently I use a bash script instead of wal-push like this one:

```bash
start=`date +%s%3N`
wal-push ...  # parameters here
end=`date +%s%3N`
current_dt="`date +'%Y/%m/%d %H:%M:%S.%6N'`"
echo "INFO: ${current_dt} WAL PUSH TIME: $(($end - $start)) ms"

# metrics are uploaded manualy to prometheus here
```

magnus5552:
i'll take this issue

PR #2274, opened 2026-05-05
Add walg_s3_upload_time metric (#1805)

magnus5552:
Added new statsd metric walg_s3_upload_time. I used prometheus Gauge type for collecting metric. I have to use statsd Timing type for export because a Gauge can overwrite values when multiple metrics are sent in a short period

### DIFF
--- a/internal/databases/postgres/wal_push_handler.go
+++ b/internal/databases/postgres/wal_push_handler.go
@@ -9,8 +9,11 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


serprex:
would expect float64 to be int64

magnus5552:
> would expect float64 to be int64

Gauge.Set accepts only a float64, and its value is converted to an int64 when sent to statsd

debebantur (approved):
LGTM

PR #2275, opened 2026-05-07
Cloudberry: Add incremental backups for PAX storage

ostinru:
Add PAX incremental backups.

Basically, new code routes PAX files to new shared `/paxfiles/` folder and records these files in `pax_files_metadata.json`.


ostinru:
@chipitsine , could you help me understand whats wrong with windows smoke tests?

chipitsine:
no idea.

let's try this https://github.com/wal-g/wal-g/issues/2278

chipitsine:
windows does not allow to delete open files. it looks like a good catch.

I'll try patch on top of your branch a bit later

chipitsine:
the followin patch solves the issue (let's give it another look before merge)


```diff
diff -ur wal-g-orig/internal/databases/greenplum/ao_storage_uploader_test.go wal-g-chipitsine/internal/databases/greenplum/ao_storage_uploader_test.go
--- wal-g-orig/internal/databases/greenplum/ao_storage_uploader_test.go 2026-05-08 09:45:12.950271600 +0200
+++ wal-g-chipitsine/internal/databases/greenplum/ao_storage_uploader_test.go   2026-05-08 09:43:36.945638700 +0200
@@ -460,13 +460,12 @@

        cfi := internal.NewComposeFileInfo(f.Name(), fInfo, false, false, header)

-       err = os.Remove(f.Name())
-       if err != nil {
-               t.Log(err)
-       }
+       assert.NoError(t, f.Close())
+       assert.NoError(t, os.Remove(f.Name()))

        err = uploader.AddFile(cfi, meta, &location)
        assert.NoError(t, err)
+       assert.Empty(t, uploader.GetFiles().Files)
 }

 func runSingleTest(t *testing.T, baseFiles greenplum.BackupAOFiles,
diff -ur wal-g-orig/internal/databases/greenplum/pax/storage_uploader_test.go wal-g-chipitsine/internal/databases/greenplum/pax/storage_uploader_test.go
--- wal-g-orig/internal/databases/greenplum/pax/storage_uploader_test.go        2026-05-08 09:45:12.958396600 +0200
+++ wal-g-chipitsine/internal/databases/greenplum/pax/storage_uploader_test.go  2026-05-08 09:43:36.957628800 +0200
@@ -256,7 +256,8 @@
        header.Name = name

        cfi := internal.NewComposeFileInfo(f.Name(), info, false, false, header)
-       _ = os.Remove(f.Name())
+       assert.NoError(t, f.Close())
+       assert.NoError(t, os.Remove(f.Name()))

        err = uploader.AddFile(cfi, pax.RelFileMetadata{}, pax.FileKey{})
        assert.NoError(t, err)
```

my-ship-it:
Please review the PR when you are free, thanks! @Mulily0513

internal/databases/greenplum/tar_ball_composer.go
```diff
@@ -252,8 +328,22 @@ func (c *GpTarBallComposer) addFile(cfi *internal.ComposeFileInfo) error {
...
+	if isPax && cfi.FileInfo.Size() >= c.paxFileSizeThreshold {
+		tracelog.DebugLogger.Printf("%s is a PAX file, will process it through the PAX storage manager",
+			cfi.Path)
+		// this PAX file visible in catalog => it was durably saved to disk. No partial file reads.
+		return c.paxStorageUploader.AddFile(cfi, paxMeta, paxKey)
+	}
+
+	if isAo || isPax {
+		tracelog.DebugLogger.Printf("%s is an AO/AOCS/PAX file, however it will be archived through a regular tar file packer",
+			cfi.Path)
```
ostinru, line 342: After reading Max Yang's feedback,

> Sibling-writer merges happen on still-open writers. OrcWriter::MergeTo()
(orc_writer.cc:556) folds another writer's data in before either has
Closed, then deletes the source's temporary file via DeleteUnstateFile().
Those temporary files never appear in the manifest, so the rule "back up
only what's visible in the manifest" handles this correctly. But "back up
everything I see on disk" would be unsafe.

I am thinking about skipping PAX-related-files that were not seen in `pg_ext_aux.*`. However I still doesn't know scenario when `wal-g` "back up
everything I see on disk" behaviour can be unsafe. Only difference - is to save some space (in backup & on disk).

ostinru:
@Mulily0513, have you had a chance to look at this PR?

debebantur (approved):
LGTM

docker/cloudberry_tests/scripts/tests/pax_storage_test.sh
vovik0134, line 1: It’s not related to this PR, but I’d like to note that it would be great to write such tests using `testcontainers‑go`.

internal/databases/greenplum/pax/metadata.go
```diff
@@ -0,0 +1,53 @@
...
+// FilesMetadataDTO is the shape persisted to `pax_files_metadata.json`.
+type FilesMetadataDTO struct {
+	Files BackupFiles
+}
+
+func NewFilesMetadataDTO() *FilesMetadataDTO {
+	return &FilesMetadataDTO{Files: make(BackupFiles)}
+}
+
+func (m *FilesMetadataDTO) AddFile(localPath, storagePath string, mTime, initialUplTS time.Time,
```
vovik0134: nit: This style of declaring a function signature — where a single type is specified for multiple parameters — is an antipattern and performs poorly when maintaining and refactoring a codebase.

It’s possible that this style is consistent with the rest of the project, in which case no changes are needed here.

However, I’d suggest specifying parameter types individually for each parameter — this approach is more maintainable.

internal/databases/greenplum/pax/query.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/wal-g/wal-g/internal/walparser"
+)
+
+const paxRelationsQuery = `
+SELECT
+    c.oid                                                                  AS pax_oid,
+    md5(quote_ident(n.nspname) || '.' || quote_ident(c.relname))           AS rel_md5,
+    c.relfilenode,
+    c.reltablespace,
+    paxt.auxrelid::regclass::text                                          AS aux_table_fqn
```
vovik0134, line 22: I’m not sure about the performance of `::regclass::text` — it might be worth replacing it with a `self‑join` here. I suggest looking at the execution plans and timings for comparison.

Alternatively, just keep in mind that there could potentially be extra delays in this part of the code.

internal/databases/greenplum/pax/query.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, errors.Wrap(err, "failed to list PAX relations")
+	}
+
+	result := make(RelFileStorageMap)
+	for _, r := range relations {
+		spc := walparser.Oid(r.spcNode)
+		if spc == 0 {
+			spc = dbInfo.TblSpcOid
+		}
+		if err := fetchPaxBlocks(ctx, conn, dbInfo, spc, r, result); err != nil {
```
vovik0134: The `result` variable is passed as an argument to the `fetchPaxBlocks` function and is mutated inside the function. This coding style isn’t very clear. I suggest having `fetchPaxBlocks` return the data, which can then be assigned to `result` on the calling side.

internal/databases/greenplum/pax/query.go
```diff
@@ -0,0 +1,136 @@
...
+			tracelog.WarningLogger.Printf(
```
vovik0134, line 59: It’s not very clear why we only log errors from `fetchPaxBlocks` but don’t terminate execution. I suggest adding a comment in the code to explain this — if this behaviour is indeed intended.

internal/databases/greenplum/pax/query.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		out = append(out, r)
+	}
+	return out, rows.Err()
+}
+
+func fetchPaxBlocks(ctx context.Context, conn *pgx.Conn, dbInfo postgres.PgDatabaseInfo,
+	spcNode walparser.Oid, r paxRelation, dst RelFileStorageMap) error {
+	// table FQN comes from regclass::text, which is pre-escaped by Postgres. It is safe to put into queries.
+	query := fmt.Sprintf("SELECT ptblockname, ptvisimapname, ptexistexttoast FROM %s", r.auxTableFqn)
```
vovik0134, line 90: I suggest using `pgx.Identifier` to safely pass the table name.

vovik0134 (changes requested):

leborchuk (approved):
LGTM, at first step we gather metadata, and then backup files for gathered metadata

internal/databases/greenplum/pax/query.go
ostinru: Tables already escaped because it was fetched as `::regclass::text` (see comment above)

internal/databases/greenplum/pax/query.go
ostinru: This RelFileStorageMap helps to deduplicate files. If file not in this map - it will be backed up as regular file. So, it is safe when some PAX tables doesn't endded up in this map. So, it is Ok to skip table if there is no PAX-related records in auxilary table.

I can see couple reasons why these 2 SQL queries can have different view:
* There is no transaction between `fetchPaxRelations` and `fetchPaxBlocks`. So, it is possible that table was deleted
* it is new with no files.

In both cases, I think, it is safe to proceed with backup.

PR #2276, opened 2026-05-07
improve daemon docs

serprex:
https://github.com/wal-g/walg_archive/issues/3

### DIFF
--- a/docs/PostgreSQL.md
+++ b/docs/PostgreSQL.md
@@ -587,18 +587,48 @@ wal-g wal-restore path/to/target-pgdata path/to/source-pgdata
 
 ### ``daemon``
 
-Archives and fetch all WAL segments in the background. Works with the PostgreSQL archive library `walg_archive` or `walg-daemon-client`.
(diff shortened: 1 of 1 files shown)


docs/PostgreSQL.md
```diff
@@ -587,18 +587,48 @@ wal-g wal-restore path/to/target-pgdata path/to/source-pgdata
...
+```
+
+Commands:
+- `wal-push wal_filepath` — relays to `wal-g wal-push`
+- `wal-fetch wal_name destination_filename` — relays to `wal-g wal-fetch`. On a missing archive, exits `74` (`EX_IOERR`) so PostgreSQL keeps recovering rather than treating it as fatal; matches `wal-fetch` behaviour, see [PR #1195](https://github.com/wal-g/wal-g/pull/1195).
+
+`postgresql.conf` example:
+```conf
+archive_mode = on
+archive_command = 'walg-daemon-client /var/run/wal-g.sock wal-push %p'
```
jorsol: There might be an issue in wal-g daemon, but the wal-push command has hardcoded the path `pg_wal`, so this needs to be set to `%f`

debebantur (approved):

Issue #1949, opened 2025-03-26
[BUG] Commands hang indefinately using invalid or unavailible S3 storage.


seppelucas:
### Database name

PostgreSQL

### WAL-G Version

v3.0.5, v3.0.3

### Describe your problem

When using an s3 backend that is not real or unavailable all wal-g commands hang indefinitely. 
When using wrong credentials we get a clear error message, but when we use a fake server or when our Minio (s3) is unavailable all commands hang until the service becomes available.

An example is a backup-list that took 57 minutes, and only completed when the server finally was available. More importantly the `wal-g wal-push` archive command in our postgreSQL cluster hangs. resulting in WAL-files that are not uploaded and also not retried. 

An assumption here is that the code works fine when using AWS because it is always available, but there are some edge-cases when using another backend. 

We are thinking about adding a timeout to the archive command (like is used in wal-g testcases). But it is better is the S3 connection is checked/tracked for a valid handshake and progress.

Looking into the source we see `//TODO` 's for a lot of the [s3 functions](https://github.com/wal-g/wal-g/blame/master/pkg/storages/s3/folder.go#L72). It looks to me there maybe are some extended settings for timeouts and retries missing like in [this](https://github.com/aws/aws-sdk-go/issues/3386) issue. But i don't have any real experience with GO SDK's.

### Steps to reproduce

Commands hanging indefinitely when using a fake config is easily replicated:
1. Create config file with fake endpoint (does not need to be valid site)
config.env:
``` env
AWS_ACCESS_KEY_ID="test"
AWS_ENDPOINT="https://example.com"
AWS_SECRET_ACCESS_KEY="xxx"
WALG_S3_PREFIX="s3://test"
```
2. Create empty file:
``` bash
touch testfile
```
3. Run command:
``` bash
wal-g wal-push testfile --config config.env
```

### Config and WAL-G logs

An example run:
``` bash
$ time wal-g wal-push testfile --config config.env
INFO: 2025/03/26 11:39:38.930549 Files will be uploaded to storage: default
ERROR: 2025/03/26 11:39:38.931827 Error of parallel upload: open archive_status: no such file or directory
^C
real	11m46,349s
user	0m0,157s
sys	0m0,059s
```
or
```
 $ wal-g backup-list --config config.env
INFO: 2025/03/26 13:34:47.112480 List backups from storages: [default]
^C # Stopped after multiple minutes.

```

stackane:
We've also noticed this in case there are networking-related issues.

[...]

PR #2288, opened 2026-05-10
feat(s3): add S3_REQUEST_TIMEOUT to bound response-header wait

mvanhorn:
## Summary

Refs #1949, #1362.

The S3 storage backend currently relies entirely on aws-sdk-go's default HTTP
client, which sets no timeout on response headers. When the configured
endpoint is wrong, unreachable, or temporarily black-holing connections,
wal-g commands hang indefinitely instead of failing fast. From issue #1949:

```
$ time wal-g wal-push testfile --config config.env
INFO: 2025/03/26 11:39:38.930549 Files will be uploaded to storage: default
ERROR: 2025/03/26 11:39:38.931827 Error of parallel upload: open archive_status: no such file or directory
^C
real	11m46,349s
```

```
$ wal-g backup-list --config config.env
INFO: 2025/03/26 13:34:47.112480 List backups from storages: [default]
^C # Stopped after multiple minutes.
```

This adds an opt-in `S3_REQUEST_TIMEOUT` (in seconds) that bounds how long
the SDK will wait for response headers from the S3 server. Once headers
arrive, body transfer is not constrained -- so multi-GB uploads and
downloads still complete normally.

## Why ResponseHeaderTimeout (not Client.Timeout)

`net/http` exposes several semantically distinct timeouts:

| Timeout | What it bounds | Risk for wal-g |
| --- | --- | --- |
| `http.Client.Timeout` | end-to-end including body transfer | aborts large PUT/GET mid-stream |
| `http.Transport.ResponseHeaderTimeout` | server has N seconds to start responding with headers | none for body transfer |
| `http.Transport.DialContext` timeout | TCP/TLS handshake only | doesn't catch hangs after connect |

`ResponseHeaderTimeout` is the right knob for this issue. The reproducer
(invalid endpoint) hangs at the response-header phase; legitimate large
uploads/downloads complete header negotiation in milliseconds and then
spend the rest of their wallclock streaming bytes -- which we don't want
to cut off.

This matches the pattern of `GCS_CONTEXT_TIMEOUT` in the GCS backend
(opt-in setting bounding a timeout dimension that wouldn't truncate normal
operations) so users who run multi-storage configs see consistent behavior.

## Implementation

| File | Change |
| --- | --- |
| `pkg/storages/s3/configure.go` | New `S3_REQUEST_TIMEOUT` setting, registered in `SettingList`, parsed via existing `setting.IntOptional`; default `0` (disabled). |
| `pkg/storages/s3/storage.go` | New `RequestTimeout time.Duration` field on `Config`. |
| `pkg/storages/s3/session.go` | New `applyRequestTimeout` helper called from `configureSession` before the logging-transport wrapping. Helper extracted to keep `configureSession`'s cyclomatic complexity within the project's `gocyclo=15` budget. |
| `pkg/storages/s3/storage_test.go` | Five tests: setting parses, omission produces zero default, non-integer values error, configured value reaches the underlying `*http.Transport`, omission leaves `ResponseHeaderTimeout` at zero. |
| `docs/STORAGES.md` | Documentation in the S3 "Optional variables" section. |

The `applyRequestTimeout` helper clones the underlying transport rather
than mutating it in place, so we never touch `http.DefaultTransport` (a
process-wide singleton) or any user-supplied transport that may be shared
with other clients.

## Behavior change

`S3_REQUEST_TIMEOUT` is opt-in. The default is `0`, which disables the
header timeout and preserves the existing wal-g behavior exactly. Users
who want the safety net set the env variable to e.g. `30` for thirty
seconds.

## Verification

```
GOEXPERIMENT=jsonv2 go build ./pkg/storages/s3/...
GOEXPERIMENT=jsonv2 go test ./pkg/storages/s3/ -run "TestS3Folder|TestConfigureSession" -v

=== RUN   TestS3FolderRequestTimeoutDefaultIsDisabled       --- PASS (0.00s)
=== RUN   TestS3FolderRequestTimeoutAcceptsValue            --- PASS (0.00s)
=== RUN   TestS3FolderRequestTimeoutRejectsNonInteger       --- PASS (0.00s)
=== RUN   TestConfigureSessionAppliesRequestTimeout         --- PASS (0.00s)
=== RUN   TestConfigureSessionLeavesTimeoutZeroWhenUnset    --- PASS (0.00s)
PASS

GOEXPERIMENT=jsonv2 go vet ./pkg/storages/s3/        # clean
GOEXPERIMENT=jsonv2 golangci-lint run ./pkg/storages/s3/...
0 issues.
gofmt -l pkg/storages/s3/                            # clean
```

(`GOEXPERIMENT=jsonv2` is required because the project's go.mod targets
1.25 and one transitive import uses `encoding/json/v2`. Not related to
this change.)

## Notes / out of scope

- This PR only addresses the response-header hang. Issue #1949 also
  surfaces a related concern about `//TODO`s around connect/retry
  behavior in `pkg/storages/s3/folder.go`. Those touch the deeper
  retry policy and feel like a separate change to me; happy to follow
  up if there's appetite for it.

- I did not change any default. Picking the right default header
  timeout for S3 is a policy decision (AWS itself can take seconds
  to start responding under load), and shipping a default could
  silently change behavior for users with high-latency endpoints.
  Opt-in is the safe step here.

- `S3_REQUEST_TIMEOUT` is parsed as integer seconds via the existing
  `setting.IntOptional` helper, matching the convention used by other
  numeric S3 settings like `S3_MAX_RETRIES`. Happy to switch to a
  duration string (e.g. `"30s"`) if the maintainers prefer.

### DIFF
--- a/docs/STORAGES.md
+++ b/docs/STORAGES.md
@@ -103,6 +103,10 @@ By default, WAL-G checks whether versioning is enabled on the bucket. When the u
 
 Sets batch size for object deletion. By default is set to 1000.
 
+* `S3_REQUEST_TIMEOUT`
+
+Sets the maximum number of seconds WAL-G will wait for the S3 server to send response headers for any request. Once headers arrive, body transfer is not constrained -- multi-GB uploads and downloads still complete. This protects against unresponsive endpoints (wrong host, network black hole, dead Minio, etc.) where commands would otherwise hang indefinitely. The default is `0`, which disables the timeout and preserves prior behavior.
+
 GCS
 -----------
 To store backups in Google Cloud Storage, WAL-G requires that this variable be set:
--- a/pkg/storages/s3/configure.go
+++ b/pkg/storages/s3/configure.go
@@ -48,6 +48,13 @@ const (
 	disable100ContinueSetting      = "S3_DISABLE_100_CONTINUE"
 	enableVersioningSetting        = "S3_ENABLE_VERSIONING"
 	deleteBatchSizeSetting         = "S3_DELETE_BATCH_SIZE"
+	// requestTimeoutSetting bounds how long WAL-G will wait for the S3 server
+	// to send response headers. Once headers arrive, body transfer is not
+	// constrained by this timeout, so large uploads/downloads still succeed.
+	// Default 0 disables the timeout, preserving prior behavior; this avoids
+	// silently changing the contract for existing users on flaky-but-eventually-
+	// responsive endpoints.
+	requestTimeoutSetting = "S3_REQUEST_TIMEOUT"
 )
 
 var SettingList = []string{
@@ -86,6 +93,7 @@ var SettingList = []string{
 	disable100ContinueSetting,
 	enableVersioningSetting,
 	deleteBatchSizeSetting,
+	requestTimeoutSetting,
 }
 
 const (
@@ -104,6 +112,9 @@ const (
 	defaultDisabledRetentionPeriod = -1
 	defaultDisable100Continue      = false
 	defaultDeleteBatchSize         = 1000
+	// defaultRequestTimeoutSeconds = 0 means no response-header timeout. Existing
+	// users see no behavior change unless they opt in by setting S3_REQUEST_TIMEOUT.
+	defaultRequestTimeoutSeconds = 0
 )
 
 // TODO: Unit tests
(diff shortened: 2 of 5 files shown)


serprex:
easier for this after #2280 where this would adjust `buildHTTPClient`

mvanhorn:
Makes sense. I'll wait for #2280 to land, then rebase and retarget the timeout hook to buildHTTPClient.

PR #2289, opened 2026-05-10
replace github.com/wal-g/json with json/v2

serprex:


ostinru (approved):

internal/serializer.go
ostinru: discuss: drop support `RegularJSON` and inline `StreamedJSON` code into places where it used. It will be easier to handle errors and handle a goroutine.

serprex: pushed 2nd commit showing how much code that removes

internal/config/config.go
```diff
@@ -102,7 +102,6 @@ const (
 	MaxDelayedSegmentsCount              = "WALG_INTEGRITY_MAX_DELAYED_WALS"
 	PrefetchDir                          = "WALG_PREFETCH_DIR"
 	PgReadyRename                        = "PG_READY_RENAME"
-	SerializerTypeSetting                = "WALG_SERIALIZER_TYPE"
```
ostinru, line 105: I will suggest leaving this option in config - just for backward compatibility.

serprex: done, with deprecation notice & warning if setting isn't set to stream

ostinru (approved):
LGTM.

PR #2291, opened 2026-05-21
Greenplum: specify hostname for psql CLI

ostinru:
We have observed following error in logs:
```
20260520:17:38:07 wal-g:gpadmin:example.com:030833-[ERROR]:-Unable to switch xlog on cluster on segment 3 on host example.com with error exit status 2: psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.6001" failed: No such file or directory
        Is the server running locally and accepting connections on that socket?
20260520:17:38:07 wal-g:gpadmin:example.com:030833-[DEBUG]:-Command was: ssh -o StrictHostKeyChecking=no gpadmin@example.com PGOPTIONS='-c gp_role=utility' psql -p 6001 -d postgres -c 'select pg_switch_wal();'
```

In case when `unix_socket_directories` is not default `/var/run/postgresql/`, wal-g cannot switch WALs. There are 2 possible sollutions:
1. specify `--host`, so TCP connection will be used. Greenplum hosts always reachable via TCP.
2. directly connect from wal-g to segments and execute SQL


internal/databases/greenplum/restore_point.go
```diff
@@ -183,7 +184,7 @@ func createRestorePoint(conn *pgx.Conn, restorePointName string) (restoreLSNs ma
 							pgOptions = "-c gp_role=utility"
 							switchFunction = "pg_switch_wal()"
 						}
-						return fmt.Sprintf("PGOPTIONS='%s' psql -p %d -d postgres -c 'select %s;'", pgOptions, seg[0].Port, switchFunction)
+						return fmt.Sprintf("PGOPTIONS='%s' psql -h localhost -p %d -d postgres -c 'select %s;'", pgOptions, seg[0].Port, switchFunction)
```
vovik0134: It looks like `--host` option should be a parameter, just like the port is a parameter now.

vovik0134 (changes requested):

vovik0134 (approved):

debebantur (approved):

ostinru:
@copilot resolve the merge conflicts in this pull request

PR #2293, opened 2026-05-22
postgres: support delta with wal summaries

serprex:
@x4m suggested incremental format wouldn't be too useful outside of wal summary support, so native format support has been removed from this PR

### DIFF
--- a/cmd/pg/backup_push.go
+++ b/cmd/pg/backup_push.go
@@ -24,6 +24,7 @@ const (
 	useDatabaseComposerFlag   = "database-composer"
 	deltaFromUserDataFlag     = "delta-from-user-data"
 	deltaFromNameFlag         = "delta-from-name"
+	deltaFromWalSummariesFlag = "delta-from-wal-summaries"
 	addUserDataFlag           = "add-user-data"
 	withoutFilesMetadataFlag  = "without-files-metadata"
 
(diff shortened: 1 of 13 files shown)


cmd/pg/backup_push.go
```diff
@@ -24,6 +26,7 @@ const (
 	useDatabaseComposerFlag   = "database-composer"
 	deltaFromUserDataFlag     = "delta-from-user-data"
 	deltaFromNameFlag         = "delta-from-name"
+	deltaFromWalSummariesFlag = "delta-from-wal-summaries"
```
ostinru, line 27: Let's add this parameter to our documentation in `docs/`

docs/PostgreSQL.md
```diff
@@ -333,6 +333,23 @@ INFO: Delta will be made from full backup.
 INFO: Delta backup from base_000000010000000100000040 with LSN 140000060.
 ```
 
+#### Build the delta map from WAL summaries
+When making a delta backup, WAL-G computes which blocks changed since the base backup. By default it downloads the previous backup's WAL files from storage and replays them, falling back to a full scan of the data directory if that fails (see `WALG_FORCE_WAL_DELTA`).
+
+On PostgreSQL 17+ the server can track changed blocks itself through the `walsummarizer` background process, writing WAL summaries to `$PGDATA/pg_wal/summaries`. The `--delta-from-wal-summaries` flag tells WAL-G to build the delta map straight from these summaries instead of downloading and replaying WAL. This avoids the storage round-trip and the full-scan fallback.
```
chipitsine, line 356: can we maybe limit the scope to 17+ for the sake of simplicity (with summarize_wal = on as requirement)

internal/databases/postgres/wal_summaries.go
```diff
@@ -0,0 +1,327 @@
...
+// union fully covers the requested range with no gaps.
+func selectWalSummariesForRange(files []walSummaryFile, timeline uint32,
+	firstUsedLSN, firstNotUsedLSN LSN) ([]walSummaryFile, error) {
+	if firstNotUsedLSN <= firstUsedLSN {
+		return nil, errors.Errorf("empty LSN range [%s, %s)", firstUsedLSN, firstNotUsedLSN)
+	}
+
+	var kept []walSummaryFile
+	for _, f := range files {
+		if f.timeline != timeline {
```
boosterKRD, line 241: After a timeline switch, summaries from the previous TL are dropped here and the check below fails with WAL summary gap at start, so deltas keep failing until a new full backup is taken, even though the old-TL files are still on disk.

As far as I know/understend, PostgreSQL's own incremental backup walks the timeline history (basebackup_incremental.c) and collects summaries per TL. Could we do the same here, or at least document that a timeline switch requires a full backup?

PR #2298, opened 2026-05-29
Schema-qualify all util queries

reshke:
Safety first


x4m (approved):

ostinru (changes requested):
We have similar query in `internal/databases/greenplum/pax/query.go`. We should have some systematic approach on this - at least cover all queries that QueryRunnar has.

reshke:
> We have similar query in `internal/databases/greenplum/pax/query.go`. We should have some systematic approach on this - at least cover all queries that QueryRunnar has.

yes, lets cover all gp-related queries.

PR #2300, opened 2026-05-29
cmd/pg/exporter: use main golangci-lint config

serprex:
also update to golangci-lint 2.12


cmd/pg/exporter/exporter.go
```diff
@@ -138,7 +138,8 @@ func (b *BackupInfo) GetDeltaOriginName(backups []BackupInfo) string {
 
 	// Extract the part after "_D_" which contains the base backup identifier
 	expectedParent := "base_" + b.BackupName[deltaIndex+3:] // +3 to skip "_D_"
-	for _, candidate := range backups {
```
debebantur, line 141: I like the old style more in this place. Is this linter related change?

serprex: yes, the lint is warning about range copying a pretty large struct, this change avoids that copy

debebantur (approved):

PR #2308, opened 2026-05-30
Add delete flag to skip backup check

debebantur:
### Database name
Postgres

# Pull request description
Before delete garbage we check that non-permanent backups are present.
Let`s add flag to bypass this behaviour

### DIFF
--- a/cmd/pg/delete.go
+++ b/cmd/pg/delete.go
@@ -19,6 +19,7 @@ const DeleteGarbageUse = "garbage [ARCHIVES|BACKUPS]"
 const afterFlag = "after"
 
 var confirmed = false
+var deleteWithoutBackups = false
 var useSentinelTime = false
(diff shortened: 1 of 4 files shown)


cmd/pg/delete.go
```diff
@@ -164,6 +165,8 @@ func init() {
 		&deleteTargetUserData, internal.DeleteTargetUserDataFlag, "", internal.DeleteTargetUserDataDescription)
 	deleteRetainCmd.Flags().StringP(afterFlag, "a", "", "Set the time after which retain backups")
 
+	deleteGarbageCmd.Flags().BoolVar(&deleteWithoutBackups, "without-backup-check", false, "skip check for existing non-permanent backups")
```
ostinru, line 168: Let's add it to `docs/*.md`

ostinru (approved):
LGTM

PR #2311, opened 2026-05-30
update cmd/pg/exporter dependencies, add to dependabot

serprex:


.github/dependabot.yml
```diff
@@ -10,4 +10,24 @@ updates:
       interval: "weekly"
     allow:
```
debebantur: As far as i understood, when `allow` is set, dependabot checks only explicitly defined dependencies. Is there any reason not to update everything?

jorsol (approved):

.github/dependabot.yml
serprex: removed allow

debebantur (approved):

PR #2316, opened 2026-06-01
Feat/retain only one restore point in Greenplum

### LINKED ISSUE #2261: [FEATURE] retain only one restore point in Greenplum to save space

We need an option to retain only one restore point in the past and remove all subsequent wal-files to save space.

When there are a lot of time past since last backup it may be impossible to do next full backup because of many wal-files are pushed. We need an option to keep only one restore point, delete subsequent wal-files and do fresh backup. Then delete the old one

Arin-nana:
### Database name

greenplum

# Pull request description
Implements delete trim-wal command for Greenplum (issue #2261).  Closes #2261
Over time, WAL files accumulate after a backup's restore point and consume significant storage space. There was no way to remove them without deleting the backup itself.

This PR adds the `delete trim-wal BACKUP_NAME|LATEST` command that:

Deletes WAL files past the RestorePointLSN cutoff for every Greenplum segment independently
Removes all `*_restore_point.json` metadata files except the one tied to the specified backup
Supports `--confirm` flag (dry-run by default)
Supports `LATEST` as a backup name

### DIFF
--- a/internal/databases/greenplum/delete_handler.go
+++ b/internal/databases/greenplum/delete_handler.go
@@ -6,15 +6,18 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/wal-g/tracelog"
+
 	"github.com/wal-g/wal-g/internal"
 	conf "github.com/wal-g/wal-g/internal/config"
 	"github.com/wal-g/wal-g/internal/databases/postgres"
 	"github.com/wal-g/wal-g/internal/multistorage"
 	"github.com/wal-g/wal-g/pkg/storages/storage"
 	"github.com/wal-g/wal-g/utility"
-	"golang.org/x/sync/errgroup"
 )
 
 type DeleteArgs struct {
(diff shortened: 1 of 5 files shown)


internal/databases/greenplum/delete_handler.go
```diff
@@ -200,6 +200,106 @@ func (h *DeleteHandler) dispatchDeleteCmd(target internal.BackupObject, delType
...
+	if err != nil {
+		return err
+	}
+
+	deleteConcurrency, err := conf.GetMaxConcurrency(conf.GPDeleteConcurrency)
+	if err != nil {
+		tracelog.WarningLogger.Printf("config error: %v", err)
+	}
+
+	errorGroup, _ := errgroup.WithContext(context.Background())
```
serprex: this method should take context as parameter

internal/databases/greenplum/delete_handler.go
```diff
@@ -200,6 +200,106 @@ func (h *DeleteHandler) dispatchDeleteCmd(target internal.BackupObject, delType
...
+		return err
+	}
+
+	if sentinel.RestorePoint != nil {
+		return h.deleteRestorePointsExcept(*sentinel.RestorePoint, h.args.Confirmed, baseBackupFolder)
+	}
+	return nil
+}
+
+func (h *DeleteHandler) deleteRestorePointsExcept(targetRestorePoint string, confirm bool, baseBackupFolder storage.Folder) error {
```
debebantur: Here we do not need to delete all restore points. We only need to delete those after the given one

internal/databases/greenplum/delete_handler.go
```diff
@@ -200,6 +200,106 @@ func (h *DeleteHandler) dispatchDeleteCmd(target internal.BackupObject, delType
...
+	return segDeleteHandler.DeleteObjectsWhereWithPermanentCheck(confirm, func(object storage.Object) bool {
+		name := object.GetName()
+		if !strings.HasPrefix(name, utility.WalPath) {
+			return false
+		}
+		_, segNo, ok := postgres.TryFetchTimelineAndLogSegNo(name)
+		if !ok {
+			return false
+		}
+		return postgres.WalSegmentNo(segNo) > cutoffSegNo
```
debebantur, line 331: I believe, here we need to check timeline too. There might be a situation, when without this check we will delete files of an older timeline. This does not seem as an expected behaviour
@ostinru what do you think?

ostinru: I think that `wal-g backup-mark` + `wal-g delete garbage` should lead to the same result.
It seems it what @Arin-nana want to achieve.

PR #2322, opened 2026-06-04
Fix permanent backup detection for delete-target and relative storage paths

sshivampeta:
### Database name
PostgreSQL

# Pull request description

When `wal-g delete target` runs, backup and WAL objects are listed from storage subfolders (`basebackups_005/`, `wal_005/`), so object paths are **relative** to those folders (e.g. `base_.../backup_sentinel.json`), not prefixed with `basebackups_005/` or `wal_005/`.

`IsPermanent()` only matched absolute paths with those prefixes, so permanent backups and WALs were treated as non-permanent during delete. That could allow `delete target` to remove a permanent backup.

This PR:
- Parses backup/WAL names from both absolute and relative object paths in `IsPermanent()`
- Checks permanent status by backup name when the object is a `BackupObject` in `makePermanentFunc()`
- Returns `ForbiddenActionError` instead of `Fatalf` when `delete target` tries to delete a permanent backup

1. Create or mark a backup as permanent:
   ```bash
   wal-g backup-mark base_000000010000000000000002
   # or: wal-g backup-push --permanent


internal/databases/postgres/delete_util.go
```diff
@@ -69,20 +69,51 @@ func GetPermanentBackupsAndWals(folder storage.Folder) (map[PermanentObject]bool
 }
 
 func IsPermanent(objectName, storageName string, permanentBackups, permanentWals map[PermanentObject]bool) bool {
-	if strings.HasPrefix(objectName, utility.WalPath) && len(objectName) >= len(utility.WalPath)+24 {
+	if walName, ok := walNameFromObjectPath(objectName); ok {
```
serprex, line 72: this breaks calling IsPermanent on backups, picking up any 24 hex name as wal

I think `IsPermanent` can be left as-is

sshivampeta: Good catch — the default branch in walNameFromObjectPath was incorrectly treating backup paths as WAL because StripWalFileName matches the 24-char hex embedded in backup names. Fixed by rejecting basebackups_.../ and base_-prefixed paths before WAL parsing, and tightening backupNameFromObjectPath. Added unit tests; all TestIsPermanent* cases pass.

sshivampeta (commented):
Good catch — the default branch in walNameFromObjectPath was incorrectly treating backup paths as WAL because StripWalFileName matches the 24-char hex embedded in backup names. Fixed by rejecting basebackups_.../ and base_-prefixed paths before WAL parsing, and tightening backupNameFromObjectPath. Added unit tests; all TestIsPermanent* cases pass.

PR #2340, opened 2026-06-17
replace actions/cache with artifacts

serprex:
cache inconsistent between jobs


debebantur (approved):

chipitsine:
artifacts are limited to 500Mb per repo

chipitsine:
image

PR #2342, opened 2026-06-18
Revert "replace actions/cache with artifacts"

serprex:
Reverts wal-g/wal-g#2340


chipitsine (approved):

chipitsine:
there's about 50Mb in "/tmp" of base image which in theory might be removed by using multistage build

PR #2348, opened 2026-06-20
Replace go-sql-driver with go-mysql-org

serprex:
go-mysql-org already needed for binlog handling

DSN parsing designed to match go-sql-driver, but added support for explicit zlib or zstd compression

### DIFF
--- a/internal/databases/mysql/binlog_server_handler.go
+++ b/internal/databases/mysql/binlog_server_handler.go
@@ -2,7 +2,6 @@ package mysql
 
 import (
 	"context"
-	"database/sql"
 	"encoding/binary"
 	"fmt"
 	"hash/crc32"
@@ -13,10 +12,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-mysql-org/go-mysql/client"
 	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-mysql-org/go-mysql/replication"
 	"github.com/go-mysql-org/go-mysql/server"
-	mysqldriver "github.com/go-sql-driver/mysql"
 	"github.com/wal-g/tracelog"
 	"github.com/wal-g/wal-g/internal"
 	conf "github.com/wal-g/wal-g/internal/config"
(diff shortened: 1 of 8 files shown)


internal/databases/mysql/binlog_server_handler.go
```diff
@@ -128,28 +126,49 @@ func (h *Handler) waitReplicationIsDoneSafe() {
...
-	db, err := sql.Open("mysql", h.replicaSource)
+	dsn, err := parseMySQLDatasource(h.replicaSource)
 	if err != nil {
-		tracelog.ErrorLogger.Fatalf("Failed to connect to replica SQL port: %v", err)
+		tracelog.ErrorLogger.Fatalf("Failed to parse replica datasource: %v", err)
 	}
-	defer db.Close()
+	conn, err := connectMySQL(h.ctx, dsn, "")
+	if err != nil {
+		tracelog.WarningLogger.Printf("Failed to connect to replica SQL port: %v", err)
```
teem0n: Let's either use ErrorLogger.Fatalf here to exit on connection failure, or add a max reconnect counter (e.g. 10 attempts) — right now the loop retries indefinitely and silently, which makes misconfiguration hard to notice.

I guess, counter is even more preferrable...

teem0n (changes requested):

internal/databases/mysql/binlog_server_handler.go
serprex: consolidated connection logic & added counter for number of failed connection attempts

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -128,28 +127,48 @@ func (h *Handler) waitReplicationIsDoneSafe() {
...
 
-	db, err := sql.Open("mysql", h.replicaSource)
+	dsn, err := parseMySQLDatasource(h.replicaSource)
 	if err != nil {
-		tracelog.ErrorLogger.Fatalf("Failed to connect to replica SQL port: %v", err)
+		tracelog.ErrorLogger.Fatalf("Failed to parse replica datasource: %v", err)
 	}
-	defer db.Close()
+	var conn *client.Conn
+	connCount := 0
```
teem0n, line 135: There is not check on connCount : (
Only initialization + increment

serprex: whoops, forgot to add parameter to warning

only having it log connection try count, now that I think about it I also reset on success. do we want a limit before failing?

teem0n: Yes, let's limit by 10 by default

serprex: implemented

internal/databases/mysql/binlog_server_handler.go
```diff
@@ -128,28 +127,55 @@ func (h *Handler) waitReplicationIsDoneSafe() {
...
-		err := db.QueryRowContext(h.ctx, "SELECT @@global.gtid_executed").Scan(&executedStr)
-		if err == nil {
-			replicaSet, _ := mysql.ParseGTIDSet("mysql", executedStr)
-			if replicaSet != nil && replicaSet.Contain(h.sentGTIDs) {
-				tracelog.InfoLogger.Println("Replica has successfully caught up! We are safely done.")
-				os.Exit(0)
+		if conn == nil {
+			if conn, err = connectMySQL(h.ctx, dsn, ""); err != nil {
+				connCount++
+				if connCount == 10 {
```
zkramorev: hi! wouldn't it be better to use `>= 10` here?

teem0n (approved):

PR #2353, opened 2026-06-22
consolidate minio images on RELEASE.2024-06-13T22-53-53Z, use ghcr.io

serprex:


chipitsine:
it is very bad idea to put immutable minio image in cache.

current "cache" is one-time, i.e. it is rebuild on each commit, with very small caching depth.

downloading minio image takes 0 sec, while adding it to cache, compressing, downloading, uncompressing takes much more.

if we want to reduce a load on image registry, we can put minio image to ghcr.io and use as cache

chipitsine (approved):

debebantur (approved):

Issue #2120, opened 2025-11-27
[FEATURE] Refactor prepare_config.sh


debebantur:
### What feature would you like?

Lets rewrite prepare_config.sh using `jq`
This will help us to:

- use standard tooling
- have valid JSONs in `docker/pg_tests/scripts/configs`

mingdaoy:
Hi there, I would like to work on this issue

TiredBeer:
I will handle this task.

PR #2366, opened 2026-06-24
Refactor prepare_config.sh

den-ibr:
### Database name
PostgreSQL

# Pull request description

The PR fixes Issue #2120


serprex (approved):

debebantur:
Seems that we need to use prepare_config.sh in some tests (transfer_backup_test.sh, crypto_test.sh, copy_backup_test.sh, copy_all_test.sh) Like in https://github.com/wal-g/wal-g/pull/2104

chipitsine:
orioledb is using separate scripts, not sure them are inherited from postgres image
likely pg10 might have separate scripts ?

image

debebantur:
> orioledb is using separate scripts, not sure them are inherited from postgres image likely pg10 might have separate scripts ?
> 
> image

I thought that problem is [here](https://github.com/wal-g/wal-g/blob/e4843b849ca05a27655794888341de2662035853/docker/pg_tests/scripts/tests/copy_backup_test.sh#L5). We have 2 valid json files, but then just append them. I believe, we just forgot to use prepare_config here.

As for orioledb, maybe it is better to use jq [there](https://github.com/wal-g/wal-g/blob/e4843b849ca05a27655794888341de2662035853/docker/orioledb/scripts/tests/compatibility_test.sh#L7) too

PR #2370, opened 2026-06-28
daemon: abort Accept when context cancelled

serprex:


internal/databases/postgres/daemon_handler.go
```diff
@@ -232,9 +232,22 @@ func HandleDaemon(ctx context.Context, options DaemonOptions) {
...
+func serve(ctx context.Context, l net.Listener, multiSt *multistorage.Storage) {
+	go func() {
+		<-ctx.Done()
+		utility.LoggedClose(l, "close daemon listener")
+	}()
+
 	for {
 		fd, err := l.Accept()
 		if err != nil {
+			if ctx.Err() != nil {
```
debebantur, line 247: Maybe move this if out of `if err != nil {`? Seems, that context cancellation is not dependent on `l.Accept()` error

serprex: goroutine closes `l` after context cancelled

debebantur (approved):

PR #2383, opened 2026-07-09
Direct IO backup-push IO size configuration to support high-throughput local NVMe disks

achudnovskij:
## Make the O_DIRECT read block count configurable

WALG_DIRECT_IO already reads PGDATA with O_DIRECT during backup-push, but the block count was hardcoded to 32 (128 KiB per read). Expose it via WALG_DIRECT_IO_BLOCK_COUNT (default 32, so unset behavior is unchanged) so operators can request larger aligned reads (e.g. 1024 => 4 MiB), which cuts per-read syscall overhead on fast NVMe. The resolved block size is computed once at reader construction and reused on Seek. A non-positive value is rejected with an error (it would otherwise yield a zero-length aligned buffer).

## backup-push --verify: buffer the O_DIRECT verify tee into block-sized chunks

For --verify, the file read is tee'd to a page-verification consumer through an unbuffered io.Pipe, so the read is throttled to the verifier's lock-step drain rate. On the DIRECT_IO path the file arrives in WALG_DIRECT_IO_BLOCK_COUNT-sized blocks, so add newTeeReadCloserDirectIO, which buffers the tee writes to that block size (verifier woken once per block instead of once per io.Copy chunk) and flushes before closing the pipe so the final block is verified. The non-DIRECT_IO path keeps the original unbuffered newTeeReadCloser unchanged.

### DIFF
--- a/docs/PostgreSQL.md
+++ b/docs/PostgreSQL.md
@@ -27,6 +27,11 @@ To configure how many times failed file will be retried during ```backup-fetch``
 
 An experimental feature that allows you to perform direct_io reads during a ```backup-push``` without flushing the disk cache. To activate it, set the value of the environment variable to `true`.
 
+* `WALG_DIRECT_IO_BLOCK_COUNT`
+
+Set number of disk blocks read and processed in a single batch with direct_io during a ```backup-push```. Default 32 blocks is 32 * 4kb = 128kb
+single read. Larger reads up to 1024 blocks reduce syscall CPU costs and improve throughput at the expense of higher memory consumption.
+
 * `WALG_PREFETCH_DIR`
 
 By default WAL prefetch is storing prefetched data in pg_wal directory. This ensures that WAL can be easily moved from prefetch location to actual WAL consumption directory. But it may have negative consequences if you use it with pg_rewind in PostgreSQL 13.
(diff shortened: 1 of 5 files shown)


serprex (approved):

serprex:
please update docs

achudnovskij:
> please update docs

Done

docs/PostgreSQL.md
```diff
@@ -27,6 +27,11 @@ To configure how many times failed file will be retried during ```backup-fetch``
 
 An experimental feature that allows you to perform direct_io reads during a ```backup-push``` without flushing the disk cache. To activate it, set the value of the environment variable to `true`.
 
+* `WALG_DIRECT_IO_BLOCK_COUNT`
+
+Set number of disk blocks read and processed in a single batch with direct_io during a ```backup-push```. Default 32 blocks is 32 * 4kb = 128kb
+single read. Larger reads up to 1024 blocks reduce syscall CPU costs and improve throughput at the expense of higher memory consumption.
```
serprex:
```suggestion
single read. Larger reads reduce syscall CPU costs and improve throughput at the expense of higher memory consumption.
```

PR #2384, opened 2026-07-10
Retry S3 requests on HTTP 499 (ClientDisconnected)

sebglon:
## Problem

We run WAL-G backups against a MinIO gateway. During a scale-down of the backend pool, one of the MinIO pods was terminated while it had an active multipart upload PUT in flight. MinIO returned its own `ErrClientDisconnected` error (`Code: "ClientDisconnected"`, HTTP status 499) — this is raised whenever the request context is cancelled, which includes the server's own shutdown, not only an actual client-side disconnect.

`ConnResetRetryer.ShouldRetry` did not treat HTTP 499 as retryable, so a single occurrence aborted the entire backup:

```
ERROR: failed to upload 'my-cluster/basebackups_005/base_.../tar_partitions/part_021.tar.lz4' to bucket 'my-backups-bucket':
MultipartUpload: upload multipart failed
	upload id: 01ABCXYZ...
caused by: ClientDisconnected: Client disconnected before response was ready
	status code: 499, request id: ..., host id:
ERROR: Unable to continue the backup process because of the loss of a part 21.
```

This is the same class of problem fixed in #2218 (transient network errors aborting the whole backup) — here it's a transient server-side response instead of a raw socket error.

## Fix

Add an HTTP 499 check to `ConnResetRetryer.ShouldRetry`, following the exact same pattern already used for HTTP 409 (`OperationAborted`). Added a unit test mirroring `TestConnResetRetryerOperationAborted`.

## Testing

```
go test ./pkg/storages/s3/... -run TestConnResetRetryer -v
```
All pass, including the new `TestConnResetRetryerClientDisconnected`. `go vet` and `go build` clean on the package.

### DIFF
--- a/pkg/storages/s3/retryer.go
+++ b/pkg/storages/s3/retryer.go
@@ -35,7 +35,7 @@ func (r ConnResetRetryer) ShouldRetry(req *request.Request) bool {
 	}
 
 	// Some S3-compatible servers (e.g. MinIO) return HTTP 499 with error code
-	// "ClientDisconnected" when the request context is cancelled server-side
+	// "ClientDisconnected" when the request context is canceled server-side
 	// (including on the server's own shutdown), not necessarily because the
 	// client actually disconnected. Treat it as a transient failure worth
 	// retrying rather than aborting the whole backup on a single occurrence.
--- a/pkg/storages/s3/retryer.go
+++ b/pkg/storages/s3/retryer.go
@@ -34,5 +34,15 @@ func (r ConnResetRetryer) ShouldRetry(req *request.Request) bool {
 		return true
 	}
 
+	// Some S3-compatible servers (e.g. MinIO) return HTTP 499 with error code
+	// "ClientDisconnected" when the request context is cancelled server-side
+	// (including on the server's own shutdown), not necessarily because the
+	// client actually disconnected. Treat it as a transient failure worth
+	// retrying rather than aborting the whole backup on a single occurrence.
+	if req.HTTPResponse != nil && req.HTTPResponse.StatusCode == 499 {
+		tracelog.InfoLogger.Printf("S3 returned HTTP 499 (ClientDisconnected), retrying request")
+		return true
+	}
+
 	return r.Retryer.ShouldRetry(req)
 }
--- a/pkg/storages/s3/retryer_test.go
+++ b/pkg/storages/s3/retryer_test.go
@@ -49,6 +49,12 @@ func TestConnResetRetryerOperationAborted(t *testing.T) {
 	assert.True(t, retryer.ShouldRetry(&request.Request{HTTPResponse: resp}))
 }
 
+func TestConnResetRetryerClientDisconnected(t *testing.T) {
+	retryer := NewConnResetRetryer(client.DefaultRetryer{NumMaxRetries: 15})
+	resp := &http.Response{StatusCode: 499}
+	assert.True(t, retryer.ShouldRetry(&request.Request{HTTPResponse: resp}))
+}
+
 func TestConnResetRetryerThrottling(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/storages/s3/retryer.go
```diff
@@ -34,5 +34,15 @@ func (r ConnResetRetryer) ShouldRetry(req *request.Request) bool {
 		return true
 	}
 
+	// Some S3-compatible servers (e.g. MinIO) return HTTP 499 with error code
+	// "ClientDisconnected" when the request context is cancelled server-side
+	// (including on the server's own shutdown), not necessarily because the
+	// client actually disconnected. Treat it as a transient failure worth
+	// retrying rather than aborting the whole backup on a single occurrence.
+	if req.HTTPResponse != nil && req.HTTPResponse.StatusCode == 499 {
```
serprex, line 42: For nonstandard codes can we check full `Status` to make sure it's right code type?

chipitsine: according to what does nginx, 499 is **logged** on server side when there's no client on the wire. if minio mimics that behaviour, no chance to catch it on client side. on client side it would be some tcp level error

sebglon: @chipitsine good point to raise, and it would be a real problem if MinIO followed nginx's convention (499 only ever written to the access log, never actually put on the wire). But that's not what's happening here — MinIO defines `ErrClientDisconnected` as a fully-formed S3 API error (`cmd/api-errors.go`: `Code: "ClientDisconnected"`, `HTTPStatusCode: 499`) that it writes back to the client as a real structured XML error response, not just a log line.

Here's the actual wal-g log from the incident that motivated this PR (bucket/cluster names redacted, everything else verbatim):

```
ERROR: 2026/07/10 08:12:57.641420 failed to upload 'spilo/my-cluster/<uuid>/wal/14/basebackups_005/base_.../tar_partitions/part_021.tar.lz4' to bucket 'my-backups-bucket': MultipartUpload: upload multipart failed
	upload id: 01KX5HBKH2EF6DSBT4BNF8Z6E2
caused by: ClientDisconnected: Client disconnected before response was ready
	status code: 499, request id: 18C0DF760B6B0C78, host id:
ERROR: 2026/07/10 08:12:57.641456 Unable to continue the backup process because of the loss of a part 21.
```

The `request id: 18C0DF760B6B0C78` is the tell: aws-sdk-go only populates that once it has successfully parsed a proper `<Error>` XML body out of an actual HTTP response. A raw socket-level failure (broken pipe / connection reset / EOF) never reaches that code path — it surfaces as `RequestError: send request failed` instead, with no request id (that's exactly the shape of the error #2218 already fixed). So in this case MinIO really did send a parseable 499 response with a request id, it's not a TCP-level disconnect being misreported — the fix retries on a real, well-formed response, not a guessed-at socket error string.

@serprex re: checking the full `Status` — I followed the existing HTTP 409 check just above (status-code-only, no code-string match), so this stays consistent with that pattern. If you'd rather I also assert on the AWS error `Code()` (i.e. `awserr.Error` with `Code() == "ClientDisconnected"`) to be stricter about which servers/situations this applies to, I'm happy to add that — let me know and I'll push an update.

serprex: 409 is a standard http status code, so calls for less ceremony

please reduce slop verbosity in your human interactions

sebglon: @serprex  the code here is 499 not 409.
But it's a standard http code.

serprex: > I followed the existing HTTP 409 check just above

I was replying to why 499 (a non-standard HTTP status code) might differ in handling to 409 (a standard HTTP status code)

chipitsine (approved):

PR #2401, opened 2026-07-12
go.mongodb.org/mongo-driver/v2

serprex:
also update go-pretty to v6 to remove transitive dependency on old mongo-driver

https://github.com/mongodb/mongo-go-driver/blob/master/docs/migration-2.0.md


serprex:
@x4m please merge

PR #2412, opened 2026-07-15
drop noop mongo_test

chipitsine:
as mentioned in https://github.com/wal-g/wal-g/pull/2270 current mongo_test is noop.

let's land it


serprex:
ci fixed by #2413

serprex (approved):

PR #2427, opened 2026-07-20
attempts to make partial backups faster

NotAmigo:
### Database name
MongoDB

# Pull request description

I commented debug print due to efficiency, because, theoretically we spend a lot of time to print this if we have large number of namespaces. Also, I replaced fmt.Sprintf with strings.Builder, because it's much faster

// it can really help

<p>
```bash
any logs here
```
</p>
</details>

### DIFF
--- a/internal/databases/mongo/binary/partial.go
+++ b/internal/databases/mongo/binary/partial.go
@@ -1,7 +1,6 @@
 package binary
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/mongodb/mongo-tools/common/util"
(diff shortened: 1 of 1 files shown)


VGoshev (approved):

internal/databases/mongo/binary/partial.go
```diff
@@ -141,11 +140,15 @@ func GetTarFilesFilter(
...
+	builder := strings.Builder{}
+	builder.WriteRune('/')
+	builder.WriteString(ident)
+	builder.WriteString(".wt")
+	return builder.String()
 }
 
 func localPathFromURI(uri string) string {
-	tracelog.InfoLogger.Printf("URI: %s", uri)
+	//tracelog.DebugLogger.Printf("URI: %s", uri) removed due to efficiency
```
serprex, line 151:
```suggestion
```

NotAmigo: I commented it cause it's still useful for debug, but there is no need usually. The idea was that this debug print spends much time to go to the stdout for every namespaces and I see some efficiency problems here if there are large amount of namespaces, but I am not sure in this suggestion. So, will see

Issue #2425, opened 2026-07-20
[FEATURE] more flexible zstd configuration


chipitsine:
### What feature would you like?

zstd is lovely.

however, chosen way of configuration might not be optimal (additional research is required).

maybe we can benefit from running in parallel (on my tests it provides some benefit)

```
diff --git a/internal/compression/zstd/compressor.go b/internal/compression/zstd/compressor.go
index d19d1403..2bb297e1 100644
--- a/internal/compression/zstd/compressor.go
+++ b/internal/compression/zstd/compressor.go
@@ -2,6 +2,7 @@ package zstd

 import (
        "io"
+       "runtime"

        "github.com/klauspost/compress/zstd"
        "github.com/wal-g/wal-g/internal/ioextensions"
@@ -15,7 +16,7 @@ const (
 type Compressor struct{}

 func (compressor Compressor) NewWriter(writer io.Writer) ioextensions.WriteFlushCloser {
-       zw, err := zstd.NewWriter(writer, zstd.WithEncoderLevel(zstd.SpeedDefault))
+       zw, err := zstd.NewWriter(writer, zstd.WithEncoderLevel(zstd.SpeedDefault), zstd.WithEncoderConcurrency(runtime.GOMAXPROCS(0)), zstd.WithConcurrentBlocks(true) )
        if err != nil {
                panic(err)
        }
```

also, it would be nice to choose compression level, not just `zstd.SpeedDefault`

maybe something like

WALG_COMPRESSION_METHOD={zstd,  zstd-fast, zstd-better, ... }

### Why is this feature useful?

depending of size/time requirements we can either choose fast or better compression

PR #2429, opened 2026-07-20
feat(zstd): configurable compression level and parallel block encoding

Meliceanu:
### Database name
All databases (compression is database agnostic).

# Pull request description

Implements #2425. zstd compression was hardcoded to SpeedDefault with no way to tune it. This adds a `WALG_ZSTD_LEVEL` setting (`fastest`, `default`, `better`, `best`) mapped to the encoder levels, defaulting to the previous behavior when unset and failing explicitly on an unknown value. It also enables parallel block encoding on the zstd writer, which speeds up streaming compression while keeping a valid single frame zstd output, so existing backups and restores are unaffected.

Not a bug. To use: set `WALG_ZSTD_LEVEL=best` (or `fastest`/`better`) together with `WALG_COMPRESSION_METHOD=zstd`.

Not applicable (feature change). Covered by unit tests in `internal/compression/zstd/compressor_test.go`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -72,6 +72,10 @@ To configure where WAL-G stores backups, please consult the [Storages](STORAGES.
 To configure the compression method used for backups. Possible options are: `lz4`, `lzma`, `zstd`, `brotli`, `none`. The default method is `lz4`. LZ4 is the fastest method, but the compression ratio is bad.
 LZMA is way much slower. However, it compresses backups about 6 times better than LZ4. Brotli and zstd are a good trade-off between speed and compression ratio, which is about 3 times better than LZ4. None compression method disables compression.
 
+* `WALG_ZSTD_LEVEL`
+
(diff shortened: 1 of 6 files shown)


chipitsine:
it does not fail on contradictory settings

```
"WALG_COMPRESSION_METHOD": "lz4",
"WALG_ZSTD_LEVEL": "fastest"
```

Meliceanu:
good catch, thanks. it silently ignored the level before, now it errors out if you set WALG_ZSTD_LEVEL with a method that isn't zstd. added tests for it

chipitsine:
great

```
wal-g  backup-push /var/lib/postgresql/18/main
INFO: 2026/07/20 14:35:15.103026 Backup will be pushed to storage: default
ERROR: 2026/07/20 14:35:15.103101 failed to configure compression: WALG_ZSTD_LEVEL is set but the compression method is 'lz4', not 'zstd'

```

chipitsine:
also, smoke tests confirm that time is indeed better (while size is bigger)

chipitsine (approved):

jorsol:
On my tests the memory consumption was more than double when having concurrent blocks enabled, have you measured the impact of having this enabled by default?

chipitsine:
let's revert that part

PR #2443, opened 2026-07-22
Make copy a bit more unified across different databases

secwall:


serprex:
I'd expect unification to remove more lines of code

secwall:
> I'd expect unification to remove more lines of code

Why so?

x4m:
I took a look at this. The shared `Plan`/`ExecuteRaw` framework is a good design and the PR fixes several real bugs, but I think there are a couple of blockers.

### Blocking: plain `gp copy` breaks on pre-existing restore points

`endpoint` is assigned unconditionally in `addGreenplumBackupToPlan` (copy.go:100), so `addGreenplumSegmentHistory` always calls `resolveGreenplumTimeline` (copy.go:292) — including without `--with-history`. Restore point metadata written by earlier wal-g versions has no `TimelineBySegment`, so it falls through to archive inference:

```go
if len(candidates) != 1 {
    return 0, fmt.Errorf("cannot safely infer timeline for legacy Greenplum restore point %q segment %d: ...")
}
```

That fails whenever the endpoint segment number exists on two timelines (normal on the segment where a switch happened) or when the WAL is already gone from the archive. Previously this path never needed the restore point's timeline at all. `HistoryCopyingInfo` derived it from the backup name. So backups that copy fine today will stop copying, with no flag to opt out.

Worth noting: legacy metadata *does* carry the scalar `TimeLine` for the coordinator, but `resolveGreenplumTimeline` ignores it. Grep shows the field is only ever written (restore_point.go:251, backup_push_handler.go:255), never read. Using it as a fallback for `contentID == -1` would remove most of this risk.

### Blocking: new failure mode in `backup-push`

`createRestorePoint` now additionally runs `ReadTimelineBySegment` and hard-fails via `validateRestorePointTimelines`. The keys line up correctly (`gp_segment_id` from `gp_dist_random` matches the content IDs in `LsnBySegment`), so this looks right. But it adds a way for **backup push** to fail for the sake of a copy feature. Could this land separately, or be non-fatal with a warning?

### Re: "I'd expect unification to remove more lines of code"

Agreed, the PR leaves a dead island behind instead of deleting it:

| Function | Callers after this PR |
|---|---|
| `greenplum.GetCopyingInfos` | none (test only) |
| `postgres.BackupCopyingInfo` | only the dead `GetCopyingInfos` |
| `postgres.HistoryCopyingInfo` | only the dead `GetCopyingInfos` |
| `postgres.WildcardInfo` | none |
| `mysql.WildcardInfo` | none |

The PR even adds new logic *inside* the already-dead `GetCopyingInfos` (the "Older callers of GetCopyingInfos expect..." block). After this change the old `copy_common.go` path is only reachable in production from `st copy`. Dropping the above plus their tests should cut the diff noticeably.

### Smaller items

- `restore_point.go:159` — `_, err = queryRunner.ReadTimeline(ctx)` keeps the round-trip but throws the result away.
- `Phase` is declared as a 3-value enum but used as a priority number (`CommitPhase+50`, `+100+depth`, `+200+depth`, `CommitPhase+Phase(depth)`). Undocumented bands that silently collide at chain depth ≥ 50; named constants would be safer.
- `resolveGreenplumTimeline` ends in `panic("unreachable")` (copy.go:335) — an error return would be friendlier.
- Swapping `uploader.Upload` for `folder.PutObject` in `copy_common.go` drops the `UploadedFilesTotal`/`UploadedFilesFailedTotal` counters. The new comment also implies it avoids payload transformation, but `RegularUploader.Upload` never compressed — it just wrapped `PutObject` in metrics.
- A test asserting `compressionExtensions` covers every registered decompressor extension would guard against drift (the list is correct today).

### Nice fixes worth calling out

- The prefix collision is fixed: the old `hasBackupPrefix` matched `basebackups_005/base_1` against `base_10`'s objects; `AddBackupAt` uses a trailing slash plus an exact sentinel path.
- `HandleCopy` used to `return` on a storage config error and report success — now it's `FatalOnError`.
- Following `IncrementFrom` with correct commit ordering (base sentinel before increment) is new and genuinely useful.

x4m (approved):
LGTM

PR #2446, opened 2026-07-24
Remove binamilegacy

serprex:
Also switch some tests_func Dockerfiles to debian slim where change minimal


docker-compose.yml
```diff
@@ -787,17 +794,13 @@ services:
...
-    environment:
-      - ALLOW_NONE_AUTHENTICATION=yes
 
   etcd_tests:
     build:
       dockerfile: docker/etcd_tests/Dockerfile
       context: .
     image: wal-g/etcd_tests
     container_name: wal-g_etcd_tests
-    depends_on:
```
chipitsine, line 799: it's good catch, and very small change. 

I'd suggest to move it to dedicated PR for the sake of simplicity

chipitsine, line 799: also, as long as etcd_tests container does not depend on "etcd" container, likely we can remove "etcd" completely from compose

serprex: folded etcd into etcd_tests

serprex: this all builds on `base` somewhat, are you suggesting `base` as its own PR & etcd stuff as followup?

with your other feedback applied feel like this PR is already pretty simple, if a bit scattered

chipitsine: I was wrong. ignore my suggestion

.github/workflows/dockertests.yml
```diff
@@ -12,6 +12,7 @@ env:
   CACHE_FILE_UBUNTU_18_04: ~/docker-images-${{ github.sha }}/wal-g_ubuntu-18.04.tar
   CACHE_FILE_UBUNTU_22_04: ~/docker-images-${{ github.sha }}/wal-g_ubuntu-22.04.tar
   CACHE_FILE_GOLANG: ~/docker-images-${{ github.sha }}/wal-g_golang.tar
+  CACHE_FILE_BASE: ~/docker-images-${{ github.sha }}/wal-g_base.tar
```
chipitsine: those images are built on the very first job. 
other jobs are waiting.
despite "cache" is used, them are not cached, they are used only 1 time, no other way, cache key is commit hash.
after that image(s) are compressed, uploaded to github, downloaded and decompressed.

I really doubt we shall introduce more images to this bottleneck.

chipitsine: at least we shall compare overall timing before and after

serprex: in other PR my goal is to replace other cache files here with base, however if you'd prefer I can have this PR use an existing base for now

chipitsine (approved):

PR #2467, opened 2026-08-03
Postgres & Greenplum journal counting

ostinru:
This PR adds journal counting for Greenplum/Cloudberry and Postgres. For Greenplum/Cloudberry shared part of backup (AO/AOCS/PAX) also counted.

Desing decisions:
1. WAL is counted in the same way as for MySQL
2. Shared Part counted in simplified way - for first version count only bytes uploaded during backup. No recalculation when older backups deleted.


ostinru:
based on #1806 and #1874

ostinru:
Split this PR into:
https://github.com/wal-g/wal-g/pull/2500 (journal counting)
https://github.com/wal-g/wal-g/pull/2507 (shared size)

PR #2475, opened 2026-08-06
feat: mariadb gtid support

Dmytro-Opria:
### Database name
MySQL / MariaDB

# Pull request description

`binlog-push`'s GTID-based checkpoint filtering (`WALG_MYSQL_CHECK_GTIDS`) currently only supports the MySQL flavor. Enabling it against a MariaDB server hits the `default: tracelog.ErrorLogger.Fatalf("Unsupported flavor type...")` branch and fails outright, because MariaDB uses a different GTID representation than MySQL (`domain-server-sequence`, e.g. `0-1-1011`, vs MySQL's `uuid:interval` format), and MariaDB GTIDs behave as monotonic high-water marks per (domain, server) rather than interval sets.

This PR adds a `mariadbGtidFilter` implementing the existing `binlogGtidFilter` interface (`isValid`, `shouldUpload`, `getArchivedGTIDString`), and wires it into `HandleBinlogPush` via a new `case mysql.MariaDBFlavor:` branch, so GTID-based deduplication of already-archived binlogs now works for MariaDB the same way it already does for MySQL.

**Files changed:**
- `internal/databases/mysql/mariadb_gtid_filter.go` (new) — MariaDB GTID
  checkpoint filter, using `MARIADB_GTID_LIST_EVENT`-based end-state
  comparison to decide whether a binlog is already archived.
- `internal/databases/mysql/mariadb_gtid_filter_test.go` (new) — unit tests
  covering `isValid`, GTID parsing, `Contain`, `Clone`, and `Update`
  semantics for MariaDB GTID sets.
- `internal/databases/mysql/binlog_push_handler.go` — add the MariaDB case
  to the flavor switch; add `getArchivedGTIDString()` to the existing
  MySQL `gtidFilter` so both filters satisfy the same interface.

No behavior change for existing MySQL users.

Not a bug fix — new feature. To exercise it: run `binlog-push` against a
MariaDB server with `WALG_MYSQL_CHECK_GTIDS=true`.

N/A — unit tests included in this PR (`mariadb_gtid_filter_test.go`).

**Build verification:**
`go build ./internal/databases/mysql/...` — build successful, no errors.
`go test ./internal/databases/mysql/...` — all tests pass.

### DIFF
--- a/internal/databases/mysql/binlog_push_handler.go
+++ b/internal/databases/mysql/binlog_push_handler.go
@@ -24,9 +24,15 @@ type LogsCache struct {
 	LastArchivedBinlog string `json:"LastArchivedBinlog"`
 }
 
+// binlogGtidFilter decides, based on a database's GTID state, whether a binlog
+// needs to be uploaded again and which GTID set has been archived so far.
+// Implementations exist per MySQL flavor (mysqlGtidFilter, mariadbGtidFilter).
 type binlogGtidFilter interface {
+	// isValid returns true if the filter is properly configured for the current flavor.
 	isValid() bool
+	// shouldUpload determines if a binlog should be uploaded based on GTID checkpoint comparison.
 	shouldUpload(binlog, nextBinlog string) bool
+	// getArchivedGTIDString returns the string representation of the archived GTID checkpoint for the current flavor.
 	getArchivedGTIDString() string
 }
 
@@ -73,17 +79,24 @@ func HandleBinlogPush(ctx context.Context, uploader internal.Uploader, untilBinl
 
 		switch flavor {
 		case mysql.MySQLFlavor:
+			var gtidArchived *mysql.MysqlGTIDSet
 			gtid, err := mysql.ParseMysqlGTIDSet(binlogSentinelDto.GTIDArchived)
 			if err != nil {
-				tracelog.ErrorLogger.Fatalf("Failed to parse MySQL GTID set '%s': %v", binlogSentinelDto.GTIDArchived, err)
+				tracelog.WarningLogger.Printf(
+					"Failed to parse MySQL GTID set '%s': %v. Uploading all binlogs and rebuilding the archived GTID checkpoint from scratch.",
+					binlogSentinelDto.GTIDArchived, err)
+			} else {
+				var ok bool
+				gtidArchived, ok = gtid.(*mysql.MysqlGTIDSet)
+				if !ok {
+					tracelog.WarningLogger.Printf(
+						"Failed to convert GTID to MysqlGTIDSet type. Uploading all binlogs and rebuilding the archived GTID checkpoint from scratch.")
+					gtidArchived = nil
+				}
 			}
-			gtidArchived, ok := gtid.(*mysql.MysqlGTIDSet)
-			if !ok {
-				tracelog.ErrorLogger.Fatalf("Failed to convert GTID to MysqlGTIDSet type")
-			}
-			filter = &gtidFilter{
-				BinlogsFolder: binlogsFolder,
-				Flavor:        flavor,
+			filter = &mysqlGtidFilter{
+				binlogsFolder: binlogsFolder,
+				flavor:        flavor,
 				gtidArchived:  gtidArchived,
 				lastGtidSeen:  nil,
 			}
@@ -92,17 +105,22 @@ func HandleBinlogPush(ctx context.Context, uploader internal.Uploader, untilBinl
 			if binlogSentinelDto.GTIDArchived != "" {
 				gtid, err := mysql.ParseMariadbGTIDSet(binlogSentinelDto.GTIDArchived)
 				if err != nil {
-					tracelog.ErrorLogger.Fatalf("Failed to parse MariaDB GTID set '%s': %v", binlogSentinelDto.GTIDArchived, err)
-				}
-				var ok bool
-				mariadbGTID, ok = gtid.(*mysql.MariadbGTIDSet)
-				if !ok {
-					tracelog.ErrorLogger.Fatalf("Failed to convert GTID to MariadbGTIDSet type")
+					tracelog.WarningLogger.Printf(
+						"Failed to parse MariaDB GTID set '%s': %v. Uploading all binlogs and rebuilding the archived GTID checkpoint from scratch.",
+						binlogSentinelDto.GTIDArchived, err)
+				} else {
+					var ok bool
+					mariadbGTID, ok = gtid.(*mysql.MariadbGTIDSet)
+					if !ok {
+						tracelog.WarningLogger.Printf(
+							"Failed to convert GTID to MariadbGTIDSet type. Uploading all binlogs and rebuilding the archived GTID checkpoint from scratch.")
+						mariadbGTID = nil
+					}
 				}
 			}
 			filter = &mariadbGtidFilter{
-				BinlogsFolder: binlogsFolder,
-				Flavor:        flavor,
+				binlogsFolder: binlogsFolder,
+				flavor:        flavor,
 				gtidArchived:  mariadbGTID,
 			}
 			tracelog.InfoLogger.Printf("Using MariaDB GTID filter for binlog push")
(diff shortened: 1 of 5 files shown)


ostinru:
Hi @Dmytro-Opria ! Thank you for your PR. I will be able to thoughtfully review this PR next week.

The most interesting part of uploading binlogs is how to download and apply them to MariaDB to provide Point in Time Recovery (PiTR). MySQL PiTR in wal-g heavily relies on GTID-sets and assumption that binlogs should be applied in order of uploading. In my experiments[1] MariaDB failed to apply binlogs when there are duplicates without explicity binlog name and position specification. So, I explicitly disabled PiTR for MariaDB in #1313 just to make it safe and don't make promises we can't keep.

@Dmytro-Opria, do you have clear picture of how to make PiTR works for MariaDB?

[1] https://github.com/wal-g/wal-g/pull/1313#issuecomment-1195932637

Dmytro-Opria:
> Hi @Dmytro-Opria ! Thank you for your PR. I will be able to thoughtfully review this PR next week.
> 
> The most interesting part of uploading binlogs is how to download and apply them to MariaDB to provide Point in Time Recovery (PiTR). MySQL PiTR in wal-g heavily relies on GTID-sets and assumption that binlogs should be applied in order of uploading. In my experiments[1] MariaDB failed to apply binlogs when there are duplicates without explicity binlog name and position specification. So, I explicitly disabled PiTR for MariaDB in #1313 just to make it safe and don't make promises we can't keep.
> 
> @Dmytro-Opria, do you have clear picture of how to make PiTR works for MariaDB?
> 
> [1] [#1313 (comment)](https://github.com/wal-g/wal-g/pull/1313#issuecomment-1195932637)

Hi @ostinru  — thanks for the thoughtful question, and for the context on #1313.

Quick introduction: I'm from STACKIT, and we run wal-g in production as our backup/PITR solution for MariaDB. Since our setup depends on wal-g directly, we'd like to help bring MariaDB PITR support up to full parity with MySQL upstream, rather than carrying fixes only in our internal fork.

This PR is the first in a small series we're planning — GTID-based deduplication for `binlog-push` here, with more to follow covering the rest of the MariaDB binlog-replay path.

On your specific concern: you're right to be cautious about this. We're actively testing our replay logic against real MariaDB instances on our side before proposing further changes, precisely because of the duplicate- application problem you ran into with #1313. We don't want to repeat that mistake or make promises the implementation can't keep.

That said, we don't assume we've caught everything — if you or anyone else spots gaps or edge cases while reviewing, we'd genuinely welcome that. Given the history here, your scrutiny is exactly what this needs.

internal/databases/mysql/binlog_push_handler.go
```diff
@@ -246,6 +277,9 @@ type gtidFilter struct {
 }
```
ostinru: Let's rename this to `mysqlGtidFilter` and move to separate file to keep it consistent with a new code.

internal/databases/mysql/binlog_push_handler.go
```diff
@@ -24,6 +24,12 @@ type LogsCache struct {
 	LastArchivedBinlog string `json:"LastArchivedBinlog"`
 }
 
+type binlogGtidFilter interface {
+	isValid() bool
+	shouldUpload(binlog, nextBinlog string) bool
+	getArchivedGTIDString() string
+}
```
ostinru, line 37: Now we have more than one implementation. Let's document contract for these methods (just copy/paste first line of comments you added for mariadbGtidFilter's methods)

internal/databases/mysql/binlog_push_handler.go
```diff
@@ -60,21 +66,46 @@ func HandleBinlogPush(ctx context.Context, uploader internal.Uploader, untilBinl
...
-			gtidArchived, _ := gtid.(*mysql.MysqlGTIDSet)
-			filter = gtidFilter{
+			gtid, err := mysql.ParseMysqlGTIDSet(binlogSentinelDto.GTIDArchived)
+			if err != nil {
+				tracelog.ErrorLogger.Fatalf("Failed to parse MySQL GTID set '%s': %v", binlogSentinelDto.GTIDArchived, err)
+			}
+			gtidArchived, ok := gtid.(*mysql.MysqlGTIDSet)
+			if !ok {
+				tracelog.ErrorLogger.Fatalf("Failed to convert GTID to MysqlGTIDSet type")
+			}
```
ostinru, line 96: For MySQL it is always safe to upload binlog. And there is a risk of loosing data if we decided not to upload.

I don't have clear example when `GTIDArchived` is not possible to parse... However I would prefer not to halt wal-g in this case. Just upload all binlogs & upload new binlog sentinel.

Let's replace `tracelog.ErrorLogger.Fatalf` with `tracelog.WarningLogger.Printf`

internal/databases/mysql/mariadb_gtid_filter.go
```diff
@@ -0,0 +1,128 @@
...
+// instead of MySQL's uuid:interval format (e.g., "3E11FA47-71CA-11E1-9E33-C80AA9429562:1-5").
+//
+// MariaDB GTIDs behave as monotonic high-water marks per (domain, server), not as
+// interval sets like MySQL. The sequence number for each domain only increases.
+//
+// This filter uses GTID checkpoints (end-state markers) to determine whether a binlog
+// has already been archived, avoiding unnecessary uploads.
+type mariadbGtidFilter struct {
+	BinlogsFolder string
+	Flavor        string
```
ostinru: Since we are moving from `gtidFilter` as struct to `binlogGtidFilter` as interface - let's make these fields private.

internal/databases/mysql/binlog_push_handler.go
ostinru: It's up to you to define default policy for MariaDB.

ostinru:
> On your specific concern: you're right to be cautious about this. We're actively testing our replay logic against real MariaDB instances on our side before proposing further changes, precisely because of the duplicate- application problem you ran into with https://github.com/wal-g/wal-g/pull/1313.

Hi, @Dmytro-Opria ! Which MariaDB version do you use in production? Currently our e2e tests uses mariadb-10.11 (released in 2023). It seems that I tested #1313 with 10.3. MariaDB users says that replication significantly improved in 10.8... So, there is a chance that this will work!

Dmytro-Opria:
> > On your specific concern: you're right to be cautious about this. We're actively testing our replay logic against real MariaDB instances on our side before proposing further changes, precisely because of the duplicate- application problem you ran into with #1313.
> 
> Hi, @Dmytro-Opria ! Which MariaDB version do you use in production? Currently our e2e tests uses mariadb-10.11 (released in 2023). It seems that I tested #1313 with 10.3. MariaDB users says that replication significantly improved in 10.8... So, there is a chance that this will work!

 10.11 is the minimum MariaDB version we plan to run in production, so that lines up well with your e2e test baseline.

internal/databases/mysql/binlog_push_handler.go
Dmytro-Opria: Done — split it into mysql_gtid_filter.go as mysqlGtidFilter, same layout as mariadbGtidFilter.

internal/databases/mysql/binlog_push_handler.go
Dmytro-Opria: Added — copied over the doc lines from mariadbGtidFilter's isValid/shouldUpload/getArchivedGTIDString.

internal/databases/mysql/binlog_push_handler.go
Dmytro-Opria: Switched both flavors to `WarningLogger` — on a parse failure we now fall back to an empty archived GTID set and rebuild it from scratch (upload everything) instead of halting. Went with the same policy for MariaDB, for consistency — let me know if you'd rather it behave differently there.

internal/databases/mysql/mariadb_gtid_filter.go
Dmytro-Opria: Done — `binlogsFolder/flavor` are private now.

ostinru (approved):
LGTM.

@teem0n please also look into this

ostinru:
@Dmytro-Opria, looks good. Usually we put PRs on hold for couple days to let anyone involved to review it as well.
If I will not merge after 2 days - please ping me.

ostinru:
@Dmytro-Opria test `binlog_push_with_gtids_check` failed:
```
wal-g_mariadb_tests  | WALG_MYSQL_CHECK_GTIDS is broken for MariaDB. It shouldn't allow users to misconfigure wal-g
```

It was added in #1313. Could be removed in our case.

Dmytro-Opria:
> @Dmytro-Opria test `binlog_push_with_gtids_check` failed:
> 
> ```
> wal-g_mariadb_tests  | WALG_MYSQL_CHECK_GTIDS is broken for MariaDB. It shouldn't allow users to misconfigure wal-g
> ```
> 
> It was added in #1313. Could be removed in our case.

Done: removed

chipitsine:
was https://github.com/wal-g/wal-g/pull/2162 superseded by this PR ?
shall we close it ?

ostinru:
> was #2162 superseded by this PR ? shall we close it ?

Yes

PR #2332, opened 2026-06-10
base image based on debian bullseye

serprex:
consolidating CI images to reduce cache footprint / update old images

drops testing for mysql 5

drop usage of bitnamilegacy etcd, bitnamilegacy images aren't updated

cuts down docker-images by 400MB in cache

PR #2486, opened 2026-08-17
Replace wget with curl

serprex:
split out from #2332


docker/pg_tests/Dockerfile_pgbackrest
```diff
@@ -1,10 +1,10 @@
 FROM ubuntu:18.04 as build
 
-RUN apt-get update && apt-get install -y wget
+RUN apt-get update && apt-get install -y curl
```
ostinru, line 3: Isn't `curl` part of standard Ubuntu image?

serprex: No

chipitsine (approved):

PR #2497, opened 2026-08-19
mysql: track binlog position/GTID from backup and skip already-applied binlogs on PITR replay

Dmytro-Opria:
# Description

## Summary

Store exact binlog file, position, and last GTID from xtrabackup_info / mariadb_backup_info in the backup sentinel during backup-push, and use it during binlog-replay to skip already-restored binlogs and resume from the exact position — including across a failover, where restore replays binlogs from a server other than the one the backup was taken from.

- Parse binlog_pos from xtrabackup_info (xtrabackup) and mariadb_backup_binlog_info (mariabackup); add BinLogFileName, BinLogFilePosition, BinLogLastGTID to StreamSentinelDto
- Pass WALG_MYSQL_BINLOG_START_POSITION / WALG_MYSQL_BINLOG_LAST_GTID to the replay command
- Skip a binlog whose GTID content is already covered by the backup's checkpoint, even when file numbering has no relation to the boundary file
- Fix two more gaps found testing against a real ha-agent failover: a file-number pre-filter that bypassed the GTID check and could drop new writes, and `--start-position` trusted on filename match alone, unsafe when a different server produces a same-named file
- Docker integration tests (pitr_mariabackup_full_test.sh, pitr_failover_gtid_skip_test.sh, pitr_failover_low_file_number_test.sh, pitr_failover_filename_collision_test.sh) plus unit tests for the skip/replay logic

## Database name
MySQL / MariaDB

## What this PR fixes

During binlog-replay for point-in-time recovery, WAL-G replayed binlogs starting from the backup's file-number boundary alone. Since the backup already contains data up to a specific position within a later binlog, blindly replaying past that boundary re-applies DDL/DML that already exists, causing "Table already exists" and similar errors.

This gets worse across a failover: restore may fetch binlogs from a server other than the one the backup was taken from, whose file numbering has no relation to the backup's own boundary — a plain "skip everything before this file name" check can't tell whether such a binlog is already covered, and replaying it again fails or silently duplicates data.

Both xtrabackup and mariabackup record the exact binlog file and position in their info file, and MariaDB's GTID is available from the same source. This PR stores that information in the backup sentinel, resumes replay from the exact position, and — for binlogs file numbering can't classify — checks GTID containment against the backup's checkpoint, replaying only what's genuinely new.

Related: #2195 (continues the binlog position tracking started there, addressing the failover concern raised in review)

See also: #2475 — earlier MariaDB GTID work for `binlog-push`; this PR covers the corresponding `binlog-replay` side.


chipitsine:
there's another PITR activity  https://github.com/wal-g/wal-g/pull/2195/changes
can you please have a look

Dmytro-Opria:
> there's another PITR activity https://github.com/wal-g/wal-g/pull/2195/changes
> can you please have a look

This is a continuation of #2195, which Santiago originally opened — I'm picking it up now. Same approach (binlog position/GTID tracking from the backup sentinel), and this PR also addresses the review comment left on #2195 about the failover case (GTID-skip when binlog file numbering doesn't line up with the backup's boundary file after a switchover).

dvcer (changes requested):
Hi, thank you for working on this.

I'd suggest splitting this PR into two and doing it one by one: (1) GTID-based skip filter, (2) binlog file position application. This makes each easier to review/test. For the current version, I'd like to suggest the following:

- Move the GTID-based skip logic into fetchLogs(). It's currently in replayHandler.replayLogs(), which only the replay path uses. fetchLogs is shared by all binlog stream consumers (fetch, binlog-server, replay), so filtering by GTID will be available for all paths.

- Binlog numbers aren't comparable across servers. The pre-filter at internal/databases/mysql/binlog_replay_handler.go:89 compares binlog file numbers, but on HA clusters S3 can hold binlogs from different instances whose numbering has no relation to each other (for instance, we can have mysql1.00001...mysql1.00005 binlogs pushed to S3, then after a switchover the next binlog file can be mysql2.00001). So this optimization seems unsafe and probably should be dropped.

- Let's not hardcode MariaDB handling directly in the shared replay/fetch logic — instead, branch on flavor, the same way it's done in binlog_push_handler.go.

Dmytro-Opria:
> Hi, thank you for working on this.
> 
> I'd suggest splitting this PR into two and doing it one by one: (1) GTID-based skip filter, (2) binlog file position application. This makes each easier to review/test. For the current version, I'd like to suggest the following:
> 
> * Move the GTID-based skip logic into fetchLogs(). It's currently in replayHandler.replayLogs(), which only the replay path uses. fetchLogs is shared by all binlog stream consumers (fetch, binlog-server, replay), so filtering by GTID will be available for all paths.
> * Binlog numbers aren't comparable across servers. The pre-filter at internal/databases/mysql/binlog_replay_handler.go:89 compares binlog file numbers, but on HA clusters S3 can hold binlogs from different instances whose numbering has no relation to each other (for instance, we can have mysql1.00001...mysql1.00005 binlogs pushed to S3, then after a switchover the next binlog file can be mysql2.00001). So this optimization seems unsafe and probably should be dropped.
> * Let's not hardcode MariaDB handling directly in the shared replay/fetch logic — instead, branch on flavor, the same way it's done in binlog_push_handler.go.

  Thanks for the review!                                                                                                                                                                                         
                                                                                                                                                                                                                 
  1. Splitting the PR — agreed, will split into position-tracking and GTID-skip.                                                                                                                                 
  2. Moving skip logic into fetchLogs() — fetchLogs is shared with binlog-fetch/binlog-server, neither of which has a notion of "this backup's GTID checkpoint." Did you mean pulling out just the containment check as a helper, or something broader?                                                                                                                                                                    
  3. Dropping the file-number pre-filter — agreed, good point. It only kicks in when there's no GTID checkpoint, but in a real HA cluster after failover, a low file number from the new node can still contain new data, and skipping it there is silent data loss during PITR, which is worse than the duplicate-key error we'd get by just replaying it. We'll pair this with 4 below — MySQL GTID is usually present in real HA setups even when the current code doesn't parse it, so extending GTID-skip to MySQL flavor should shrink the no-GTID fallback to genuinely GTID-less setups, which don't do automated failover anyway.                                                                                                                                                                                                     
  4. Flavor hardcoding — agreed, will follow the mysqlGtidFilter/mariadbGtidFilter pattern from binlog_push_handler.go and extend GTID-skip to also handle MySQL-flavor GTID, not just MariaDB.

dvcer:
> 2. Moving skip logic into fetchLogs() — fetchLogs is shared with binlog-fetch/binlog-server, neither of which has a notion of "this backup's GTID checkpoint." Did you mean pulling out just the containment check as a helper, or something broader?

Actually, binlog-server is just another implementation of binlog-replay. In binlog-server we have a starting point (the backup) and want to apply binlogs since then. So in binlog-server we can also apply a GTID-based filter to drop all binlogs that the backup already covers. This is why I think it would be useful to make this logic common to binlog-replay and binlog-server.

Speaking of how to implement it, maybe we can add a new GTIDSet argument to fetchLogs() and implement the filtering logic on the fetchLogs() side (what you currently do in replayLogs). For fetchLogs() consumers, it will be the same binlog stream, just without the binlog files that were dropped.

Or we can create a new entity — a GTID-based filter — and decorate fetchLogs() with it wherever we want. The filter will consume the stream of binlogs from fetchLogs, drop binlogs that are covered by the GTIDSet, and produce another binlog stream that we'll consume in binlog-replay/binlog-server.

What do you think?

Dmytro-Opria:
> Actually, binlog-server is just another implementation of binlog-replay. In binlog-server we have a starting point (the backup) and want to apply binlogs since then. So in binlog-server we can also apply a GTID-based filter to drop all binlogs that the backup already covers. This is why I think it would be useful to make this logic common to binlog-replay and binlog-server.
> 
> Speaking of how to implement it, maybe we can add a new GTIDSet argument to fetchLogs() and implement the filtering logic on the fetchLogs() side (what you currently do in replayLogs). For fetchLogs() consumers, it will be the same binlog stream, just without the binlog files that were dropped.
> 
> Or we can create a new entity — a GTID-based filter — and decorate fetchLogs() with it wherever we want. The filter will consume the stream of binlogs from fetchLogs, drop binlogs that are covered by the GTIDSet, and produce another binlog stream that we'll consume in binlog-replay/binlog-server.
> 
> What do you think?

Good point — Handler.requiredGTIDs in binlog-server is the same "already covered, drop it" concept, just sourced from COM_BINLOG_DUMP_GTID instead of the backup checkpoint. Today it's applied per-event after downloading and parsing the whole file; hoisting it to file-level in fetchLogs() means binlog-server could skip fully-covered files without downloading them at all, so this helps binlog-server too, not just replay.                                                                                                                                                                                                        
                                                                                                                                                                                                                 
We'd lean toward the decorator approach — a binlogHandler that wraps another one and does the lookahead/drop-or-forward logic — over adding a param to fetchLogs(), since binlog-fetch has no GTID concept and shouldn't need to care.
For the type, go-mysql's mysql.GTIDSet interface already has Contain() and ParseGTIDSet(flavor, s) parses either flavor generically, so the filter itself won't need MariaDB-specific  hardcoding.

One thing worth calling out separately: requiredGTIDs in binlog-server only handles MySQL right now (*mysql.MysqlGTIDSet) — MariaDB replicas don't negotiate GTID state through COM_BINLOG_DUMP_GTID at all, so there's nothing there yet to even plug a filter into on that side. Happy to generalize the filter itself as part of this, but actually wiring up MariaDB's side of that negotiation feels like its own separate chunk of work. 
Do you want us to fold that in here too, or keep it out of scope for now?

dvcer:
> > Actually, binlog-server is just another implementation of binlog-replay. In binlog-server we have a starting point (the backup) and want to apply binlogs since then. So in binlog-server we can also apply a GTID-based filter to drop all binlogs that the backup already covers. This is why I think it would be useful to make this logic common to binlog-replay and binlog-server.
> > Speaking of how to implement it, maybe we can add a new GTIDSet argument to fetchLogs() and implement the filtering logic on the fetchLogs() side (what you currently do in replayLogs). For fetchLogs() consumers, it will be the same binlog stream, just without the binlog files that were dropped.
> > Or we can create a new entity — a GTID-based filter — and decorate fetchLogs() with it wherever we want. The filter will consume the stream of binlogs from fetchLogs, drop binlogs that are covered by the GTIDSet, and produce another binlog stream that we'll consume in binlog-replay/binlog-server.
> > What do you think?
> 
> Good point — Handler.requiredGTIDs in binlog-server is the same "already covered, drop it" concept, just sourced from COM_BINLOG_DUMP_GTID instead of the backup checkpoint. Today it's applied per-event after downloading and parsing the whole file; hoisting it to file-level in fetchLogs() means binlog-server could skip fully-covered files without downloading them at all, so this helps binlog-server too, not just replay.
> 
> We'd lean toward the decorator approach — a binlogHandler that wraps another one and does the lookahead/drop-or-forward logic — over adding a param to fetchLogs(), since binlog-fetch has no GTID concept and shouldn't need to care. For the type, go-mysql's mysql.GTIDSet interface already has Contain() and ParseGTIDSet(flavor, s) parses either flavor generically, so the filter itself won't need MariaDB-specific hardcoding.
> 

Cool.

> One thing worth calling out separately: requiredGTIDs in binlog-server only handles MySQL right now (*mysql.MysqlGTIDSet) — MariaDB replicas don't negotiate GTID state through COM_BINLOG_DUMP_GTID at all, so there's nothing there yet to even plug a filter into on that side. Happy to generalize the filter itself as part of this, but actually wiring up MariaDB's side of that negotiation feels like its own separate chunk of work. Do you want us to fold that in here too, or keep it out of scope for now?

Yes, perhaps in this pr we could focus on implementing the filter and plug it in binlog-replay.

Dmytro-Opria:
Split this into #2527 (position/GTID tracking, merged) and #2528 (GTID-skip during replay). Closing this in favor of those two.

PR #2498, opened 2026-08-19
fix(mysql): guard binlog server disconnect cleanup against nil connection

chipitsine:
The MySQL binlog replication server could panic while tearing down a replica connection after a broken pipe / proxy disconnect. The close path called into the go-mysql server connection without checking whether the underlying packet connection had been initialized, which caused a nil-pointer dereference during reconnect churn.

addresses: https://github.com/wal-g/wal-g/actions/runs/32118514347/job/95655686325

### DIFF
--- a/internal/databases/mysql/binlog_server_handler.go
+++ b/internal/databases/mysql/binlog_server_handler.go
@@ -500,7 +500,12 @@ func handleBinlogConnection(
 		}
 		return
 	}
-	defer conn.Close()
+
+	defer func() {
+		if !conn.Closed() {
+			conn.Close()
+		}
+	}()
 
 	for {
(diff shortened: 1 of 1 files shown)


dvcer (changes requested):
Hi, thank you for the patch! It seems that the packet conn could be nil here only when the connection was already closed in HandleCommand. So maybe we can just wrap the defer close call with !conn.Closed(), like this:

```
defer func() {
    if !conn.Closed() {
        conn.Close()
    }
}()
```

chipitsine (commented):
applied

dvcer (approved):
LGTM, thanks!

teem0n (approved):

PR #2499, opened 2026-08-19
go 1.27

serprex:
no more GOEXPERIMENT=jsonv2

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -9,9 +9,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
-env:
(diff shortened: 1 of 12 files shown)


serprex:
https://github.com/golangci/golangci-lint/pull/6642

ostinru:
Previously, we switched golang version on 1.x.2 - just to let golang team fix any regressions in fresh releases.

However, I will agree that `GOEXPERIMENT=jsonv2` is a pain =(

chipitsine:
Shall we fire new release before ?

ostinru:
> Shall we fire new release before ?

I think, we should create new release.

chipitsine:
ok, new release emitted

serprex:
@ostinru could I get approval here? just some context cleanup in tests flagged code owners

ostinru (approved):

PR #2500, opened 2026-08-20
Postgres and Greenplum journal (WAL) counting

ostinru:
### Postgres and Greenplum

Common:
* Use `JournalInfo` in our code instead of `storage.Object` - in this way we do not depend on backup naming pattern (Postgres' naming doesn't contain timestamp).

For postgres:
* Add `count-journals` feature like MySQL has.

For greenplum/cloudberry:
  * Each segment's WAL-G instance counts the WAL it archives into its own `segments_005/seg<N>/wal_005` during `seg-backup-push`, and the coordinator sums the per-segment journals into a cluster-wide `journal_<backup>`.

### DIFF
--- a/docs/PostgreSQL.md
+++ b/docs/PostgreSQL.md
@@ -214,6 +214,18 @@ If a backup is started from a standby server, WAL-G will monitor the timeline of
 
 ``backup-push`` can also be run with the ``--permanent`` flag, which will mark the backup as permanent and prevent it from being removed when running ``delete``.
 
+#### Journal size (WAL volume) accounting
+
+Run ``backup-push`` with the ``--count-journals`` flag to maintain a ``journal_<backup>`` object per backup in storage, tracking the volume of WAL accumulated between that backup and the next one. The size is computed from the actual storage object sizes of the archived WAL segments, so it correctly reflects compression, unlike an LSN-difference estimate.
+
+```bash
+wal-g backup-push $PGDATA --count-journals
+```
+
+Journal accounting is skipped for permanent backups (marked with ``--permanent``), since they are not expected to be removed and don't take part in WAL retention planning.
+
+Currently, only ``delete target`` cleans up and re-merges the corresponding journal entry when a backup is removed this way; other ``delete`` modes (``before``, ``retain``, ``retain-after``, ``everything``, ``garbage``) leave existing journals untouched.
+
 #### Remote backup
 
 WAL-G backup-push allows for two data streaming options:
--- a/internal/journal.go
+++ b/internal/journal.go
@@ -17,10 +17,9 @@ import (
 )
 
 const (
-	JournalPrefix           = "journal_"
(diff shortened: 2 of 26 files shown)


internal/journal.go
```diff
@@ -54,15 +53,15 @@ type JournalInfo struct {
...
+	currentBackupEnd time.Time,
 	journalDir string,
 ) JournalInfo {
 	return JournalInfo{
 		JournalName:          fmt.Sprintf("%s%s", JournalPrefix, backupName),
 		JournalDirectoryName: journalDir,
-		PriorBackupEnd:       currentBackupEnd,
-		CurrentBackupEnd:     priorBackupEnd,
+		PriorBackupEnd:       priorBackupEnd,
+		CurrentBackupEnd:     currentBackupEnd,
```
ostinru, line 64: There were a bug - variables messed up twice (so, there were no actual issue here)

docs/PostgreSQL.md
```diff
@@ -214,6 +214,18 @@ If a backup is started from a standby server, WAL-G will monitor the timeline of
 
 ``backup-push`` can also be run with the ``--permanent`` flag, which will mark the backup as permanent and prevent it from being removed when running ``delete``.
 
+#### Journal size (WAL volume) accounting
+
+Run ``backup-push`` with the ``--count-journals`` flag to maintain a ``journal_<backup>`` object per backup in storage, tracking the volume of WAL accumulated between that backup and the next one. The size is computed from the actual storage object sizes of the archived WAL segments, so it correctly reflects compression, unlike an LSN-difference estimate.
```
debebantur, line 219: You mean, if we have backup A and then call backup B with --count-journals, it will count journals between A and B, or between B and the next backup?

internal/journal.go
```diff
@@ -319,31 +337,45 @@ func filterJournalsInfoFiles(objects []storage.Object) []storage.Object {
...
+	return journals, nil
 }
 
-func filterJournalsInfoNewerThen(objects []storage.Object, timestamp time.Time) []storage.Object {
-	newObjects := make([]storage.Object, 0, len(objects))
-	for _, obj := range objects {
-		objTimestamp := getJournalTimestamp(obj.GetName())
-		if objTimestamp.After(timestamp) {
-			newObjects = append(newObjects, obj)
+func filterJournalsInfoOlderThen(journals []JournalInfo, timestamp time.Time) []JournalInfo {
```
debebantur, line 356: Isn`t lsn more reliable? Or lsn will make this code too db-specific?

docs/PostgreSQL.md
ostinru: It will count WAL between A and B. Result will be written to file journal_A.json

Probabaly, there will be journal_B.json with 0 WAL size

internal/journal.go
ostinru: LSN is db-specific.

debebantur (approved):

PR #2504, opened 2026-08-22
test(pg): run the integration suite on PostgreSQL 10 and 14-18

### LINKED ISSUE #2252: [FEATURE] upgrade Docker integration tests from PostgreSQL 10 to PostgreSQL 14+

# Problem
The Docker integration tests are currently hardcoded to PostgreSQL 10, which has been end-of-life since November 2022. Testing against an unsupported version provides false confidence and may mask real bugs on supported versions.

I ran into this concretely while writing a new integration test that required pg_checksums — a tool that does not exist in PostgreSQL 10 but is available starting from PostgreSQL 12.

# Proposal
Replace the hardcoded PostgreSQL 10 image with PostgreSQL 14 at minimum, or ideally refactor the tests to be version-agnostic and cover PostgreSQL 14–18.

**I have already done this work once across ~58 files, but lost the changes when my laptop died :( . I am willing to do it again if the community agrees this is the right direction.**

- PostgreSQL 10 is EOL
- Several modern PostgreSQL features used in backup/restore workflows (e.g. pg_checksums) are unavailable on version 10
- Testing only on an EOL version means we may miss bugs that only surface on actively maintained PostgreSQL releases.

P.S. 
Happy to submit a PR if this gets a thumbs up.

boosterKRD:
### Database name

PostgreSQL.

# Pull request description

Closes #2252 — *[FEATURE] upgrade Docker integration tests from PostgreSQL 10 to PostgreSQL 14+*.
This PR touches **only the PostgreSQL tests**. The other engines are not changed.

Done as @chipitsine suggested in the issue — the tests run sequentially, on several versions. The suite now runs on **10 and 14-18**. The version list is one line in the Makefile (`PG_VERSIONS`, line 116), so adding PostgreSQL 19 is two lines.

### How a CI run works

- **Build images** — one job. It builds and caches one test image per version.
- **PostgreSQL** — one job per version, so the six versions run **in parallel** with each other.
- Inside each of those jobs the whole suite runs **sequentially**, in a single container: `run_integration_tests.sh` walks `tests/*.sh`, so a new test cannot be forgotten.
- **Other engines** — unchanged: one job per test, all in parallel.

Six versions cost more machine time, but almost no extra waiting, because PostgreSQL is the longest job anyway:

image

A full run on my fork: all green, 37 of 37 jobs, nothing re-run — [link](https://github.com/boosterKRD/wal-g/actions/runs/32579293018).

### Version differences

Version differences are wrapped in helpers in `test_functions/pg_compat.sh`, so each one is written once instead of being copied as an `if` into 20+ tests: recovery settings (PG 12), `wal_keep_size` (PG 13), `scram-sha-256` (PG 14), `pg_switch_wal()` (PG 10), `\restrict` markers in `pg_dump` (PG 18).

The same file has the guards that say which versions a test applies to: 
- `require_pg_ge <version>`
- `require_pg_le <version>`
- `skip_pg <version>`.

A test calls one as its first line and then exits with `77`, which the runner reports as `SKIPPED` — not passed and not failed. Today only the two `wale_*` tests need it (`require_pg_le 10`), because wal-e installs on python3.7 only.

### Removed

`dockertests-seq.yml` (the nightly run). It existed to catch tests that were missing from the hand-written list in `dockertests-par.yml`. That list is gone — every job now runs the whole suite — so IMHO the nightly run is not needed any more.

### For other open PRs

Open PRs that add or change PostgreSQL tests will need a small update after this is merged. For instance #2251 (mine) and #2293.


chipitsine:
I think, as long as postgresql image is used only once, no more need to build it in advance and pass later (that step is blocking and pretty much resource consuming).

more improvements to go :)

but definetly we can do it later

chipitsine (approved):

boosterKRD:
> I think, as long as postgresql image is used only once, no more need to build it in advance and pass later (that step is blocking and pretty much resource consuming).

Thanks, good spot on the PG image caching. It is absolutely pointless now. I can fix it in a separate PR (later)

docker-compose.bake.yml
```diff
@@ -54,26 +54,35 @@ services:
...
+  # These two services cover PostgreSQL 14-18 and build on ubuntu 22.04, not
+  # bionic. PG_MAJOR is part of the cache scope, so every version has its own
+  # cache; with one shared cache each version change would rebuild everything.
+  pg:
     build:
       additional_contexts:
         wal-g/ubuntu:22.04: service:ubuntu_22_04
       cache_from:
-        - type=gha,scope=walg-pg18
+        - type=gha,scope=walg-pg${PG_MAJOR:-18}
```
chipitsine: I'd prefer not to specify default value.

it introduces complexity and might lead to some errors.

boosterKRD: [fixed ](https://github.com/wal-g/wal-g/pull/2504/changes/a81ce7865405c9585727f757fb8d46bf2a84df5a)

chipitsine: I would merge it if nobody minds.

In theory it can introduce excessive load, in such case we can later limit matrix to 10 + 18

boosterKRD:
@chipitsine 
1. Conflict resolved.
2. Adapted `journals_test` from #2500 to the new layout
3. Local tests passed; now waiting for the CI run.

chipitsine:
merge 'em all, let god sort 'em out ?

Issue #2462, opened 2026-07-30
[FEATURE] MZRA: make zstandard reusable again


chipitsine:
### What feature would you like?

in short - we must enable "**reuse zstd encoders**" feature described [here](https://github.com/klauspost/compress/tree/master/zstd#zstd):

```
Especially when encoding blocks you should take special care to reuse the encoder. This will effectively make it run without allocations after a warmup period. To make it run completely without allocations, supply a destination buffer with space for all content.
```

precaution should be taken to overall memory consumption.

rationale.

database was created and populated using pgbench

```
createdb pgbench
pgbench -i -s 650 pgbench
```

config.json

```
{
"PGHOST": "localhost",
"PGPORT": 5432,
"PGUSER": "postgres",
"PGPASSWORD": "xxx",
"WALG_FILE_PREFIX": "/mnt/xxx",
"WALG_COMPRESSION_METHOD": "none",
"WALG_LIBSODIUM_KEY": "6610844b654ecf33731a9cf135d20500b6d51d5b9e87a91b0bcf3f7b83b10875",
"WALG_LIBSODIUM_KEY_TRANSFORM": "hex",
"WALG_COMPRESSION_METHOD": "zstd",
"WALG_ZSTD_LEVEL": "fastest"
}
```

wal-g was built `make STRIP_BINARIES=0 pg_build`

performance report collected

```
perf record -g wal-g backup-push /var/lib/pgsql/data
perf report --stdio > perf-report.txt
```

from aggregated report I can see that about 99% cpu cycles are spent on zstd.
no need to optimize any other subsystem.

full collected "perf.data" as well as exported report and useless aggregation from chatgpt are attached here as well

chipitsine:
[perf-report.txt](https://github.com/user-attachments/files/30538888/perf-report.txt)

chipitsine:
[perf.zip](https://github.com/user-attachments/files/30538926/perf.zip)

chipitsine:
# Performance analysis: Zstandard-compressed backup profile

This run is compression-bound. The bottleneck has shifted from encryption to Zstandard.

| Bottleneck | Evidence | Meaning |
| --- | ---: | --- |
| Zstd block encoding | `zstd.(*blockEnc).encode` **44.30% inclusive**, **21.52% self** | Compression is the primary CPU ceiling. |
[...]

PR #2508, opened 2026-08-25
perf(zstd): reuse encoders across writers

thaildhe172591:
### Database name
All databases (compression is database agnostic).

# Pull request description

Implements the encoder reuse asked for in #2462. Every tar part and WAL segment built a fresh zstd encoder, so the encoder's match tables and buffers were allocated once per uploaded member. Encoders now live in a sync.Pool per level and are Reset onto the next destination in NewWriter.

benchstat over 10 runs on my machine (i7-1355U, Windows), old vs new, one member per op:

```
                    sec/op (old)   sec/op (new)   vs base
Compressor/1MiB     12.875m ± 31%   4.784m ± 14%  -62.84% (p=0.000)
Compressor/16MiB     53.40m ± 61%   52.70m ± 27%        ~ (p=1.000)

                    B/op (old)     B/op (new)     vs base
Compressor/1MiB     18720.0Ki      937.2Ki        -94.99% (p=0.000)
Compressor/16MiB    18735.9Ki      954.0Ki        -94.91% (p=0.000)
```

To be upfront about what this does and does not do: it removes per-writer setup and allocation cost. It does not touch block encoding itself, so on a profile like the one in the issue, where blockEnc.encode dominates, elapsed time on large members stays within noise. The win is allocations and GC pressure, plus real time on small members. My laptop numbers are noisy; I would be glad to have this measured on your stand as offered in the issue.

On the memory precaution from #2429/#2430: sync.Pool is drained every GC cycle, so retained encoder state is bounded by the number of writers actually in flight, not by history. The wrapper also returns its encoder to the pool exactly once and refuses Write and Flush after Close with the encoder's own closed error; without that, a double Close through CascadeWriteCloser could hand one encoder to two writers and corrupt both streams. Encoders whose Close failed are dropped rather than recycled. Each of those guards has a test that fails when the guard is removed.

Not a bug. To compare: `go test -run XXX -bench BenchmarkCompressor -benchtime 20x -count 10 ./internal/compression/zstd/` on master and on this branch, then benchstat the two outputs.

Not applicable (no behavior change; output remains a standard single-frame zstd stream). Covered by unit tests in internal/compression/zstd/compressor_test.go, including double close, write after close, concurrent writers under the race detector, and per-level pool separation.

### DIFF
--- a/internal/compression/zstd/compressor.go
+++ b/internal/compression/zstd/compressor.go
@@ -2,6 +2,7 @@ package zstd
 
 import (
 	"io"
+	"sync"
 
 	"github.com/klauspost/compress/zstd"
 	"github.com/wal-g/wal-g/internal/ioextensions"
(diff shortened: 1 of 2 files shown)


chipitsine:
I measured `/usr/bin/time -v wal-g backup-push /var/lib/pgsql/data`  (10Gb of data)

walg.config
```
{
"PGHOST": "localhost",
"PGPORT": 5432,
"PGUSER": "postgres",
"PGPASSWORD": "xxx",
"WALG_COMPRESSION_METHOD": "zstd",
"WALG_ZSTD_LEVEL": "fastest",
"AWS_SECRET_ACCESS_KEY": "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY",
"AWS_ACCESS_KEY_ID": "AKIAIOSFODNN7EXAMPLE",
"AWS_ENDPOINT": "http://127.0.0.1:9000",
"AWS_S3_FORCE_PATH_STYLE": "true",
"WALE_S3_PREFIX": "s3://gppartialtablebucket",
"WALG_PGP_KEY_PATH": "/tmp/PGP_KEY",
"WALG_UPLOAD_CONCURRENCY": "4",
"WALG_UPLOAD_DISK_CONCURRENCY": "1",
"WALG_UPLOAD_QUEUE": "1",
"WALG_S3_MAX_PART_SIZE": "8388608"
}
```

before

```
        Command being timed: "wal-g backup-push /var/lib/pgsql/data"
        User time (seconds): 35.59
        System time (seconds): 7.39
        Percent of CPU this job got: 126%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:34.03
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 140700
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 5
        Minor (reclaiming a frame) page faults: 26997
        Voluntary context switches: 429898
        Involuntary context switches: 32153
        Swaps: 0
        File system inputs: 17657744
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

after

```
        Command being timed: "wal-g backup-push /var/lib/pgsql/data"
        User time (seconds): 33.47
        System time (seconds): 6.94
        Percent of CPU this job got: 141%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:28.61
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 137260
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 6
        Minor (reclaiming a frame) page faults: 25606
        Voluntary context switches: 419954
        Involuntary context switches: 28713
        Swaps: 0
        File system inputs: 17490840
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

3.35Mb of "Maximum resident set size" less

chipitsine (approved):

thaildhe172591:
Thanks for measuring it on real data. That is a much better signal than the microbenchmark I ran.

The 3.35Mb RSS drop is the part I expected. The wall clock going 34.03s to 28.61s is larger than I would have predicted, and I think your config explains why: with `WALG_S3_MAX_PART_SIZE` at 8388608, a 10Gb backup is on the order of a thousand tar members, so the old path was allocating and warming a zstd encoder's match tables about a thousand times. The pooled encoders turn that into a Reset per member, and at `WALG_ZSTD_LEVEL: fastest` the setup cost is a bigger share of the total than the compression itself.

That would mean the saving tracks the number of members rather than the byte count, so a run with a much larger part size should show the same RSS win and less of the time win. Worth knowing if anyone benchmarks this again with different settings.

chipitsine:
thanks for your contribution!

PR #2516, opened 2026-08-26
Replace minio with seaweedfs

secwall:


docker/s3_fault_proxy/envoy.yaml
```diff
@@ -0,0 +1,67 @@
...
+                    typed_config:
+                      "@type": type.googleapis.com/envoy.extensions.filters.http.fault.v3.HTTPFault
+                      max_active_faults: 1
+                      delay:
+                        fixed_delay: 1s
+                        percentage:
+                          numerator: 0
+                          denominator: HUNDRED
+                      abort:
+                        http_status: 503
```
debebantur, line 47: As far as i understood, this proxy sends 503 status to specified percent of requests. Can we make it send also 429, 499 and 409 codes? (We retry them too, so why not check them all)

secwall: Sure, np. I've added 409, 429 and 504 to fault injection.

debebantur (approved):
LGTM

NotAmigo:
LGTM

reshke (approved):

PR #2517, opened 2026-08-26
ci: do not cache pg_test images

chipitsine:
a followup of https://github.com/wal-g/wal-g/pull/2504
unfortunately, overall cache size was exhausted by increased amount of cached entities

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,28 +132,18 @@ pg_matrix_test:
 pg_save_image: install_and_build_pg pg10_build_image
 	mkdir -p ${CACHE_FOLDER}
 	sudo rm -rf ${CACHE_FOLDER}/*
-	for v in $(PG_VERSIONS); do \
-		if [ "$$v" != "10" ]; then $(MAKE) PG_MAJOR=$$v pg_build_image; fi; \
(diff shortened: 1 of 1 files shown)


chipitsine:
@boosterKRD 

there were several cache eviction (due to 1.3Gb single cache size)

image


I decided to force a little bit

serprex (approved):

boosterKRD:
> e were several cache eviction (due to 1.3Gb single c

My bad, I did not know the 10 GB cache limit is per repository, not per branch or PR.
It worked for me because I never ran more than one or two workflows at a time, so the cache was never under pressure.

The image caching should be dropped, as you proposed.

boosterKRD:
@chipitsine 

I am cooking a PR with cleanup of caching PG images (fix comments, remove pointless code from Makefile and workflow). Tomorrow I am going to publish it

chipitsine:
I think images wal-g/ubuntu:18.04 and wal-g/ubuntu:22.04 are pretty heavy (because of Ubuntu), on other hand them do not contain nothing build from repo, just Ubuntu + some packages.

What would be really nice

1) we can establish daily job which will build & push them to ghcr.io
2) from workflow we do not cache those 2 images. instead we try to pull from ghcr.io, if not then just build

PR #2523, opened 2026-08-28
fix(pg): re-read a page before reporting a checksum error

boosterKRD:
### Database name

PostgreSQL

# Pull request description

This is the second attempt at #2251. 

During `backup-push --verify` WAL-G validates page checksums as it reads data files. When PostgreSQL writes an 8KB page, the OS may split the write into two 4KB operations. If WAL-G reads the page in between, it gets a torn page and the checksum does not match. The page is reported as corrupt, but nothing is wrong with it: the WAL holds a full page image, and recovery replays it over the torn page.

### What this PR does

On a checksum mismatch the block is read from disk once more and compared with what went into the backup stream:

- **the block changed** — PostgreSQL is writing it right now, so a FPI is in the WAL. Not reported.
- **the block is unchanged** — nothing is writing it, so the bad checksum is real. Reported as before.
- **empty read or missing file** — the relation was truncated or dropped while we worked. The page is gone and cannot be corrupt.
- **any other I/O error** — the original verdict stands. A real corruption must not hide behind an error that stopped us from checking.

The re-read happens only when a checksum fails, so on a healthy cluster this costs nothing.

### Why not compare the page LSN, as #2251 did

The first version skipped every page with `LSN > backupStartLSN` when `full_page_writes` was on. Two problems killed it.

**It trusts a page that has already failed its own integrity check.** The LSN is read from the header of the very page we suspect is damaged. Garbage in `pd_lsn` reads as a recent LSN and quietly takes the page out of validation.

**The `FPI` check was wrong in both directions.** `pg_backup_start` forces full page writes for the whole backup regardless of the GUC, while `SHOW full_page_writes` keeps returning `off`. So on a primary with the parameter off the skip never fired — exactly where it was needed. And on a standby the local value says nothing at all, because the WAL is produced by the node being replayed.

Re-reading makes no assumptions: it observes whether the page is moving.

### A note on standby backups

The full page image is guaranteed on a primary. When backing up a standby, the WAL comes from the node it replays, so **FPI must be enabled there** — this is a documented requirement of base backups from a standby and is up to the user. 

### A note on Windows

The commit [bb742c2](https://github.com/wal-g/wal-g/pull/2523/commits/bb742c2cd8cf43e52bc496584057ec97177d3146) makes the new unit test build its file path with a forward slash. Without it the
test panics on the `windows build smoke test` job, and that may be masking something worth a separate look: `GetRelFileIDFrom` parses the path with `path.Base`, which does not understand the Windows separator.  So `backup-push --verify` may already be affected on Windows. **I have not looked into it further.**


boosterKRD:
@chipitsine, could you review please? 
The first four commits are the torn page change itself; the last one
is a Windows fix that my tests (generated by Claude) uncovered by accident, and I would like a second opinion on whether it belongs here at all.

On Windows `GetRelFileIDFrom` crashes on any relation path, so `backup-push --verify` looks broken there. It parses filesystem paths as if they were slash-separated (this is not something my change introduced). 

The commit makes it handle both separators and return an error instead of panicking. If PostgreSQL
backups on Windows are not something wal-g supports anymore, I am happy to drop that commit.

@joekiller you added the Windows job in f81943e6 commit at the beginning of 2026, so you probably have the best view on how far Windows support
is meant to go here; does this look worth fixing?

chipitsine:
I'll check

chipitsine:
I use it on windows, but I do not do --verify ))

boosterKRD:
@chipitsine  
I just realised that I have оne blind spot I should flag: a page that is both corrupt and rewritten constantly changes between the two reads every time, so it is never reported **(unlikely in practice, but possible in this universe in general)** . It only bites with `full_page_writes`= off.
On ZFS, say, where copy-on-write makes torn writes impossible (with proper config), and people do turn it off (at least I do).

Would it be worth adding a setting to turn the re-read off, on by default? Then these setups would see the corruption reported at once. I did not add it here, because the value would have to be passed down through the verifier, and I would rather ask for your opinion first.

chipitsine:
I'm not sure when it should fail on windows

I ran 

```
c:\i\wal-bin\wal-g.exe --config c:\i\wal-bin\config.yaml backup-push "C:\Program Files\PostgreSQL\18\data"  --verify
c:\i\wal-bin\wal-g.exe --config c:\i\wal-bin\config.yaml delete retain FULL 3 --confirm
```

either with or without the last commit. 
no fail

boosterKRD:
@chipitsine 
Yes, I rechecked :))) higher up in the [code](https://github.com/wal-g/wal-g/blob/f69c4bd29cb362dc7ce8d6a1f74a547e14653b0c/internal/databases/postgres/bundle.go#L185) the backslashes are replaced with slashes, so it never fails there. 
`filepath.ToSlash(path)`

The failure was only in my test, because I called that function directly (and I took it from ours code).

Well, I don't know — maybe just keep my "fix" or just remove it?

chipitsine:
Let's remove it then

joekiller:
@chipitsine @boosterKRD during the windows updates I found that the program needed to use ToSlash on any incoming path from user land to ensure that whatever form they provided that internally the program used / and then only when the program again needed to interact with user land did an os filepath get warranted. So filepath join internally is bad, path.join good. Userland tests that asset paths need to first be corrected to first filepath ToSlash and then filepath joined to keep consistent assertions around path evals.

Like in 
https://github.com/wal-g/wal-g/blame/f81943e64bdf97aa66f6c52fec55114703f97af7/internal/configure_test.go#L89

chipitsine:
@boosterKRD  in theory we can postpone cleanup for the later PR. nothing urgent

boosterKRD:
> Let's remove it then

Removed

boosterKRD:
@chipitsine, all the changes are in; could you take another look?

- The Windows "fix" is gone, as agreed. The unit test now builds the relation path with filepath.ToSlash, the way the walker does (thanks @joekiller for the pointer).
- Added WALG_VERIFY_PAGE_CHECKSUMS_RETRY (default true) to turn the re-read off for the full_page_writes = off setups I mentioned above. Documented in PostgreSQL.md.
- The Corruption found warning is now logged only after the re-read decision, so a torn page no longer produces a misleading warning in the log.
- Last commit is a small refactoring: GetRelFileIDFrom is now called once per file instead of once per page.

chipitsine:
works on windows with and without `--verify`
not sure how to emulate storage failure

chipitsine:
I tested on ntfs. Will test on ReFS tomorrow

PR #2548, opened 2026-09-10
Greenplum: reassign shared storage after backup deletion

ostinru:

Greenplum AO/AOCS and PAX objects are stored in shared storage and may be referenced by multiple backups. Previously, SharedSize was recorded only when a backup was created. Deleting an older backup therefore removed its accounted bytes even when surviving backups still referenced the corresponding objects.

This change reassigns shared-storage ownership after backup deletion:

- the oldest surviving backup owns all shared objects it references;
- every following backup owns objects absent from the immediately preceding surviving backup;
- AO/AOCS size is calculated from EOF in AO metadata;
- PAX metadata now records the original file Size, which is used for accounting;
- segment UploadedSharedSize values and cluster-level SharedSize objects are recalculated;
- dry runs do not update metadata;
- recalculation replaces values and is idempotent.

PAX files are uploaded without compression. Encryption overhead is intentionally ignored for this accounting.


internal/databases/greenplum/shared_size.go
```diff
@@ -28,67 +28,70 @@ type SharedSizeDTO struct {
...
-//
-// Both live under basebackups_005/<backup>/, which is what makes the cluster-level objects
-// disappear together with the backup: the path reduces to the backup name exactly like the sentinel
-// does, so every delete mode picks them up without any code of its own (see
-// utility.StripLeftmostBackupName).
-type sharedStorageKind struct {
-	name            string
-	path            func(backupName string) string
-	readSegmentSize func(ctx context.Context, baseBackupsFolder storage.Folder, backupName string) (int64, error)
-}
```
ostinru, line 43: I have removed this stucture - it leads to multiple levels of indirections and callbacks. Inline all functions. Use simple code rather than abstractions.

Code is not as clean as it can be... But it is easy to read and understand. I think this is go-way.
https://overreacted.io/goodbye-clean-code/

internal/databases/greenplum/shared_size.go
```diff
@@ -114,29 +116,19 @@ func sumOverSegments(ctx context.Context, rootFolder storage.Folder, backupName
...
-// learn the uploaded volume. The file lists those objects also carry are skipped: only the segments
-// have any use for them.
-type aoUploadedSizeView struct {
-	UploadedSharedSize int64
-}
-
-type paxUploadedSizeView struct {
-	UploadedSharedSize int64
-}
-
```
ostinru, line 138: hide these types inside functions. They are used only in these functions.

internal/databases/greenplum/tars_to_extract_provider.go
```diff
@@ -56,13 +56,14 @@ func (t FilesToExtractProviderImpl) Get(ctx context.Context, backup SegBackup, f
 		tracelog.DebugLogger.Printf("PAX files metadata was not found. Skipping PAX file unpacking.")
 	} else {
 		tracelog.InfoLogger.Printf("PAX files metadata found. Will perform PAX file unpacking.")
-		for extractPath, meta := range paxMeta.Files {
+		for extractPath := range paxMeta.Files {
```
ostinru, line 59: make linter happy:
```
 rangeValCopy: each iteration copies 128 bytes (consider pointers or indexing)
```

debebantur (approved):
LGTM

PR #2573, opened 2026-09-18
Reload WALG_DISK_RATE_LIMIT from config file during backup-push

InakiAtDatadog:
Operators previously had to restart a long-running backup-push to change disk I/O throttling, since WALG_DISK_RATE_LIMIT was read once at startup and baked into a fixed `rate.Limiter`. Add a private viper watcher that re-reads the setting from the config file and updates the limiter in place, guarded by a mutex against concurrent readers.

The watcher is skipped in Turbo mode, when the value came from a real environment variable (unobservable by a running process), or when no config file is in use. Remote/streaming backups are unaffected since the rate is handed to Postgres's BASE_BACKUP protocol once at stream start.

### Database name
Postgresql

# Pull request description

When backups run they are often competing for disk with the live work the cluster needs to do (serve customer queries on a leader or applying WAL on a replica). Live work is usually the priority and also often irregular. This means that a fixed disk rate limit for the entire backup, which is a long running task, results in either:
- disk contention if rate limit is too high, which can impact live traffic or at least cause the replica to fall behind
- underused disk if rate limit is too low, which results in a slower backup

N/A

# Validation

Testing captured [here](https://datadoghq.atlassian.net/wiki/x/1oIwrgE).


chipitsine:
just curious. wouldn't be easier to use `ionice`
(well, it's linux only...)

InakiAtDatadog:
> just curious. wouldn't be easier to use `ionice` (well, it's linux only...)

@chipitsine It's a good question (thanks) Using `ionice` was the original idea but in our current setup we don't really have access to an I/O scheduler that supports it.

PR #2576, opened 2026-09-20
MySQL binlog-server heartbeat support

dvcer:
###
1. Add heartbeat support to binlog-server. Heartbeats prevent replica reconnects during long idle periods, such as downloading binlogs from S3 or waiting for the replica to apply all changes. They also help detect closed connections earlier. Only heartbeat V2 is supported.

2. Rework shutdown and concurrency in binlog-server. The server handles one connection at a time and waits for the previous handler to finish cleanup before processing the next connection. This prevents streaming and downloads from overlapping after reconnects.

3. Remove transaction-skipping logic in binlog-server. WAL-G now forwards already-applied transactions and relies on MySQL’s GTID deduplication. This reduces how much the binlog server needs to understand about transaction boundaries and event types, simplifying streaming and reducing the risk of bugs as the binlog format evolves. The tradeoff is sending some additional events, whose overhead is expected to be small.


teem0n (approved):

internal/databases/mysql/binlog_server_handler.go
ostinru, line 152: You have connection to database. I think `WALG_MYSQL_BINLOG_SERVER_DISABLE_HEARTBEAT` logic can be automated. Isn't it?

PR #2579, opened 2026-09-22
Unify MySQL integration tests for 5.7, 8.0, 8.4, and 9.7

dvcer:
Unify MySQL integration tests for 5.7, 8.0, 8.4, and 9.7 using shared Percona Dockerfiles, Compose services, and a single Make target. Default to 8.4 base tests and run all five suites for every version in GitHub Actions.


.github/workflows/dockertests-par.yml
```diff
@@ -76,6 +76,12 @@ jobs:
             printf '"make PG_MAJOR=%s pg_integration_test",' "$v"
           done)
 
+          mysql=$(for series in 57 80 84 97; do
+            for suite in base_tests delete_tests copy_tests xbtool_tests binlog_server_tests; do
+              printf '"make USE_BROTLI= USE_LZO= USE_LIBSODIUM= MYSQL_SERIES=%s MYSQL_TEST_DIR=%s mysql_integration_test",' "$series" "$suite"
```
ostinru: Let's make test names readable - move important part for variables to the beginning, and not important to the end.

Current test names are truncated:
```
parallel / test (make USE_BROTLI= USE_LZO= USE_LIBSODIUM= MYSQL_SERIES=80 MYSQL_TEST_DIR=base_tests mysql_in...
```

ostinru: probably, we can get rid of `MYSQL_TEST_DIR` - just run all tests )

ostinru (approved):

.github/workflows/dockertests-par.yml
dvcer: Agreed. I removed MYSQL_TEST_DIR. It was useful for running a subset of tests, so I added MYSQL_TEST_FILTER to filter tests by name instead. Thanks.

